PR #8, opened 2017-04-17
fix(docs): Corrected location of public folder

ryardley:
Public folder provided in docs was incorrect as `src/public` when in fact sku was looking in `./public`. This PR fixes this.


markdalgleish:
Oops—not that it's a big deal, but this documentation change triggered a patch release because of the commit message. Something to look out for next time.

ryardley:
Ahh right yup probably should have been a `docs` descriptor. Apologies!

markdalgleish:
No worries, I didn't realise either 👍

PR #9, opened 2017-04-23
docs(setup): Add git instructions, use diff notation for npm scripts

markdalgleish:
Ran through the setup process and found a few rough edges, so figured I'd clean them up as I went.


markdalgleish:
Just updated the PR so that diff notation is only used for deletion, since the + symbols make it difficult to copy-and-paste the new npm scripts.

PR #17, opened 2017-06-23
feat(webpack): Allow PostCSS to be used independently of Less

nicolassenechal:
Also added support for theming.

If you have a better idea to generate the css loaders, please let me know as I am not 100% happy


mattsjones:
Can we get some documentation around how to utilise the theming and what you can achieve with it?

PR #19, opened 2017-06-30
feat(lint): Add lint script

### LINKED ISSUE #15: Setup eslint with IDE support
Tasks to complete:
- [ ] Setup linter to run eslint as part of a lint (or more general analysis?) script
- [ ] Decide on approach for [module resolution linting](https://github.com/benmosher/eslint-plugin-import). After playing with it I don't think the babel resolver is the way to go. Simplest option is to use the [Node resolver](https://www.npmjs.com/package/eslint-import-resolver-node) which I'm thinking is the way to go. 

We currently have the sku eslint config [here](https://github.com/seek-oss/eslint-config-sku]) which enables us to expose the settings to an IDE (such as Atom).

mattsjones:
Resolves #15


package.json
```diff
@@ -53,10 +53,12 @@
     "babel-preset-react": "^6.22.0",
     "babel-preset-react-optimize": "^1.0.1",
     "bluebird": "^3.5.0",
+    "chalk": "^2.0.1",
     "classnames": "^2.2.5",
     "cross-spawn": "^5.0.1",
     "css-loader": "^0.26.1",
     "deasync-promise": "^1.0.1",
+    "eslint-config-sku": "github:seek-oss/eslint-config-sku",
```
markdalgleish: Is this not on npm? If not, shouldn't we publish it?

markdalgleish:
Might be a good idea to add this to the readme too, but it doesn't need to block this PR.

markdalgleish:
Nice one, looks good 👍

PR #21, opened 2017-07-05
feat(format): Add Prettier to new Format script and linting process

jahredhope:
[Prettier](https://github.com/prettier/prettier) is an opinionated formatter that can both format the code for the developer but also check that the formatting is correct during the CI build.

This change will add support for `sku format` to have prettier format all source code, currently limited to 'src' folder and 'js' files. The required formatting is then asserted in the `sku lint` step.

Implemented:
- New `sku format` will format project code
- Modified `sku lint` will now check for correct formatting

Not implemented:
- lint-staged step to either check for or fix the formatting of code.
- Ability to toggle prettier off for a project to avoid it's linting stage
- Format LESS files
- Documentation of `sku format` and formatting requirements

This PR relates to the change in eslint-config-sku to remove prettier concerns from ESLint rules in
https://github.com/seek-oss/eslint-config-sku/pull/3

### DIFF
--- a/config/jest/jest.config.js
+++ b/config/jest/jest.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   testPathIgnorePatterns: ['<rootDir>[/\\\\](dist|node_modules)[/\\\\]'],
   moduleNameMapper: {
-    '\(seek-style-guide\/react|\.(css|less)$)': require.resolve(
+    '(seek-style-guide/react|.(css|less)$)': require.resolve(
       'identity-obj-proxy'
     )
   },
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -16,10 +16,7 @@ const jsLoaders = [
 ];
 
 const makeCssLoaders = (options = {}) => {
-  const {
-    server = false,
-    styleGuide = false
-  } = options;
+  const { server = false, styleGuide = false } = options;
 
   const debugIdent = isProductionBuild
     ? ''
@@ -58,7 +55,7 @@ const makeCssLoaders = (options = {}) => {
       // Hacky fix for https://github.com/webpack-contrib/css-loader/issues/74
       loader: require.resolve('string-replace-loader'),
       options: {
-        search: '(url\\([\'"]?)(\.)',
+        search: '(url\\([\'"]?)(.)',
         replace: '$1\\$2',
         flags: 'g'
       }
@@ -67,9 +64,7 @@ const makeCssLoaders = (options = {}) => {
 };
 
 const makeImageLoaders = (options = {}) => {
-  const {
-    server = false
-  } = options;
+  const { server = false } = options;
 
   return [
     {
--- a/lib/runPrettier.js
+++ b/lib/runPrettier.js
@@ -9,8 +9,8 @@ function runPrettier({ write, listDifferent, config }) {
   const prettierConfig = config || {};
 
   const prettierPath = path.join(
-    path.resolve('node_modules/prettier'),
-    'bin/prettier.js'
+    require.resolve('prettier'),
+    '../bin/prettier.js'
   );
 
   const prettierArgs = [filePattern];
--- a/lib/runPrettier.js
+++ b/lib/runPrettier.js
@@ -5,13 +5,22 @@ const cwd = process.cwd();
 
 const filePattern = 'src/**/*.js';
 
+/*
+ * Prettier must be ran from
+ * command line to make use of command line output.
+ * Use the bin file identified by it's package.json
+ */
+function getPrettierBinLocation() {
+  const prettierPackageJson = require('prettier/package.json');
+  return require.resolve(
+    path.join('prettier', prettierPackageJson.bin.prettier)
+  );
+}
+
 function runPrettier({ write, listDifferent, config }) {
   const prettierConfig = config || {};
 
-  const prettierPath = path.join(
-    require.resolve('prettier'),
-    '../bin/prettier.js'
-  );
+  const prettierBinPath = getPrettierBinLocation();
 
   const prettierArgs = [filePattern];
 
@@ -34,7 +43,7 @@ function runPrettier({ write, listDifferent, config }) {
     stdio: 'inherit'
   };
 
-  const prettierProcess = spawn(prettierPath, prettierArgs, processOptions);
+  const prettierProcess = spawn(prettierBinPath, prettierArgs, processOptions);
 
   return new Promise((resolve, reject) => {
     prettierProcess.on('exit', exitCode => {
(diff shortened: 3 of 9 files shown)


config/jest/jest.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   testPathIgnorePatterns: ['<rootDir>[/\\\\](dist|node_modules)[/\\\\]'],
   moduleNameMapper: {
-    '\(seek-style-guide\/react|\.(css|less)$)': require.resolve(
+    '(seek-style-guide/react|.(css|less)$)': require.resolve(
```
markdalgleish: Isn't this changing the regex? It's no longer escaping the `.`, which means it now matches any character instead.

config/webpack/webpack.config.js
```diff
@@ -58,7 +55,7 @@ const makeCssLoaders = (options = {}) => {
       // Hacky fix for https://github.com/webpack-contrib/css-loader/issues/74
       loader: require.resolve('string-replace-loader'),
       options: {
-        search: '(url\\([\'"]?)(\.)',
+        search: '(url\\([\'"]?)(.)',
```
markdalgleish: Same issue here, I believe.

config/jest/jest.config.js
jahredhope: From my understanding. And I could be wrong. `\(` is the same as `(`, the `\` is escaping the next character but the `(` doesn't need to be escaped. I would expect this to need to be `\\(`, as in a literal `\` then a literal `(`.
I'll do some testing and get back to you.
If correct, i'd probably prefer to change it to `\\` as we'll be fighting prettier otherwise.

jahredhope: Testing came back as below.
`new RegExp('a.').exec('aa')` => [ 'aa', index: 0, input: 'aa' ]
`new RegExp('a\.').exec('aa')` => [ 'aa', index: 0, input: 'aa' ]
`new RegExp('a\\.').exec('aa')` => null

From this I believe that the `\.` in the above code was always resolving to `.`, and was checking for any character rather than a literal `.`.

lib/runPrettier.js
```diff
@@ -0,0 +1,61 @@
...
+
+const cwd = process.cwd();
+
+const filePattern = 'src/**/*.js';
+
+function runPrettier({ write, listDifferent, config }) {
+  const prettierConfig = config || {};
+
+  const prettierPath = path.join(
+    path.resolve('node_modules/prettier'),
```
mattsjones: Not a big deal but we've used `require.resolve` for this in most other places.

lib/runPrettier.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const prettierProcess = spawn(prettierPath, prettierArgs, processOptions);
+
+  return new Promise((resolve, reject) => {
+    prettierProcess.on('exit', exitCode => {
+      if (exitCode === 0) {
+        resolve(exitCode);
+        return;
+      }
+      reject(exitCode);
```
mattsjones: Got the feeling we're going to be doing this kind of stuff a lot. Might be worth creating some process utils. Not need for this PR though.

config/jest/jest.config.js
markdalgleish: Ah yeah—the fact that the regex is a string is what threw me.

config/webpack/webpack.config.js
markdalgleish: Nope, ignore me 😄

lib/runPrettier.js
jahredhope: As discussed, require.resolve will resolve to the filename directed to by the packages package.json `main` attribute, and I'm trying to find the bin file.
I'd like to leave it as is, but perhaps some kind of reference to %PATH% would be more appropriate to find bin files in the future, not sure and needs discussion.

markdalgleish: Does `path.resolve` handle node module resolution like this?

markdalgleish: Or is this relative to the current working directory?

jahredhope: In this situation path.resolve will resolve to an absolute path from the cwd.

https://nodejs.org/docs/latest/api/path.html#path_path_resolve_paths

markdalgleish: Might be safer to use a path relative to `require.resolve`, since AFAIK there's technically no guarantees where Prettier will be on the file system.

lib/runPrettier.js
```diff
@@ -0,0 +1,61 @@
...
+
+const cwd = process.cwd();
+
+const filePattern = 'src/**/*.js';
+
+function runPrettier({ write, listDifferent, config }) {
+  const prettierConfig = config || {};
+
+  const prettierPath = path.join(
+    require.resolve('prettier'),
```
jahredhope: This is now relative to require.resolve.
Still not ideal as any change to the depth of the `main` attribute in Prettier will break this implementation.

markdalgleish: Maybe you could resolve relative to `package.json`, and join it with the value under `bin.prettier`: https://github.com/prettier/prettier/blob/9290dd88038a8c155730beac3201e2f861697503/package.json#L6

lib/runPrettier.js
```diff
@@ -0,0 +1,71 @@
...
+const cwd = process.cwd();
+
+const filePattern = 'src/**/*.js';
+
+/*
+ * Prettier must be ran from
+ * command line to make use of command line output.
+ * Use the bin file identified by it's package.json
+ */
+function getPrettierBinLocation() {
```
jahredhope: Now resolving based on prettier's package.json `bin` attribute

lib/runPrettier.js
```diff
@@ -0,0 +1,70 @@
...
+const path = require('path');
+
+const cwd = process.cwd();
+
+const filePattern = 'src/**/*.js';
+
+/*
+ * Prettier must be ran from
+ * command line to make use of command line output.
+ * Use the bin file identified by it's package.json
```
markdalgleish: Typo, should be "its" 😉

jahredhope:
I'm keen to get this merged. @mattsjones do you approve?

lib/runPrettier.js
```diff
@@ -0,0 +1,70 @@
...
+const cwd = process.cwd();
+
+const filePattern = 'src/**/*.js';
+
+/*
+ * Prettier must be ran from
+ * command line to make use of command line output.
+ * Use the bin file identified by Prettier's package.json
+ */
+function getPrettierBinLocation() {
```
mattsjones: Any reason your not using arrow functions here? It doesn't really matter but the rest of the project does and consistency is good. The whole point of this PR yeah? Haha

jahredhope: I've for whatever historical reason kept these kind of `just separating out code` functions as function declarations. No reason. Will change for conformity.

PR #25, opened 2017-07-10
feat(package): update lint rules

LeviSeek:
This PR would link sku to the latest "eslint-config-sku" version where some lint rules are disabled owned by Prettier.


markdalgleish:
This change actually isn't necessary since the new version number still matches the version range specified in this package. Shall we leave these non-breaking updates going forwards?

LeviSeek:
no need to do a separate release for this non-breaking change.

PR #26, opened 2017-07-14
feat(webpack): Render for each locale

mattsjones:
Will add docs. Just opening for discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,39 @@ module.exports = {
 }
 ```
 
+### Locales
+
+Often we render multiple versions of our application for different locations, eg. Australia & New Zealand. To render an HTML file for each location you can use the locales option in `sku.config.js`. Locales are preferable to [monorepos](#monorepo-support) when you need to render multiple versions of your HTML file but only need one version of each of the assets (JS, CSS, images, etc). Note: You can use `locales` inside a monorepo project.
+
+The `locales` option accepts an array of strings representing each locale you want to render HTML files for.
+
+```js
+module.exports = {
+  locales: ['AU', 'NZ']
+}
+```
+
+For each locale, sku will call your `render.js` function and pass it the locale as a parameter.
+
+```js
+const render = ({ locale }) => (
+  `<div>Rendered for ${locale}</div>`
+)
+```
+
+The name of the HTML file that is generated will be suffixed by `-{locale}`.
+
+eg.
+```js
+module.exports = {
+  locales: ['AU', 'NZ']
+}
+```
+
+will create `index-AU.html` & `index-NZ.html`.
+
+Note: When running the app in dev mode only one HTML file will be created, defaulting to the first listed locale.
+
 ### Monorepo Support
 
 If you need to build multiple projects in the same repo, you can provide an array of config objects.
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -94,7 +94,7 @@ const svgLoaders = [
   }
 ];
 
-const buildWebpackConfigs = builds.map(({ name, paths, env }) => {
+const buildWebpackConfigs = builds.map(({ name, paths, env, locales }) => {
   const envVars = lodash
     .chain(env)
     .mapValues((value, key) => {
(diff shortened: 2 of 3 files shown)


markdalgleish:
I love how little code this takes 😍

markdalgleish:
@mattsjones looks like this hasn't gone through Prettier.

config/webpack/webpack.config.js
```diff
@@ -222,7 +222,17 @@ const buildWebpackConfigs = builds.map(({ name, paths, env }) => {
       },
       plugins: [
         new webpack.DefinePlugin(envVars),
-        new StaticSiteGeneratorPlugin()
+        ...locales.map(
```
markdalgleish: The documentation says it will only render a single file in dev mode, but this appears to always map over every locale?

README.md
```diff
@@ -248,6 +248,39 @@ module.exports = {
 }
 ```
 
+### Locales
+
+Often we render multiple versions of our application for different locations. Eg. Australia & New Zealand. To render an html file for each location you can use the locales option in `sku.config.js`.
```
markdalgleish: Couple of tweaks:
- Should be "locations, e.g. Australia and New Zealand"
- Should be "HTML"

README.md
```diff
@@ -248,6 +248,39 @@ module.exports = {
...
+
+Locales accepts an array of strings representing each locale you want to render html files for.
```
markdalgleish: When referencing an option/argument name, we should use `locales` instead of "Locales". Also, you'll need to re-order this sentence so it's not the first word, otherwise it'll look weird.

(Also need to update to "HTML" here too)

README.md
```diff
@@ -248,6 +248,39 @@ module.exports = {
...
+
+```js
+module.exports = {
+  locales: ['AU', 'NZ']
+}
+```
+
+For each locale, sku will call your `render.js` function and pass it the locale as a local.
```
markdalgleish: Feel like using "locale" and "local" so close together is confusing. In fact, in this case I think you could avoid using the word "local" entirely.

README.md
```diff
@@ -248,6 +248,39 @@ module.exports = {
...
+
+```js
+const render = ({ locale }) => (
+  `<div>Rendered for ${locale}</div>`
+)
+```
+
+The name of the html file that is generated will be suffixed by `-{locale}`.
```
markdalgleish: "HTML" here too.

config/webpack/webpack.config.js
mattsjones: Somehow I missed actually implementing this :facepalm: Will do.

markdalgleish: Perfect example of why providing docs with a PR is great for reviewers 😄

config/webpack/webpack.config.js
```diff
@@ -222,7 +222,17 @@ const buildWebpackConfigs = builds.map(({ name, paths, env }) => {
       },
       plugins: [
         new webpack.DefinePlugin(envVars),
-        new StaticSiteGeneratorPlugin()
+        ...locales.slice(0, isProductionBuild ? locales.length : 1).map(
```
markdalgleish: Ooooh, nice 👏

markdalgleish:
Looks good, just one minor suggestion—might be good to explain when and why this is preferable to monorepos, talking about how we generate a single set of assets that supports both locales.

README.md
```diff
@@ -248,6 +248,39 @@ module.exports = {
 }
 ```
 
+### Locales
+
+Often we render multiple versions of our application for different locations, eg. Australia & New Zealand. To render an HTML file for each location you can use the locales option in `sku.config.js`. Locales are preferable to [Monorepos](monorepo-support) when you need to render multiple versions of your HTML file but only need one version of each of the assets (JS, CSS, images, etc). Note: You can use `locales` inside a monorepo project.
```
markdalgleish: Sorry—I think `monorepo-support` needs to be `#monorepo-support` for the link to work correctly. Also, "Monorepos" should be lowercase.

PR #29, opened 2017-07-18
feat(webpack): Add option to compile external packages as internal code

mattsjones:


README.md
```diff
@@ -248,6 +248,18 @@ module.exports = {
 }
 ```
 
+### Compile Packages
+
+Sometimes we have packages which need to be compiled through webpack along with our app. An great example of this is [seek-style-guide](https://github.com/seek-oss/seek-style-guide). Out of the box sku supports loading the seek-style-guide but if you need to treat other packages in this way you can use `compilePackages`.
```
markdalgleish: Docs should be written to address the user directly, rather than using words like "we". Also, in this case, it's probably important to frame the context of why you'd want to do this, e.g. "Sometimes you might want to extract and share code between sku projects, but this code is likely to rely on the same tooling and language features that this toolkit provides." etc.

config/webpack/webpack.config.js
```diff
@@ -152,14 +156,14 @@ const buildWebpackConfigs = builds.map(
                 use: makeCssLoaders()
               })
             },
-            {
+            ...paths.compilePackages.map(package => ({
```
markdalgleish: Ooooh, this is nice 👏

config/webpack/webpack.config.js
```diff
@@ -152,14 +156,14 @@ const buildWebpackConfigs = builds.map(
...
               test: /\.less$/,
-              include: paths.seekStyleGuide,
+              include: package,
               use: ExtractTextPlugin.extract({
                 fallback: 'style-loader',
-                use: makeCssLoaders({ styleGuide: true })
+                use: makeCssLoaders({ package })
```
markdalgleish: Love this 😍

config/webpack/webpack.config.js
```diff
@@ -198,19 +202,19 @@ const buildWebpackConfigs = builds.map(
           rules: [
             {
               test: /\.js$/,
-              exclude: thirdPartyModulesRegex,
+              include: internalJs,
```
markdalgleish: This could break apps that don't follow the `src` convention.

markdalgleish: What was the need for this change?

mattsjones: I thought we agreed that src is a set convention? Linting relies on it as well.

mattsjones: The reason for the change was I didn't want to make a dynamic regex so I moved to an absolute path approach. Personally I feel it's much cleaner add style guide isn't Harford throughout so much.

mattsjones: Hard coded*

PR #36, opened 2017-07-27
I'm going to have to rename this commit anyway

mengtzu:
Is this a breaking change?  I hope nobody's relying on this old version of the style guide


mattsjones:
🤔 This is definitely the right thing to do. I'm happy for this to be a fix release. Just confirm with Aden.

PR #38, opened 2017-07-28
feat(styles): add css-in-js-support

mengtzu:

PR #39, opened 2017-07-28
feat(args): pass through tenant arg as env variable

mengtzu:
This is the absolute minimum implementation of a tenant feature, it passes through to be seen by the consuming app and styleguide, but doesn't (yet) have any knowledge of the tenants, map styleguide aliases or anything like that.  It's enough to get us started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,49 +359,46 @@ will create `index-AU.html` & `index-NZ.html`.
 
 Note: When running the app in dev mode only one HTML file will be created, defaulting to the first listed locale.
 
-
 ### Tenants
 
 Applications that serve multiple branded experiences need to know which brand (or tenant) to render.  If invoked with a `--tenant` or `--t` command line, sku will pass through the supplied value as an environment variable named `SKU_TENANT`.
 
 For example, if we invoked
 
-`sku --tenant=jobStreet`
+`sku --tenant=awesomeBrand`
 
 And our application exposes the tenant as a prop to react app like so:
 
 ```
 <div id="app">
-    ${renderToString(
-        <App tenant={process.env.SKU_TENANT} />
-    )}
+  ${renderToString(
+    <App tenant={process.env.SKU_TENANT} />
+  )}
 </div>
 ```
 
 Then we can serve different experiences based on the tenant prop:
 
 ```
-import JobStreetHeader from `./components/JobStreetHeader/JobStreetHeader`;
-import JobsDBHeader from './components/JobsDBHeader/JobsDBHeader';
+import AwesomeHeader from './components/AwesomeHeader';
+import AmazingHeader from './components/AmazingHeader';
 
 const renderHeader = (tenant) => {
-    if (tenant === 'jobStreet') {
-        return (<JobStreetHeader>)
-    } else {
-        return (JobsDBHeader)
-    }
+  if (tenant === 'awesomeBrand') {
+    return (<AwesomeHeader>)
+  } else {
+    return (AmazingHeader)
+  }
 };
 
 const App = (props) => {
-    return (
-        <div>
-            {renderHeader(props.tenant)}
-        </div>
-    );
+  return (
+    <div>
+      {renderHeader(props.tenant)}
+    </div>
+  );
 }
 ```
-
-
 
 ### Development server
 
--- a/README.md
+++ b/README.md
@@ -359,6 +359,50 @@ will create `index-AU.html` & `index-NZ.html`.
 
 Note: When running the app in dev mode only one HTML file will be created, defaulting to the first listed locale.
 
+
+### Tenants
+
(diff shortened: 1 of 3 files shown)


config/builds.js
```diff
@@ -37,7 +37,9 @@ const builds = buildConfigs
   })
   .map(buildConfig => {
     const name = buildConfig.name || '';
-    const env = buildConfig.env || {};
+    const env = buildConfig.env || {
+      SKU_TENANT: args.tenant
```
mengtzu: This is how I did it in my spike, but I'm not sure it will let anything with a buildConfig pick up the tenant variable.  Wasn't sure if I wanted to make it a compulsory pickup by spreading `buildConfig.env` into an object with `SKU_TENANT` and exporting that though.

config/args.js
```diff
@@ -9,6 +9,12 @@ const optionDefinitions = [
     alias: 'e',
     type: String,
     defaultValue: 'production'
+  },
+  {
+    name: 'tenant',
+    alias: 't',
+    type: String,
+    defaultValue: 'seek'
```
markdalgleish: I'm not sure this should be provided as a default. Maybe if we choose a default, it should be the first tenant defined in your sku config or something?

mengtzu: I'm happy to not have a default.  Consuming apps would need to handle `SKU_TENANT` being undefined, but on reflection having to handle a case of `seek` is probably worse.

mengtzu:
I removed the default and added a bunch of doco (let me know if it's too much)

README.md
```diff
@@ -359,6 +359,50 @@ will create `index-AU.html` & `index-NZ.html`.
 
 Note: When running the app in dev mode only one HTML file will be created, defaulting to the first listed locale.
 
+
```
markdalgleish: Looks like additional whitespace here? 🕵️

README.md
```diff
@@ -359,6 +359,50 @@ will create `index-AU.html` & `index-NZ.html`.
...
+    return (
+        <div>
+            {renderHeader(props.tenant)}
+        </div>
+    );
+}
+```
+
+
+
```
markdalgleish: More whitespace here too 🕵️

README.md
```diff
@@ -359,6 +359,50 @@ will create `index-AU.html` & `index-NZ.html`.
...
+        return (JobsDBHeader)
+    }
+};
+
+const App = (props) => {
+    return (
+        <div>
+            {renderHeader(props.tenant)}
+        </div>
+    );
```
markdalgleish: Code block should be indented with 2 spaces.

README.md
```diff
@@ -359,6 +359,50 @@ will create `index-AU.html` & `index-NZ.html`.
...
+        <App tenant={process.env.SKU_TENANT} />
+    )}
+</div>
+```
+
+Then we can serve different experiences based on the tenant prop:
+
+```
+import JobStreetHeader from `./components/JobStreetHeader/JobStreetHeader`;
+import JobsDBHeader from './components/JobsDBHeader/JobsDBHeader';
```
markdalgleish: Can we replace these with more generic examples?

config/builds.js
markdalgleish: Might be worth defaulting to an empty string, like we do with the `name` config.

mengtzu:
Updated @markdalgleish

mengtzu:
I've changed the example to remove the import antipattern.  It's still a highly contrived example.

Other options:
- merge #38 and use a css-in-js example
- leave feature undocumented until we have a more mature tenancy solution in place
- use an example relying on further tools like `react-intl`

mengtzu:
Removed doco! ✂️

config/builds.js
```diff
@@ -37,7 +37,9 @@ const builds = buildConfigs
   })
   .map(buildConfig => {
     const name = buildConfig.name || '';
-    const env = buildConfig.env || {};
+    const env = buildConfig.env || {
+      SKU_TENANT: args.tenant || ''
```
markdalgleish: If I'm reading this right, this means that if you define environment variables, you'll no longer have access to the tenant option?

mengtzu: Yeah, I had a comment on that but it got blown away.

Could just spread it in?  We could have a convention of not specifying your own SKU_* in build configs and then supplying all of them every time.

markdalgleish: Yeah, it makes sense to me that your environment variables would be added to the default, and you could potentially override too.

config/builds.js
```diff
@@ -37,7 +37,10 @@ const builds = buildConfigs
   })
   .map(buildConfig => {
     const name = buildConfig.name || '';
-    const env = buildConfig.env || {};
+    const env = {
+      SKU_TENANT: args.tenant || '',
+      ...(buildConfig.env ? { ...buildConfig.env } : {})
```
mengtzu: So basically I'm going to use conditional spreads for everything now

markdalgleish: Ha, join the club 😄

nicolassenechal:
Fine with this approach. Would prefer an external reviewer to approve though

Issue #46, opened 2017-08-14
RFC: Server-side rendering support


nicolassenechal:
# Summary

Add support for SSR

# Motivation

SSR is required for performance and SEO

# Detailed design

1. Generate server.js file consumable by a node server
2. Add mechanism to start server in sku
3. Add mechanism to server-side hot module reloading in development
4. Add mechanism to resolve API calls

# How We Teach This

Presented as an alternative when static rendering does not make sense. 
Already used by many projects, so acceptance should not be a concern

# Drawbacks

Possible confusion with static rendering and ssr

# Alternatives

We could have batch jobs crawling our websites regularly and doing static rendering of all the pages and storing them in an S3 bucket.

# Unresolved questions

How configurable do we need that server to be?

mattsjones:
Is sku in charge of running your production node server? or does it purely output the code?

Also I don't hot reloading should be part of the initial setup for SSR. It's more a nice to have and tbh it's extremely difficult to get working.

nicolassenechal:
The first PR only covers the first point: delivering server.js. The other 3 points are really up to negotiation on which should be done and how.

I think hot module reloading might actually be easier than it sounded initially.

mattsjones:
I hope you're right, haven't had much luck with it though. Also, I think hot reloading should be for both static sites and SSR.

nicolassenechal:
will create separate RFC for missing features

PR #47, opened 2017-08-14
feat(ssr): add initial undocumented SSR support

nicolassenechal:
## Commit Message For Review
Added basic setup for SSR:
- Reuses setup for static rendering as I was keen to avoid multiplication of configuration files
- Does not provide a server, hot-module reloading or API consumption
- Does not force the user to provide either a static or a server-side entry point and does not default anymore back to render.js for static rendering

RFC URL:
https://github.com/seek-oss/sku/issues/46

REASON FOR CHANGE:
Base support for SSR

EXAMPLE USAGE:
In a file called server.js:
```js
import render from './render';
import express from 'express';

const server = express();

server.get('/', (req, res) => {
  res.send(render());
});

server.listen(8080);
```

In the sku.config.js file:
```module.exports = [
  {
    name: 'home',
    env: extendDefaultEnv({}),
    entry: {
      client: 'src/client.js',
      server: 'src/server.js'
    },
    public: 'public',
    target: 'dist'
   }];```

Start your node server:
```node ./dist/server.js```

### DIFF
--- a/config/builds.js
+++ b/config/builds.js
@@ -51,6 +51,15 @@ const builds = buildConfigs
     const webpackDecorator =
       buildConfig.dangerouslySetWebpackConfig || defaultDecorator;
 
+    let renderEntry = path.join(cwd, 'src/render.js');
+    let serverEntry = null;
+    if (entry.render) {
+      renderEntry = path.join(cwd, entry.render);
+    } else if (entry.server) {
+      renderEntry = null;
+      serverEntry = path.join(cwd, entry.server);
+    }
+
     const paths = {
       src: path.join(cwd, 'src'),
       compilePackages: [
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by markdalgleish** [comment]
What's the reasoning behind not providing a server at this point? Is this more of an undocumented MVP?

File: config/builds.js [hunk 1]
```diff
@@ -60,7 +60,8 @@ const builds = buildConfigs
         )
       ],
       clientEntry: path.join(cwd, entry.client || 'src/client.js'),
-      renderEntry: path.join(cwd, entry.render || 'src/render.js'),
+      renderEntry: entry.render ? path.join(cwd, entry.render) : null,
```
markdalgleish: Is this a breaking change for apps without a `sku.config.js`?

nicolassenechal:
No server: keen to cut the whole task into smaller pieces

config/builds.js
nicolassenechal: I did not even know you did not need a config file. Yes, that would be a breaking change. We can make the logic more complex to support this

markdalgleish: If we can make it a non breaking change, that would be better, I think.

nicolassenechal: It's not a breaking change anymore

PR #48, opened 2017-08-14
chore(tests): add test infrastructure

mengtzu:
## Commit Message For Review

REASON FOR CHANGE:

Lack of tests have made it difficult for contributors to work on sku with confidence that they would not ship bugs or violate the contract with consuming apps.  This change introduces the ability to run sku against a suite of test configurations/apps and compare the results to known-good snapshots.

EXAMPLE USAGE:

In `test/test-cases` create a folder named for your test (eg `example-test`).

In `test/test-cases/example-test` create a sku.config.js as follows:

```js
module.exports = {
    target: 'actual-output', //this is compulsory, test scripts look for this directory
    propertyYouWantToTest: someTestValue
}
```

Add as much app as you require to test your config.

Run `npm test`  This will fail, as you have no snapshot to compare with.

Copy the contents of `test/test-cases/example-test/actual-output` to `test/test-cases/example-test/expected-output`.

Run `npm test`  This should pass.

You may now make changes to sku and run the tests to ensure your tested config still functions.


config/args.js
```diff
@@ -1,7 +1,11 @@
 const commandLineArgs = require('command-line-args');
 
 const scriptNameRegex = /scripts[\/\\]([\w-]*)\.js$/i;
-const scriptName = process.argv[1].match(scriptNameRegex)[1];
+const scriptNameByRegex = process.argv[1].match(scriptNameRegex);
+
+const scriptName = scriptNameByRegex
```
mengtzu: Very happy to hear better ideas about this bit.

.idea/watcherTasks.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectTasksOptions" suppressed-tasks="Babel" />
+</project>
```
markdalgleish: Can you remove this file? Might want to add this to your global gitignore too.

test/test-cases/hello-world/sku.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  target: 'actual-output'
```
markdalgleish: I'd probably leave this as the default, build to `dist`, since this is normal sku behaviour. Shouldn't be hard to make our test setup use this folder structure.

test/test-cases/hello-world/actual-output/main.js
```diff
@@ -0,0 +1,31 @@
...
+    return Object.prototype.hasOwnProperty.call(e, n);
+  }), (n.p = ''), n((n.s = 0));
+})([
+  function(e, n, t) {
+    e.exports = t(1);
+  },
+  function(e, n) {
+    document.getElementById('app').innerHTML = 'Hello world!';
+  }
+]);
```
markdalgleish: Asserting against the contents of this file means that this test is a bit fragile—any changes to how this code is compiled or optimised could result in a false negative. Not sure what we should do about it in the short term, though, because this is certainly better than having no tests at all.

test/test-cases/hello-world/sku.config.js
mattsjones: I agree. Also I want to be able to test scenarios which have no `sku.config.js`.

config/args.js
mattsjones: I'm thinking we should move to [command-line-commands](https://github.com/75lb/command-line-commands)?

package.json
```diff
@@ -16,7 +16,7 @@
     "semantic-release": "semantic-release pre && npm publish && semantic-release post"
   },
   "lint-staged": {
-    "*.js": [
+    "{bin,config,scripts}/**/*.js": [
```
mengtzu: I couldn't get the exclude to work, so I changed the pattern to match `format` and `format-check`.

Is that sufficient, or is it worth adding an extra glob to explicitly include `test/index.test.js`?

config/args.js
mengtzu: command-line-commands looks good, but I don't think it fixes the problem here where the argument is actually a path :<

package.json
mattsjones: Just leave it for now. Really want to get this merged. We can fix the prettier config later if it becomes an issue.

Issue #50, opened 2017-08-16
RFC: Provide express server for SSR


nicolassenechal:
# Summary

The current SSR implementation implies that users have to implement their own server. We should provide an express server handling this kind of functionality.

# Motivation

This is a required stepping stone for further improvements, such as routing and server-side API resolution.

# Detailed design

- Add "scripts/server.js" to the project
- In this file, create and start an express server
- Expose this file through sku methods
- Allow extra middlewares to be passed in through the sku config and consumed by the express app

# How We Teach This

Readme should be updated to explain how to use and extend the express server.

# Drawbacks

That extra configuration option might make sku configs more complex and harder to understand.

# Alternatives

We could let users duplicate the logic in their applications.

# Unresolved questions

- This does not cover extra middleware we will want to include at some stage (hot module reloading for instance)
- There might be some confusion for end users between "sku start" and "sku server"

markdalgleish:
The big question for me on this is, what does your project's `server.js` file export, if anything? How does sku make use of what you provide?

The "detailed design" section is a little sparse at the moment. Could you go into a little more detail about the API?

nicolassenechal:
I had two designs in mind:
- Create the server, apply default middlewares, export server and let user add their own middlewares and start the server in their project. Export the server
- Accept user middlewares and port as a sku config options, create the server, apply default middlewares, apply user middlewares, start the server. Don't export anything

I am leaning towards the second approach as it works out of the box.

markdalgleish:
Thinking more about the second one—I wonder if `server.js` could work like this:

```js
export default app => {
  app.get('*', () => { ... });
}
```

We pass you an express app, you add your stuff to it?

nicolassenechal:
Could work too. According to the docs, you can consume another express app as any other middleware. I have never tried it but it feels like it would achieve the same result.

markdalgleish:
There's obviously a big tradeoff here, whether or not we couple ourselves to a particular library like this. What's the advantage? The alternative is that we could do something like this:

```js
import express from 'express';

export default ({ port }, callback) => {
  const app = express();

  app.get(...);

  app.listen(port, callback);
};
```

markdalgleish:
Looking into [webpack dev middleware](https://webpack.js.org/guides/development/#using-webpack-dev-middleware)...

[...]

PR #55, opened 2017-08-28
feat(ssr): add express and improve build configuration for ssr

nicolassenechal:
Commit Message for Review
======================

Added Express and improved build configuration for SSR

RFC URL: https://github.com/seek-oss/sku/issues/50

REASON FOR CHANGE: Allows users to render their application server-side with minimal configuration

EXAMPLE USAGE:

server.js
```js
import render from './render.js';

const mw = function (req, res, next) {
  console.log('Time:', Date.now())
  next()
};

export default {
    renderCallback: (req, res) => {
        res.send(render());
    },
    middlewares: [mw]
};
```

sku.config.js
```js
module.exports = {
  entry: {
    client: 'src/client.js',
    server: 'src/server.js'
  },
  public: 'src/public',
  target: 'dist',
  port: 3000
};
```

`renderCallback` is required
`middlewares` is not required. Needs to be a function, an array or an express instance

### DIFF
--- a/config/server/server.js
+++ b/config/server/server.js
@@ -7,12 +7,7 @@ const port = process.env.PORT || 8080;
 
 const app = express();
 
-if (
-  middleware &&
-  (middleware instanceof Function ||
-    middleware instanceof express ||
-    (middleware instanceof Array && middleware.length > 0))
-) {
+if (middleware) {
   app.use(middleware);
 }
 app.get('*', renderCallback);
--- a/config/server/server.js
+++ b/config/server/server.js
@@ -1,0 +1,19 @@
+const express = require('express');
+const { renderCallback, middlewares } = require('serverEntry').default;
+const port = process.env.PORT || 8080;
+
+const app = express();
+
+if (
+  middlewares &&
+  (middlewares instanceof Function ||
+    middlewares instanceof express ||
+    (middlewares instanceof Array && middlewares.length > 0))
+) {
+  app.use(middlewares);
+}
+app.get('*', renderCallback);
+
+app.listen(port, () => {
+  console.log(`App started on port ${port}`);
+});
--- a/config/server/server.js
+++ b/config/server/server.js
@@ -1,16 +1,19 @@
 const express = require('express');
-const { renderCallback, middlewares } = require('serverEntry').default;
+const {
+  renderCallback,
+  middleware
+} = require('__sku_alias__serverEntry').default;
 const port = process.env.PORT || 8080;
 
 const app = express();
 
 if (
-  middlewares &&
-  (middlewares instanceof Function ||
-    middlewares instanceof express ||
-    (middlewares instanceof Array && middlewares.length > 0))
+  middleware &&
+  (middleware instanceof Function ||
+    middleware instanceof express ||
+    (middleware instanceof Array && middleware.length > 0))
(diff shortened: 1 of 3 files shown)


config/server/server.js
```diff
@@ -0,0 +1,19 @@
...
+const { renderCallback, middlewares } = require('serverEntry').default;
+const port = process.env.PORT || 8080;
+
+const app = express();
+
+if (
+  middlewares &&
+  (middlewares instanceof Function ||
+    middlewares instanceof express ||
+    (middlewares instanceof Array && middlewares.length > 0))
```
tamoore: I don't really understand what the point of this conditional is.

```javascript

() => throw new Error('foo') // passes
[-1] // passes
[undefined] // passes

```

What sort of runtime failure are you trying to catch here?

config/server/server.js
```diff
@@ -0,0 +1,19 @@
+const express = require('express');
+const { renderCallback, middlewares } = require('serverEntry').default;
+const port = process.env.PORT || 8080;
+
+const app = express();
+
+if (
+  middlewares &&
+  (middlewares instanceof Function ||
```
tamoore: Be careful with instanceof, it's well, rough around the edges:

```javascript
Object instanceof Function // true 
```

config/server/server.js
```diff
@@ -0,0 +1,19 @@
...
+    middlewares instanceof express ||
```
tamoore: https://runkit.com/tamoore/59a3e8f35078ee0011bde7a3

config/server/server.js
```diff
@@ -0,0 +1,19 @@
+const express = require('express');
+const { renderCallback, middlewares } = require('serverEntry').default;
```
markdalgleish: FYI `middleware` is already plural, so there shouldn't be an 's' on the end :)

config/webpack/webpack.config.js
```diff
@@ -218,7 +220,15 @@ const buildWebpackConfigs = builds.map(
...
         entry: {
-          render: paths.renderEntry || paths.serverEntry
+          render:
+            paths.renderEntry || path.join(__dirname, '../server/server.js')
+        },
+        resolve: {
+          alias: isRender
+            ? {}
+            : {
+                serverEntry: paths.serverEntry
```
markdalgleish: I feel like we need to use a more self-documenting alias that makes it obvious what's going on.

e.g. `__sku_alias__serverEntry`, so that when you see `require('__sku_alias__serverEntry')` it's a hint something different is going on.

nicolassenechal:
Back for review

tamoore:
LGTM

config/server/server.js
```diff
@@ -0,0 +1,17 @@
+const express = require('express');
+const {
+  renderCallback,
+  middleware
+} = require('__sku_alias__serverEntry').default;
```
mattsjones: Sorry if I've missed something here being late to review. But why are we destructuring a default export? Can we just use named exports instead? https://github.com/standard-things/esm Could be a good idea as well.

markdalgleish: I think the current approach makes sense because the interface is dynamic—`middleware` is optional right now, and there are likely to be more changes in the future.

PR #56, opened 2017-08-31
fix(config): add seek-asia-style-guide to jest config

mengtzu:
## Commit Message For Review

Adds `seek-asia-style-guide` alongside `seek-style-guide` in the jest config for `identity-obj-proxy`.  This should be a temporary measure until a configurable approach can be added.

REASON FOR CHANGE:

Sku apps currently run tests on components importing anything from `seek-asia-style-guide`


mengtzu:
Forking sku instead for the time being; will circle back around to figure out what to do with the jest config

PR #59, opened 2017-09-06
fix(config): add ability to serve generated static files for ssr dev

mengtzu:
## Commit Message For Review

Adds serving for static files in the root directory for server side rendering. 

REASON FOR CHANGE:

Without this, when using SSR in dev requests for `main.js`, `styles.css` and their friends will simply return the output of the render callback.  In most apps this will be the rendered HTML intended for a 404.

### DIFF
--- a/config/server/server.js
+++ b/config/server/server.js
@@ -8,7 +8,12 @@ const port = process.env.PORT || 8080;
 
 const app = express();
 
-app.use(express.static(path.join(__dirname, './')));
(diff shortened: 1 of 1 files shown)


mengtzu:
@mattsjones sure, we're assuming outside dev the files are probably on an s3 bucket somewhere or otherwise not served by our express server, as I understand it.

mengtzu:
@mattsjones added dev check

Issue #62, opened 2017-09-12
RFC: Hot Module Replacement


nicolassenechal:
# Summary
Enables HMR, both for ssr and in the front-end

# Motivation
HMR is useful for improving development experience and speed, avoiding to have to rebuild the project manually systematically

# Detailed design
The design is based on this article: https://hackernoon.com/hot-reload-all-the-things-ec0fed8ab0
A basic and incomplete implementation can be found at: https://github.com/seek-oss/sku/tree/hmr-with-dev-server

## Taks
1. Update `webpack.config` to support hmr when in dev mode
- extra entries
- output.publicPath for client side hmr
- externals for server-side rendering
- extra plugins

2. Update `babel.config` to support hmr (might not be required)
3. Refactor `start.js` and `build.js` to be more re-usable

## Usage
1. The user calls `sku start`
2. If the config is of type `render`, `start.js` creates the webpack devserver with hmr support
3. If the config is of type `server`,`start.js` creates the webpack devserver with hmr support and build the server-side bundle 

## API changes
Two port numbers are now required when using server-side rendering. I would keep the current `port` option, but allow an array to be passed in. The first value would user by the server-side rendering, the second by the client-side hmr.

# How We Teach This
We should point to general documentation about hmr

# Drawbacks
- `webpack.config` is going to become very complex, with test over dev/prod, static/ssr
- that functionality is react-centric and relies on `react-hot-loader`

# Alternatives
I looked at using middleware, instead of webpack dev server. The solution requires a lot more rewriting (branch here: https://github.com/seek-oss/sku/tree/hmr)

# Unresolved questions
Initial tests showed some potential issues with async/await. Will need to investigate more

nicolassenechal:
To make the port setup more self-documenting, I propose that the sku option port be either a number or an object of the form {backend:int, client:int}.
Static rendering would only use the backend attribute in dev mode, and ignore the client attribute all the time.
Server-side rendering would use the backend attribute in both dev and prod, and the client attribute only in dev mode.

PR #65, opened 2017-09-20
feat(hmr): Add client and server hmr for react

nicolassenechal:
## Commit Message For Review
Add React hmr to sku

RFC URL:
https://github.com/seek-oss/sku/issues/62

REASON FOR CHANGE:
This PR aims at facilitating the use of sku for react projects by enabling hmr.

EXAMPLE USAGE:
- You can now provide, as part of your sku config, an array with two port numbers to your build when using SSR rendering. The first one will be used to start the dev server used for client-side hmr. The second one is used for the server-side hmr. You access your application using the second port. 

```js
module.exports = {
  entry: {
    client: 'src/client.js',
    server: 'src/server/server.js'
  },
  public: 'src/public',
  target: 'dist',
  port: [3000, 3300],
  compilePackages: ['seek-asia-style-guide']
}
```

- There is no change if you use static rendering, though this is untested. Static rendering with hmr on the client-side should work

- To enable hmr in your ssr application, you now need to load the client from the client server as those are different
```js
<script type="text/javascript" src="http://localhost:3000/main.js"></script>
```

- To enable hmr client-side, you need to add an AppContainer and define the behaviour whenever a rebuild happens as described in here: https://github.com/gaearon/react-hot-loader/tree/master/docs#migration-to-30
```js
import React from 'react';
import { render } from 'react-dom';
import App from './App/App';
import { AppContainer } from 'react-hot-loader';

const appElement = document.getElementById('app');
render(
  <AppContainer>
      <App />
  </AppContainer>,
  appElement
);

if (module.hot) {
  module.hot.accept('./App/App', () => {
    render(
      <AppContainer>
          <App />
      </AppContainer>,
      appElement
    );
  });
}
```

NOTES:
- Setup tested only against our team project which is not a mono-repo and uses server-side rendering. Happy to validate against other projects, but this emphasises the need for a project we want to test changes against
- The webpack config build function does not test anymore against the environment, but against the script being run. Happy to get rid of the production tests across the whole file or any alternative solution.
- If you use SSR, the dist folder needs to be deleted before starting the server: `rm -rf ./dist && sku start`


config/webpack/webpack.config.js
```diff
@@ -124,31 +127,45 @@ const buildWebpackConfigs = builds.map(
         return JSON.stringify(valueForEnv);
       })
       .set('SKU_ENV', JSON.stringify(args.env))
-      .set('PORT', JSON.stringify(port))
+      .set('PORT', JSON.stringify(port.length > 1 ? port[1] : ''))
```
mengtzu: Need to be PORT or can it be SKU_PORT or similar?

config/server/server.js
```diff
@@ -19,6 +19,10 @@ if (middleware) {
 }
 app.get('*', renderCallback);
 
-app.listen(port, () => {
-  console.log(`App started on port ${port}`);
-});
+if (env === 'production') {
+  app.listen(port, () => {
```
mengtzu: This port also gets listened to in server/index right?  Why is this necessary?

mengtzu:
Apart from above comments, looking good.  It does complicate an already complicated set of config but I don't see how to escape that.

config/webpack/webpack.config.js
nicolassenechal: Good point, worth updating

config/server/server.js
nicolassenechal: server/index is only for dev. This is for production

nicolassenechal: Which implied you found a bug :|

mattsjones:
So I just tried this against profile which is already running `sku`. I'm getting a `Cannot GET /` error. Are there any necessary changes I need to make?

nicolassenechal:
Previous issue should be fixed

nicolassenechal:
Changed port API. Ready for review

markdalgleish:
Even though there's some duplication now, I think this is so much better, and we can easily extract the common code into separate files where appropriate 👍

markdalgleish:
Don't forget to update the description, release notes etc. to feature the new API.

PR #66, opened 2017-09-21
chore(babel): upgrade from es2015 preset to env preset

mengtzu:
## Commit Message For Review

Replaces `babel-preset-es2015` with `babel-preset-env`

REASON FOR CHANGE:

Installing `babel-preset-es2015` currently spams an emoji-laden deprecation message, which is good for all consumers to remove.  Furthermore, at least one consumer has a use case for async/await.


markdalgleish:
Could you also update the readme?

mengtzu:
@markdalgleish readme updated

markdalgleish:
By the way, I don't think a `chore` triggers a release, this might need to be a `fix`

PR #73, opened 2017-10-12
fix(package): Upgrade dependencies, remove unused react-test-renderer

markdalgleish:
This also upgrades to node 8 to support the latest version of semantic-release.


markdalgleish:
As you can probably guess, just preparing this repo for Greenkeeper.

markdalgleish:
Yeah, that's why I'm keen on getting Greenkeeper hooked up, to encourage regular, smaller scale package upgrades with less chance of breakage.

PR #76, opened 2017-10-30
ci(travis): Remove Travis CI Node version config, use nvmrc instead

markdalgleish:
Semantic-release requires Node 8, but the nvmrc file was being ignored.


markdalgleish:
Just a heads up that the semantic-release process will fail until this PR is merged.

PR #86, opened 2017-11-13
fix(package): Upgrade babel-plugin-flow-react-proptypes to v9.0.0

markdalgleish:


markdalgleish:
There are only two extra commits to this dependency, both of which seem fine to me:
- https://github.com/brigand/babel-plugin-flow-react-proptypes/commit/0bb901a8b21f759b941540e2d6ae644c3aec52f7
- https://github.com/brigand/babel-plugin-flow-react-proptypes/commit/e15d00c25d73369c239f6604c5523e4d4c2e7204

PR #90, opened 2017-11-21
feat(eslint): Add dangerouslySetESLintConfig

dborysiewicz:
Allow consumers to override eslint config similarly to what we currently allow for jest and webpack. This is required in our use case as we wish to add a couple variables to the globals.

**sku.config.js**
```js
const dangerouslySetESLintConfig = config => {
  const eslintOverrides = {
    globals: {
      shallow: true,
      render: true
    }
  };

  return Object.assign(config, eslintOverrides);
};

module.exports = {
  ...,
  dangerouslySetESLintConfig
}
```

### DIFF
--- a/scripts/lint.js
+++ b/scripts/lint.js
@@ -1,12 +1,19 @@
 const chalk = require('chalk');
 const baseConfig = require('eslint-config-sku');
 const EslintCLI = require('eslint').CLIEngine;
+const builds = require('../config/builds');
 
 const prettierCheck = require('../lib/runPrettier').check;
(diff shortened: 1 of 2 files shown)


scripts/lint.js
```diff
@@ -1,12 +1,17 @@
 const chalk = require('chalk');
 const baseConfig = require('eslint-config-sku');
```
markdalgleish: Might be worth renaming this to `eslintConfigSku`, then you can use `baseConfig` further down the file as we do currently. Plus, I feel like this is actually clearer anyway.

PR #115, opened 2017-12-22
fix(lint): add eslint as a direct dependency

jahredhope:
Ensure eslint is available in the event that eslint-config-sku removes eslint as a dependency.

https://github.com/seek-oss/eslint-config-sku/issues/14


markdalgleish:
I'd say this needs to be a `fix` so it triggers a release.

PR #116, opened 2018-01-10
fix(package): Update eslint-config-seek to version 3.0.0

jahredhope:
As of https://github.com/seek-oss/eslint-config-seek/commit/10e9f733bf63769c5f9e87c038de92256affb340
`eslint-config-sku` has been renamed to `eslint-config-seek`.

To enable the rename a major version release was completed.
This change updates to that major version.


michaeltaranto:
You'll need to update [here](https://github.com/seek-oss/sku/blob/master/scripts/lint.js#L2) as well

PR #120, opened 2018-01-13
ci(publish): Cause build to fail on deploy error

jahredhope:
`after_success` command will not fail the build when errors occur during deployment.
This was beneficial as older versions of semantic-release caused errors to be thrown when no-release was required.
The current version of semantic-release no-longer returns an error.

This change will ensure errors, such as a misconfigured access token will show as a build error.

This change also sets the build to only deploy for `master` branch. Semantic-release will already only deploy on `master` branch by default, however it might be good to make this restriction explicit. [branch configuration semantic-release](https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/configuration.md#branch)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,15 @@
 language: node_js
+
 notifications:
   email: false
+
 before_script:
   - npm prune
-after_success:
-  - npm run semantic-release
-branches:
-  except:
-    - /^v\d+\.\d+\.\d+$/
+
+deploy:
+  provider: script
+  skip_cleanup: true
+  script:
+    - npm run semantic-release
+  on:
+    branch: master
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "precommit": "lint-staged",
     "commit": "git-cz",
     "commitmsg": "commitlint --edit --extends seek",
-    "semantic-release": "semantic-release pre && npm publish && semantic-release post"
+    "semantic-release": "semantic-release"
   },
   "lint-staged": {
     "{bin,config,scripts}/**/*.js": [
@@ -98,6 +98,6 @@
     "lint-staged": "^6.0.0",
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
-    "semantic-release": "^8.0.3"
+    "semantic-release": "^15.0.1"
   }
 }


markdalgleish:
Just in case we want to publish files generated during the build, I feel like we should always set `skip_cleanup` to `true`: https://docs.travis-ci.com/user/deployment#Uploading-Files-and-skip_cleanup

package.json
```diff
@@ -97,6 +97,6 @@
     "lint-staged": "^6.0.0",
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
-    "semantic-release": "^8.0.3"
+    "semantic-release": "^11.0.2"
```
michaeltaranto: Fyi v12 just dropped over the weekend. So maybe go straight to that

jahredhope: Thanks, upgrade to v12, however part of the v12 migration is to ensure TravisCI build only runs once.
Either with https://github.com/semantic-release/travis-deploy-once or the new **Build Stages** beta feature.
Will need to investigate this and add a follow up commit.

michaeltaranto: Yeah that used to happen by default. This is what we had to override on the internal build systems. So essentially you're re-enabling what was default behaviour in v11.

.travis.yml
```diff
@@ -3,8 +3,27 @@ notifications:
...
+  - '9'
+
+stage: 📋 Test
+
+jobs:
+  include:
+    # Publish step
+    # Only runs on master branch and only once
+    - stage: 🚀 Publish
+      script: npm run disabled
```
michaeltaranto: Is this supposed to be `npm run semantic-release`?

.travis.yml
```diff
@@ -3,8 +3,27 @@ notifications:
...
-  - npm run semantic-release
-branches:
-  except:
-    - /^v\d+\.\d+\.\d+$/
+
+# Run on multiple versions of Node
+# Will run the default stage multiple times
+node_js:
+  - '8'
+  - '9'
```
markdalgleish: I'd prefer we leave this off, otherwise `.nvmrc` won't be used in CI, which means dev and CI versions could drift.

.travis.yml
```diff
@@ -1,10 +1,13 @@
...
-  - npm run semantic-release
-branches:
-  except:
-    - /^v\d+\.\d+\.\d+$/
+
+deploy:
+  provider: script
+  skip_cleanup: true
+  script:
+    - npm run semantic-release
```
markdalgleish: Do we need to scope this to master?

```yaml
  on:
    branch: master
```

.travis.yml
```diff
@@ -1,10 +1,13 @@
...
 before_script:
   - npm prune
-after_success:
-  - npm run semantic-release
-branches:
-  except:
-    - /^v\d+\.\d+\.\d+$/
+
+deploy:
+  provider: script
```
markdalgleish: Travis CI's docs say this:

>Do note that custom scripts are still considered an experimental feature.

Any idea what they mean by this? Is it a bad idea for us to depend on this right now? 🤔

.travis.yml
jahredhope: The Semantic Release recipe for Travis CI don't include this.
https://github.com/semantic-release/semantic-release/blob/caribou/docs/recipes/travis.md
I'm not confident but was hoping to check the build logs to confirm in Travis CI.

.travis.yml
jahredhope: It's not marked in their Beta features. And seems to be quite commonly used as-well as recommended by semantic-release.
I'd be comfortable relying on the feature.

.travis.yml
jahredhope: The deployment step is already scoped to the default branch by default
> Skipping a deployment with the script provider because this branch is not permitted
https://travis-ci.org/seek-oss/sku/builds/350665289#L3740

I've added on: branch master just to be sure.

PR #122, opened 2018-02-04
feat(flow): Flow proptypes and flow type stripping

tamoore:
## Purpose

As we'll be adding flow type proptypes to the style guide, the loaders for sku
will need to be able to strip those types but also transform them to proptypes.

## Approach

Have added them to the babelConfig which I believe gets used by the webpack config. They will be needed both server side and client side.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "babel-plugin-flow-react-proptypes": "^10.0.0",
     "babel-plugin-module-resolver": "^3.0.0",
     "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-plugin-transform-flow-strip-types": "^6.22.0",
     "babel-plugin-transform-imports": "^1.4.1",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-preset-env": "^1.6.0",
--- a/config/babel/babelConfig.js
+++ b/config/babel/babelConfig.js
@@ -18,6 +18,8 @@ module.exports = ({ target }) => {
   const plugins = [
     require.resolve('babel-plugin-transform-class-properties'),
     require.resolve('babel-plugin-transform-object-rest-spread'),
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -90,6 +89,8 @@
   },
   "devDependencies": {
     "@commitlint/cli": "^6.0.1",
+    "babel-plugin-flow-react-proptypes": "^10.0.0",
+    "babel-plugin-transform-flow-strip-types": "^6.22.0",
```
markdalgleish: These need to be under `dependencies`, not `devDependencies`, because consumers are meant to install them too.

tamoore:
@markdalgleish should be updated now when you get a moment.

package.json
```diff
@@ -45,14 +45,15 @@
...
-    "babel-plugin-flow-react-proptypes": "^10.0.0",
     "babel-plugin-module-resolver": "^3.0.0",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-imports": "^1.4.1",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-preset-env": "^1.6.0",
     "babel-preset-react": "^6.24.1",
     "babel-preset-react-optimize": "^1.0.1",
+    "babel-plugin-flow-react-proptypes": "^10.0.0",
+    "babel-plugin-transform-flow-strip-types": "^6.22.0",
```
markdalgleish: Sorry to be totally pedantic, but these need to be in alphabetical order or npm will reorder them next time we install something, which adds unrelated noise to future PRs.

In my defence, it's really npm that's being pedantic here 😄

tamoore: haha! https://www.youtube.com/watch?v=BI5IA8assfk

😂

tamoore: Oh gawd, the worst part about pasting that into youtube is I'm actually seat dancing to it right now. Should be all good.

PR #123, opened 2018-02-20
feat(config): Add param to support custom config file

dborysiewicz:
Add support for custom config file via newly created `--config` parameter.

```js
sku start --config sku.custom.config.js
```

_**NOTE:** The `--config` parameter is only used for dev (`sku start`) and build steps (`sku build`). Linting (`sku lint`), formatting (`sku format`) and running of unit test (`sku test`) will still use the default config file and does **not** support it._

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,7 @@ If you need to specify a different config file you can do so with the `--config`
 $ sku start --config sku.custom.config.js
 ```
 
-_**NOTE:** The `--config` parameter is only used for dev (`sku start`) and build steps (`sku build`). Linting (`sku lint`) and unit test (`sku test`) will still use the default config file and does **not** support it._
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -298,6 +298,14 @@ module.exports = {
 }
 ```
 
+If you need to specify a different config file you can do so with the `--config` parameter.
+
+```bash
+$ sku start --config sku.custom.config.js
+```
+
+_**NOTE:** The `--config` parameter is only used for dev (`sku start`) and build steps (`sku build`). Linting (`sku lint`), formatting (`sku format`) and running of unit tests (`sku test`) will still use the default config file and does **not** support it._
```
markdalgleish: I'm interested in seeing if we can change this, but since it wouldn't be a breaking change, I don't think it needs to block this PR 👍

PR #124, opened 2018-02-22
feat(build): Support custom file output

jahredhope:
# Custom file naming
`sku build` currently hard-codes asset names according to the naming convention '[name].js' and 'style.css'.

This change allows the consumer to provide a name that includes '[chunkHash]' or other values to support more fine-grain cache controls.

The new behaviour is optional, and if not set the default behaviour remains the same. Later changes may work to change the default behaviour after a migratory period.

### Example
**File level versioning**
``` javascript
const skuConfig = {
  outputFilename: '[name]-[chunkhash].js',
  outputCssFilename: '[name]-[chunkhash].css',
  publicPath: isProductionBuild ? 'https://seekcdn.com/fooSite/barApp/' : '/'
};
```

**Folder level versioning**
``` javascript

const UI_VERSION = process.env.BUILD_NUMBER
const skuConfig = {
  outputFilename: '[name].js',
  outputCssFilename: '[name].css',
  publicPath: isProductionBuild ? `https://seekcdn.com/fooSite/barApp/${UI_VERSION}/` : '/'
};
```

# Filenames with multiple webpack builds
Currently `sku` consumers hardcode a script name, such as `main.js` into the render script, this cannot be done with dynamic asset naming.

This change combines the currently separated 'client' an 'render' configurations into a single configuration. This allows the 'render' function to have access to the assets produced, allowing the direct use of the file names.

**Note:** This change affects out ability to support `process.env.{key}` variables such as SKU_ENV. A backwards compatible fix has not been found yet.

### Example
``` javascript
export default ({ assets }) => {
  return `<!DOCTYPE html>
    <html>
      <body>
        <div id="app">${appHtml}</div>
        <script type="text/javascript" src="/${assets.client}"></script>
      </body>
    </html>`;
}
```

# API Changes

This change involves proposing the following additional parameters available for sku.config

Documented in README.md:
- outputFilename
- outputCssFilename
- publicPath

Undocumented / not currently recommended behaviour:
- historyApiFallback

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -103,7 +103,17 @@ const svgLoaders = [
 ];
 
 const buildWebpackConfigs = builds.map(
-  ({ name, paths, env, locales, webpackDecorator, port }) => {
+  ({
+    name,
+    paths,
+    env,
+    locales,
+    webpackDecorator,
+    port,
+    outputFilename,
+    outputCssFilename,
+    publicPath
+  }) => {
     const envVars = lodash
       .chain(env)
       .mapValues((value, key) => {
@@ -131,23 +141,17 @@ const buildWebpackConfigs = builds.map(
 
     const internalJs = [paths.src, ...paths.compilePackages];
 
-    const entry = [paths.clientEntry];
-    const devServerEntries = [
-      `${require.resolve(
-        'webpack-dev-server/client'
-      )}?http://localhost:${port}/`
-    ];
-
-    if (args.script === 'start') {
-      entry.unshift(...devServerEntries);
-    }
-
     return [
       {
-        entry,
+        entry: {
+          main: paths.clientEntry,
+          render: paths.renderEntry
+        },
         output: {
           path: paths.dist,
-          filename: '[name].js'
+          filename: outputFilename,
+          publicPath,
+          libraryTarget: 'umd'
         },
         module: {
           rules: [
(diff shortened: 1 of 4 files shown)


scripts/start.js
```diff
@@ -4,12 +4,12 @@ const webpackConfig = require('../config/webpack/webpack.config');
...
 const opn = require('opn');
 
-const { hosts, port } = builds[0];
+const { hosts, port, historyApiFallback } = builds[0];
 
 const compiler = webpack(webpackConfig);
 const devServer = new WebpackDevServer(compiler, {
   contentBase: builds.map(({ paths }) => paths.public),
-  historyApiFallback: true,
+  historyApiFallback: historyApiFallback ? historyApiFallback : true,
```
nicolassenechal: Not sure that the logic is correct here. What are the expected values for historyApiFallback?

config/webpack/webpack.config.js
```diff
@@ -131,23 +141,17 @@ const buildWebpackConfigs = builds.map(
...
-
-    if (args.script === 'start') {
-      entry.unshift(...devServerEntries);
-    }
-
     return [
       {
-        entry,
+        entry: {
+          client: paths.clientEntry,
```
nicolassenechal: Interesting refactor here. Definitely easier to read, makes the difference between static and server-side rendering more pronounced unfortunately.

scripts/start.js
jahredhope: The usecase may be dependant on how you serve files in production. I'm away from desk but can try and hand write an example:
`{rewrite: {from: /./, to: '404'}}`
It would be good to agree on a pattern rather than allowing custom config, but not sure what Configs there would be.

config/webpack/webpack.config.js
jahredhope: I wouldn't be surprised if we should make the same changes to SSR, but as I haven't used sku for SSR so it's hard to test.

jahredhope:
This can't go in as is. It would not support some `target: node` required behaviours.
Will look for another solution and create another PR.

PR #126, opened 2018-03-06
fix(less): Ensure only a single set of style loaders is used

jahredhope:
Currently sku creates a webpack configuration which loads loaders to handle .less files via multiple configurations.
The configurations all use exclusive include or exclude values that should mean no two set of loaders are loaded at one time.

Unfortunately, due to complexity in testing with symlink directories (created by `npm link` or `yarn link`) combined with the Babel's Resolve plugin it is possible for multiple sets of loaders to run, causing a build error.

[oneOf](https://webpack.js.org/configuration/module/#rule-oneof) will ensure a single set of style loaders is used at one time.

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -178,20 +178,23 @@ const buildWebpackConfigs = builds.map(
             },
             {
               test: /\.less$/,
-              exclude: /node_modules/,
-              use: ExtractTextPlugin.extract({
-                fallback: 'style-loader',
-                use: makeCssLoaders()
-              })
+              oneOf: [
+                ...paths.compilePackages.map(packageName => ({
+                  include: packageName,
+                  use: ExtractTextPlugin.extract({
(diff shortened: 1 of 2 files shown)


config/webpack/webpack.config.js
```diff
@@ -178,20 +178,24 @@ const buildWebpackConfigs = builds.map(
...
             {
               test: /\.less$/,
-              exclude: /node_modules/,
-              use: ExtractTextPlugin.extract({
-                fallback: 'style-loader',
-                use: makeCssLoaders()
-              })
+              oneOf: [
+                ...paths.compilePackages.map(packageName => ({
+                  test: /\.less$/,
```
michaeltaranto: This test is redundant now right?

jahredhope: Correct. All 4 inner tests were redundant and have now been removed and re-tested. Thanks

PR #128, opened 2018-03-09
feat(init): Create init script from create-seek-ui

jahredhope:
There is currently a set of tooling used to create a `sku` project.
The tooling was already functional. This change pulls in the tooling into the `sku` project under `sku init`.

This command will create a new project that uses `sku`, `seek-style-guide` and many best-practice choices.

EXAMPLE USAGE:

```js
$ npx sku init my-app
$ cd my-app
$ npm start
```


markdalgleish:
I'm imagining that in the future we'll have different presets, depending on the type of application you're trying to build. I wonder how much we should future-proof this command a little bit? 🤔

.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules/
 npm-debug.log
+yarn-error.log
```
markdalgleish: Maybe we should just change this to `*.log`?

jahredhope: I don't mind calling out these files explicitly, but I'm not fussed. I'll change to `*.log`, we can always change it in the future.

jahredhope:
Discussed separately with @markdalgleish.
I think that as implemented `sku init` doesn't support maintaining/upgrading existing frameworks making it easier to make breaking changed.
We'll have to re-address this assumption if we looked at supporting something like `sku doctor` where we upgrade/maintain existing `sku` builds.

PR #129, opened 2018-03-13
fix(start): Allow webpack dev server to be accessible via ip

dborysiewicz:
Switching this allows it to be access via `127.0.0.1` still but also your machines ip address, this is useful for if you want to showcase or test (IE11 in a VM) your work on your local machine. 

https://github.com/webpack/webpack-dev-server/issues/147


dborysiewicz:
None I'm aware of 👍

PR #131, opened 2018-03-20
fix(ssr): Add missing react-hot-loader dependency

markdalgleish:
I was improving the test suite, adding tests for our (currently-undocumented) SSR mode, but discovered that `react-hot-loader` was missing when running a local dev server. This wasn't picked up earlier because the consuming application had installed it as a direct dependency.

As a result, while this PR fixes the issue, it also includes a bunch of test changes. We're now testing that both development and production modes work for basic static sites and server rendered apps, which is a huge improvement 🙌


test/test-cases/hello-world/hello-world.test.js
```diff
@@ -0,0 +1,50 @@
...
+      const response = await fetch(devServerUrl);
+      const responseText = await response.text();
+      expect(responseText).toMatchSnapshot();
+    });
+  });
+
+  describe('build', () => {
+    beforeAll(async () => {
+      await rimrafAsync(`${__dirname}/dist/`);
+      await exec(`node ${__dirname}/../../../scripts/build`, { cwd: __dirname, stdio: 'inherit' });
```
michaeltaranto: I would suggest storing this as a variable `${__dirname}/../../../scripts` and possibly `${__dirname}/dist/` as well, just to make future maintenance easier

michaeltaranto: In fact given they are used across files, maybe something in `config` or in the `utils` section

test/test-cases/ssr-hello-world/ssr-hello-world.test.js
```diff
@@ -0,0 +1,72 @@
...
+      await waitOnAsync({ resources: [backendUrl.replace(/^http/, 'http-get')], timeout: 10000 });
+    });
+
+    afterAll(() => {
+      server.kill();
+    });
+
+    it('should generate a production server', async () => {
+      const response = await fetch(backendUrl);
+      const responseText = await response.text();
```
michaeltaranto: is `response.text()` an async call?

markdalgleish: Yup! https://developer.mozilla.org/en-US/docs/Web/API/Body/text

test/test-cases/ssr-hello-world/ssr-hello-world.test.js
```diff
@@ -0,0 +1,72 @@
...
+      
+      await waitOnAsync({
+        resources: [
+          backendUrl.replace(/^http/, 'http-get'),
+          clientJsUrl.replace(/^http/, 'http-get')
+        ],
+        headers: {
+          accept: 'text/html, application/javascript'
+        },
+        timeout: 20000
```
michaeltaranto: There are a few timeouts littered around, should they be uniform or in a config? Or are they deliberately set separately?

test/test-cases/ssr-hello-world/ssr-hello-world.test.js
michaeltaranto: Woah! Didnt know that existed. As you were... 👏

test/test-cases/ssr-hello-world/ssr-hello-world.test.js
markdalgleish: Yeah, good idea, these should be centralised somewhere. I'd probably solve this at a higher level and make a `waitForUrls` util.

hscgavin: Just did `npm run test` 8 times, 2 of them failed on ssr-hello-world.test.js because of the timeout error(took 24-ish seconds), the other 6 tests finished in 11 seconds. Do you think we can increase it to  higher value `25000` or pass a cb in waitOn to handle the error?

```
 FAIL  test/test-cases/ssr-hello-world/ssr-hello-world.test.js (24.381s)
```
```
Timeout
      at node_modules/wait-on/lib/wait-on.js:92:10
```
```
FetchError: request to http://localhost:8001/ failed, reason: connect ECONNREFUSED 127.0.0.1:8001
```

markdalgleish:
@michaeltaranto @hscgavin Updated the PR with a few PRs, addressing the points you raised. Let me know what you think 👍🏻

PR #132, opened 2018-03-21
test(manual): Add manual test script

markdalgleish:
This allows you to easily start or build any of the test cases. I've even added a nifty little disk-based cache in there which remembers your last choice, making it a bit easier to iterate on a single test case.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,13 @@
+# editorconfig.org
```
nicolassenechal: Was this file supposed to be included?

markdalgleish: Yep, it was, but forgot to call it out in the PR description. Happy to keep it in?

markdalgleish:
Looked into documenting this, but we currently don't have a good way to documenting npm scripts, beyond making them self-documenting. This one seems relatively straightforward enough to me—happy to leave as-is?

PR #134, opened 2018-04-10
feat(async): Support async/await, include Promise polyfill

markdalgleish:
For this to work in old browsers, we need to include [regenerator-runtime](https://www.npmjs.com/package/regenerator-runtime) and [promise-polyfill](https://www.npmjs.com/package/promise-polyfill). In the future, we should investigate how we can optimise this by only including polyfills for browsers that need them, but this feature is needed urgently.

Please note that, if you're already including a promise polyfill in your application code, this should now be removed.


markdalgleish:
Upon further discussion, deciding against this approach for now, to reduce polyfill/runtime bloat.

PR #141, opened 2018-05-25
feat(start): Add initialPath param to config

georgespyropoulos:
Allow an initial path to be set so a specific route can be loaded when running `sku start`

EXAMPLE USAGE:

`sku.config.js`

```js
module.exports = {
  entry: {
   ...
  initialPath: '/my-page'
  ...
};
```

### DIFF
--- a/scripts/start-ssr.js
+++ b/scripts/start-ssr.js
@@ -6,7 +6,7 @@ const opn = require('opn');
 const Promise = require('bluebird');
 const webpackPromise = Promise.promisify(require('webpack'));
 const fs = require('fs-extra');
-const { hosts, port } = builds[0];
+const { hosts, port, initialPath } = builds[0];
 const serverEntry = builds[0].paths.serverEntry;
 
 const compiler = webpack(serverEntry ? webpackConfig[0] : webpackConfig);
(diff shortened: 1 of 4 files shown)


scripts/start-ssr.js
```diff
@@ -62,7 +62,7 @@ const copyPublicFiles = () => {
 runWebpack(webpackConfig[1])
   .then(copyPublicFiles)
   .then(() => {
-    const url = `http://${hosts[0]}:${port.backend}`;
+    const url = `http://${hosts[0]}:${port.backend}${initialPath}`;
```
jahredhope: Should this use a [url joining library](https://www.npmjs.com/package/url-join) to handle missing/duplicate slashes?

markdalgleish:
Not sure a CLI argument makes sense for this one, since this is a config value that might differ for every project in a monorepo setup, e.g.

```js
module.exports = [
  {
    ...,
    name: 'my-first-page',
    initialPath: '/my-first-page'
  },
  {
    ...,
    name: 'my-second-page',
    initialPath: '/my-second-page'
  }
];
```

georgespyropoulos:
I still prefer having this option in `sku.config.js` but the consensus at a FEED meeting was to have it as a CLI argument. 

I think the argument against having it in the config was that people felt that SKU should have minimal config and the initial path could easily be set in a package.json script. 

I think that a CLI argument fragments the configuration options for SKU. They're no longer all together in the sku config.

hscgavin:
`builds.ssr.js` has been forgotten? 🤔

PR #147, opened 2018-07-17
feat(docs): Added SSR documentation

nicolassenechal:
This PR adds documentation for SSR, making it a supported feature of sku.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,7 +385,7 @@ $ npm start hello
 
 The default mode for sku is to statically render projects. However, Server-Side Rendering (SSR) can explicitly be turned on, both in development with hot module reloading for React, and in production.
 
-First, you need to create a `sku.config.js` file, which will contain the following setup at minima:
+First, you need to create a `sku.config.js` file, which will contain the following setup at minimum:
 ```js
 module.exports = {
   entry: {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -381,6 +381,55 @@ Alternatively, you can start the relevant project directly:
...
+    client: 'src/client.js',
+    server: 'src/server/server.js'
+  },
+  public: 'src/public',
+  target: 'dist/',
+  port: { client: 3300, backend: 3301 }
+}
+```
+
+Then, you need to create your `server` entry. Sku will automatically provide an Express server for the user. The entry point for SSR, `server`, is used to provide a render callback and any additional middlewares to that server. You can provide either a single middleware or an array. This can be done as follows:
```
markdalgleish: Might be a good idea to make "Express" a link to the Express documentation.

README.md
```diff
@@ -381,6 +381,55 @@ Alternatively, you can start the relevant project directly:
 $ npm start hello
 ```
 
+### Server-Side Rendering Support
+
+The default mode for sku is to statically render projects. However, Server-Side Rendering (SSR) can explicitly be turned on, both in development with hot module reloading for React, and in production.
+
+First, you need to create a `sku.config.js` file, which will contain the following setup at minima:
```
markdalgleish: Typo—should it be `at a minimum`?

README.md
```diff
@@ -381,6 +381,55 @@ Alternatively, you can start the relevant project directly:
...
+Then, you need to create your `server` entry. Sku will automatically provide an Express server for the user. The entry point for SSR, `server`, is used to provide a render callback and any additional middlewares to that server. You can provide either a single middleware or an array. This can be done as follows:
+```js
+import render from './render.js';
+import middleware from './middleware';
+
+export default {
+    renderCallback: (req, res) => {
+        res.send(render());
+    },
+    middleware: middleware
```
markdalgleish: Nitpick: could these examples use 2 spaces for indenting?

README.md
```diff
@@ -381,6 +381,55 @@ Alternatively, you can start the relevant project directly:
...
+```js
+module.exports = {
+  entry: {
+    client: 'src/client.js',
+    server: 'src/server/server.js'
+  },
+  public: 'src/public',
+  target: 'dist/',
+  port: { client: 3300, backend: 3301 }
+}
```
michaeltaranto: Given `sku init` provides you with a config file, is it worth making this a diff of what properties you are adding to the config to support server rendering?

markdalgleish:
@nicolassenechal Any word on this?

PR #151, opened 2018-07-28
fix(jest): Add testURL Jest config

mbfisher:
Preemptive defence against https://github.com/facebook/jest/issues/6766.

### DIFF
--- a/config/jest/jestConfig.js
+++ b/config/jest/jestConfig.js
@@ -11,5 +11,6 @@ module.exports = {
   },
   transform: {
     '^.+\\.js$': require.resolve('./babelTransform.js')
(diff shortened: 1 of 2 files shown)


markdalgleish:
Nice one :clap: 

The scope for this should be `fix` so it triggers a release.

mbfisher:
Why would we release changes to internal test config?

Should we be adding this to `config/jest/jestConfig.js` do you think? That would be a `fix`.

markdalgleish:
Oh, sorry, I misunderstood this PR. I assumed you were fixing this for consumers. Yeah, we should probably do that too.

PR #125, opened 2018-03-02
feat(build): Upgrade Webpack to version 4

syakiruzair:
Upgrade Webpack to version 4. Tested on fresh Candidate module of Empleolisto and Jofom. We dont mind to wait for stable version of `extract-text-webpack-plugin` as we are using my fork for the time being.

BREAKING CHANGE:

As per testing, there is no need for any change on the code base but NodeJS v6.11.5 or above is needed as required by latest version Webpack Dev Server.

TODO:

- [x] Merge https://github.com/syakiruzair/sku/pull/1
- [ ] Fix `Error: Can't resolve '__sku_alias__serverEntry'` on `start-ssr`

PR #152, opened 2018-07-30
feat(build): Upgrade to Webpack v4

mbfisher:
BREAKING CHANGE:

- Requires NodeJS v6.11.5.
- Webpack config schema has breaking changes which `dangerouslySetWebpackConfig` users should review.
- Replaces `extract-text-webpack-plugin` with `mini-css-extract-plugin`.

re #125


jahredhope:
Might be worth reviewing some of the dependencies after this PR, they shouldn't be a problem but worth knowing we are a bit behind on a few things.
screen shot 2018-08-09 at 11 52 13 am

mbfisher:
@jahredhope I was purposefully conservative on updates - don't want to break any more than we have to!

mbfisher:
On the `start-ssr` item; I'm getting 2 issues running it in my project:

image

1. `http://talent.seek.com.au.local:undefined`
2. Can't resolve `__sku_alias__serverEntry`

mbfisher:
However, I get the same with `sku@latest`, so maybe it's something I'm doing in my project. The E2E test is passing...

image

markdalgleish:
@mbfisher have you defined `entry.server` in your sku config?

mbfisher:
@markdalgleish No 😅. Now I've dug into it I hadn't done any of the things I needed to! With a basic `serverEntry` the alias error goes away 👍.

mbfisher:
Build's failing on CI though 🤦‍♂️

mbfisher:
I think we're getting intermittent CI failures because the tests are starting to take around 35s to run, so I've bumped up `jest.setTimeout`.

config/webpack/webpack.config.js
```diff
@@ -136,7 +137,10 @@ const buildWebpackConfigs = builds.map(
       .value();
 
     const internalJs = [paths.src, ...paths.compilePackages];
-    const entry = [paths.clientEntry];
+    const resolvedPolyfills = polyfills.map(polyfill => {
+      return require.resolve(polyfill, { paths: [process.cwd()] });
+    });
+    const entry = [...resolvedPolyfills, paths.clientEntry];
```
markdalgleish: Is this change related to the webpack 4 upgrade? Seems unrelated—do we know why this was changed?

syakiruzair: I cannot remember this change, @mbfisher . But if I did, it should be to clean it up as what @markdalgleish mentioned.

mbfisher: 👍

hscgavin: 🤔 how come `entry = [...devServerEntries, ...resolvedPolyfills, paths.clientEntry]`?
should be `entry = [ ...resolvedPolyfills, ...devServerEntries, paths.clientEntry]` 
☝️  @markdalgleish

mbfisher: @hscgavin good point

hscgavin: the order here is important 👇 

>  it's important for polyfills that modify the global environment to be loaded before all other code. To address this, the polyfills option allows you to provide an array of modules to import before any other code is executed.

mbfisher:
Get very inconsistent test results with timeouts now - the current commit has passed the [PR build](https://travis-ci.org/seek-oss/sku/builds/414353415) but failed the [push build](https://travis-ci.org/seek-oss/sku/builds/414353444).

PR #153, opened 2018-08-01
fix(images): Make client and server URLs consistent

tgdev:
REASON:

The current webpack config provides different limit values for server vs client when it comes to loading images via the `url-loader`. This is causing differing behaviour between the two environments during development on local machine however does not seem to impact files once pushed to production.

A specific example that I can provide is for my teams new vertical where we have a large background image on the homepage.

This image is imported into our HeroBanner component

```
import bannerImage from './hero-cafe-counter-2x.jpg';
...

class HeroBanner extends Component {
  ...
  render() {
    return (
      <div className={classes} style={{backgroundImage: `${bannerImage}`}}> ... </div>
    )
  }
```
Upon recompiling, webpack is inlining this image as a different DataURL in the browser (client.js) compared to the DataURL on the server (render.js' rendered HTML).

The result is the following warning displayed in the console:

screen shot 2018-08-01 at 10 46 30 am

The change that's been made is to set a single byte size limit for the `url-loader` to the smaller `10000 or 10Kb`. This means any image over this size will be linked to rather than inlined as `DataURL`.

We are also taking advantage of the `file-loaders` emitFile option to prevent unnecessary output of image files on the server.

NOTE: During the development of this fix it was discovered that the `file-loader` package which is a required fallback to `url-loader` was not installed as part of this repo's `package.json, so that too has been rectified.

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -75,7 +75,9 @@ const makeImageLoaders = (options = {}) => {
     {
       loader: require.resolve('url-loader'),
       options: {
-        limit: server ? 999999999 : 10000
+        limit: 10000,
+        fallback: require.resolve('file-loader'),
+        emitFile: !server
       }
     }
   ];
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -38,6 +38,8 @@ const makeCssLoaders = (options = {}) => {
 
   return (cssLoaders = [
     {
+      // On the server, we use 'css-loader/locals' to avoid generating a CSS file.
+      // Only the client build should generate CSS files.
       loader: require.resolve(`css-loader${server ? '/locals' : ''}`),
       options: {
         modules: true,
@@ -77,6 +79,8 @@ const makeImageLoaders = (options = {}) => {
       options: {
         limit: 10000,
         fallback: require.resolve('file-loader'),
+        // We only want to emit client assets during the client build.
+        // The server build should only emit server-side JS and HTML files.
         emitFile: !server
       }
     }
--- a/config/webpack/webpack.config.ssr.js
+++ b/config/webpack/webpack.config.ssr.js
@@ -78,7 +78,9 @@ const makeImageLoaders = (options = {}) => {
     {
       loader: require.resolve('url-loader'),
       options: {
-        limit: server ? 999999999 : 10000
+        limit: 10000,
+        fallback: require.resolve('file-loader'),
+        emitFile: !server
       }
     }
   ];
--- a/config/webpack/webpack.config.ssr.js
+++ b/config/webpack/webpack.config.ssr.js
@@ -41,6 +41,8 @@ const makeCssLoaders = (options = {}) => {
   return (cssLoaders = [
     ...(isProductionBuild || server ? [] : ['style-loader']),
     {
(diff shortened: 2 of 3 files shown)


markdalgleish:
I like the query parameter idea—it seems like the most pragmatic way to allow consumers to decide how the import should be handled on a case-by-case basis.

In this case, my vote would be for extracting files by default, and allowing for an `?inline` query parameter.

config/webpack/webpack.config.js
```diff
@@ -266,7 +266,7 @@ const buildWebpackConfigs = builds.map(
 
             {
               test: [/\.bmp$/, /\.gif$/, /\.jpe?g$/, /\.png$/],
```
tgdev:
```
// default to not limit images
 test: [/\.bmp$/, /\.gif$/, /\.jpe?g$/, /\.png$/],
    use: makeImageLoaders({ limit: false));

// limit/embed as dataURL if inline query param exists
 test: [/\.bmp?inline$/, /\.gif?inline$/, /\.jpe?g?inline$/, /\.png?inline$/],
    use: makeImageLoaders({ limit: true));

```

config/webpack/webpack.config.js
```diff
@@ -68,14 +68,14 @@ const makeCssLoaders = (options = {}) => {
   ]);
 };
 
-const makeImageLoaders = (options = {}) => {
-  const { server = false } = options;
+const makeImageLoaders = () => {
+  const imageFileSizeLimit = 10000; // 10Kb
```
tgdev:
```
const {limit = false } = options;

const imageFileSizeByteLimit = limit ? 1000 : 9999999999; 
``

File: config/webpack/webpack.config.js [hunk 3]
```diff
@@ -75,7 +75,9 @@ const makeImageLoaders = (options = {}) => {
     {
       loader: require.resolve('url-loader'),
       options: {
-        limit: server ? 999999999 : 10000
+        limit: 10000,
+        fallback: require.resolve('file-loader'),
+        emitFile: !server
```
jahredhope: This seems dangerous.
Isn't this relying on the the client code to also importing the image?
If for any reason the file is not loaded client side then it would be a returning a url to a file that doesn't exist.

Is there any harm in outputting the file?
If the files client/render are the same then emitting the file will make no difference.
If the files client/render are different then not emitting the file would resulting in a 404.

markdalgleish: This is what the existing logic was trying to achieve, by setting a limit of `999999999` on the server, effectively inlining every image.

Server code shouldn't be referencing images, only code shared across client and server.

jahredhope: Fair enough. We'll need to revisit this if anyone wants to create a static-only page (no client at all), or a static-only image (perhaps some `<noscript>` type scenario), but what you've got should work in the meantime.

Could we add some comment describing why we don't need to emit the file? For any future developers who are debugging this.

tgdev: @jahredhope I'm a little confused here. Aren't our new verticals for invoicing and shopfront 'static pages' in that the static html files are build via sku and the renderToString functionality of react before it's deployed to and served from S3?

If I turn off JS on shopfront so that the client JS doesn't rehydrate I'd still be able to display the images from seekcdn via the static html page.

Anyway, comments have been added :-)

PR #154, opened 2018-08-23
fix(babel): Replace babel-preset-react-optimize

mbfisher:
`babel-preset-react-optimize` applies [4 plugins](https://github.com/jamiebuilds/babel-react-optimize/blob/master/packages/babel-preset-react-optimize/src/index.js#L1), but an in "incorrect" order. This change replaces the preset with the plugins in the correct order as described in https://github.com/babel/babel/issues/5325#issuecomment-292064058. A PR against the preset would be nice, but this fixes it in the mean time.


mengtzu:
Can you expand a little on this - are those four plugins a spanning set of the preset?

mbfisher:
@mengtzu description updated.

michaeltaranto:
@mbfisher i would've expected to see this change reflected in the `dependencies` too. Given this will be [fixed in babel 7](https://github.com/babel/babel/pull/5686), maybe it's better we remove the preset and depend directly on the plugins. Thoughts?

mbfisher:
Good point @michaeltaranto, done.

By `npm install`ing the 4 plugins, we jump from `^0.2.5` to `^0.4.15` of `babel-plugin-transform-react-remove-prop-types`, compared to the deps of `babel-preset-react-optimize`. The release notes for [`0.3.0`](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/releases/tag/v0.3.0) and [`0.4.0`](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/releases/tag/v0.4.0) don't include anything concerning IMO.

PR #155, opened 2018-08-31
fix(webpack): Don't compile external modules to commonjs

mattsjones:
This PR stops webpack seeing all node_modules as commonjs. 

Webpack can perform better optimization like tree shaking and concatenation on es6 modules. Before now babel was set to compile all these modules to commonjs before webpack could optimize them. This should be a nice win across the board without too much risk.


mbfisher:
Does this achieve the same result as adding this rule?

```
{
  test: /\.mjs$/,
  include: /node_modules/,
  type: 'javascript/auto'
}
```

@kosanna have `dangerouslySet` this to handle `graphql`, which is distributed as ESM.

mattsjones:
This is a different issue. Webpack favours the `module` field in `package.json`. It's essentially the equivalent of `main` but it should point to es6 modules rather then commonjs. The `.mjs` extension is not required for those files.

PR #157, opened 2018-08-31
feat(format): Upgrade prettier version

jahredhope:
This change upgrades prettier to the current version released 22 days ago. It also updates `lint-staged` to address issues with commit checks, required to commit this change.

We currently lock the version of prettier to ~1.9.0. Any change to prettier may modify whitespace requirements and can break a consumer's build.

A recent release from Atom's apm prettier package [requires versions of prettier above v1.13.4](https://atom.io/packages/prettier-atom#problems-with-latest-release-v0550).
There are reports of other IDEs having trouble with the older version supplied by sku.

Minor changes to formatting may break `sku lint`. Please run `sku format` to reformat your code.


package.json
```diff
@@ -90,7 +90,7 @@
     "opn": "^5.3.0",
     "postcss-js": "^1.0.1",
     "postcss-loader": "^2.1.6",
-    "prettier": "~1.9.0",
+    "prettier": "1.14.2",
```
mbfisher: Why pin this? Do we not trust their semantic versioning?

Is it bad if minor releases of Prettier include rule changes that break linting? Isn't that what `sku format` is for?

jahredhope: Changes to prettier version break the `sku lint` command, breaking consumer builds.

We don't have to lock it but it might negatively hurt the reliability of sku as a product if consumers find that one day their builds pass and one day it doesn't.

Will discuss offline and see what comes of it.

mbfisher:
The build hung (as it did for me on #156), so running again.

PR #160, opened 2018-09-06
fix: Passing args to lint and format commands

mbfisher:
Fixes an error where `lint` and `format` are being passed to `eslint` and `prettier` as file paths e.g.

```
[error] No matching files. Patterns tried: lint !**/node_modules/** !./node_modules/** !**/.{git,svn,hg}/** !./.{git,svn,hg}/**
```

### DIFF
--- a/scripts/lint.js
+++ b/scripts/lint.js
@@ -9,7 +9,9 @@ const prettierConfig = require('../config/prettier/prettierConfig');
 const prettierDefaultPath = require('../config/prettier/defaultPath');
 const eslintDefaultPath = require('../config/eslint/defaultPath');
 
-const args = process.argv.slice(2);
+const args = process.argv.slice(3);
+
+console.log(chalk.cyan('Linting'));
 
 // Decorate eslint config is not supported for monorepo
 const eslintConfig =
@@ -25,9 +27,9 @@ const cli = new EslintCLI({
 const pathsToCheck = args.length === 0 ? [eslintDefaultPath] : args;
 
 const formatter = cli.getFormatter();
+console.log(chalk.gray(`eslint ${pathsToCheck.join(' ')}`));
 const report = cli.executeOnFiles(pathsToCheck);
 
-console.log(chalk.cyan('Linting'));
 console.log(formatter(report.results));
 
 if (report.errorCount > 0) {
--- a/lib/runPrettier.js
+++ b/lib/runPrettier.js
@@ -1,6 +1,6 @@
 const spawn = require('cross-spawn');
 const path = require('path');
-
+const chalk = require('chalk');
 const cwd = process.cwd();
 
 /*
@@ -22,7 +22,7 @@ const runPrettier = ({ write, listDifferent, config, filePattern }) => {
 
   const prettierBinPath = getPrettierBinLocation();
 
-  const prettierArgs = [];
+  let prettierArgs = [];
 
   if (prettierConfig.singleQuote) {
     prettierArgs.push('--single-quote');
@@ -34,6 +34,8 @@ const runPrettier = ({ write, listDifferent, config, filePattern }) => {
     prettierArgs.push('--list-different');
   }
 
+  prettierArgs = prettierArgs.concat(filePattern);
+
   /*
    * Show Prettier output with stdio: inherit
    * The child process will use the parent process's stdin/stdout/stderr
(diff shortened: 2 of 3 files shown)


markdalgleish:
Why did the tests not pick this up? 🤔

mbfisher:
**This fix only works if you use `sku [lint|format]`**

These examples log `process.argv`:

```
➜ sku lint 'src/components/**'
process.argv [ '/path/to/node',
  '/path/to/sku',
  'lint',
  'src/components/**' ]
Linting
eslint src/components/**
...
Checking Prettier format rules
prettier --single-quote --list-different src/components/**
```

```
➜ node node_modules/sku/scripts/lint.js 'src/components/**'
process.argv [ '/path/to/node',
  '/path/to/node_modules/sku/scripts/lint.js',
  'src/components/**' ]
Linting
eslint src
...Checking Prettier format rules
prettier --single-quote --list-different src/**/*.js
```

In the second example console logs of the commands being run contain the default file patterns, not the pattern passed as a CLI arg.

I'd suggest that we need a better way to get CLI args than `argv.slice(n)`.

markdalgleish:
> This fix only works if you use sku [lint|format]

Sorry—as opposed to what?

scripts/lint.js
```diff
@@ -9,7 +9,9 @@ const prettierConfig = require('../config/prettier/prettierConfig');
 const prettierDefaultPath = require('../config/prettier/defaultPath');
 const eslintDefaultPath = require('../config/eslint/defaultPath');
 
-const args = process.argv.slice(2);
+const args = process.argv.slice(3);
```
markdalgleish: Am I right in thinking this PR could be reduced to these two `slice` lines? Do we need to keep the other changes?

test/utils/runSkuScriptInDir.js
```diff
@@ -1,14 +1,14 @@
 const spawn = require('./gracefulSpawn');
 const { exec } = require('child-process-promise');
 
-const scriptsPath = `${__dirname}/../../scripts/`;
+const skuBin = `${__dirname}/../../bin/sku.js`;
```
markdalgleish: Oh, awesome—I'd made the same change locally and was about to add it to your PR. Should have guessed you're way ahead of me 😄

package.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "sku",
   "version": "0.0.0-development",
-  "description": "Front-end development toolkit, powered by Webpack, Babel, CSS Modules, Less and Jest",
+  "description":
```
markdalgleish: It looks like you didn't change this file, and it's just been formatted with Prettier—is that right?

markdalgleish: I'm not sure we want to format package.json with prettier, since npm already enforces its own formatting via JSON.stringify when installing packages.

lib/parseArgs.test.js
```diff
@@ -0,0 +1,55 @@
+const parseArgs = require('./parseArgs');
+
+describe('arg parsing', () => {
```
markdalgleish: 👏👏👏

lib/parseArgs.js
```diff
@@ -0,0 +1,91 @@
...
+  const options = commandLineArgs(optionDefinitions, {
+    argv: args,
+    stopAtFirstUnknown: true,
+    partial: true
+  });
+
+  let { scriptName, _unknown: argv = [] } = options;
+
+  // Detect if you're trying to execute a script directly...
+  const scriptNameRegex = /scripts[\/\\]([\w-]*)\.js$/i;
```
markdalgleish: Is this something we want to support? Why do we need this?

mbfisher: Good question! I'm happy to enforce using the `sku` executable; my first go at this did exactly that. However it's the same amount of work to detect and adapt as it is to prevent, and I figured that's more flexible.

I don't have any evidence for users running scripts directly, the only example was directly was in our tests, which I've updated here too.

package.json
mbfisher: Whoops, `formatOnSave`!

lib/parseArgs.js
markdalgleish: IMHO I think we're better off restricting it, just to keep things more predictable.

lib/parseArgs.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {String[]} args - should look like process.argv
+ */
+module.exports = _args => {
+  const args = [..._args];
+
+  // The first argv is always the node executable, which we don't care about.
+  args.shift();
+
+  // If you're running `sku something` then the next arg is the sku executable, which we don't
+  // care about either.
```
markdalgleish: Loving the usage of comments here—sku really needs more of this.

lib/parseArgs.js
mbfisher: After a chat @markdalgleish and I decided to not support the `node scripts/lint.js` API.

PR #163, opened 2018-09-13
feat(test): Mock CSS Modules for .less and .css.js in tests

markdalgleish:
## Description

When importing `.less` and `.css.js` files in tests, you'll now get realistic mock CSS Modules objects.

## Migration Guide

Your snapshot tests are likely to fail due to different classes. You can quickly update your snapshots with this command (assuming `npm test` runs `sku test`):

```
$ npm test -- -u
```

Apart from snapshots, if you've manually referenced CSS class names in your tests, they will need to be prefixed with the file name and two underscores, e.g. when selecting by class name with [Enzyme](https://github.com/airbnb/enzyme):

```diff
-wrapper.find('.root')
+wrapper.find('.Text__root')
```

## Why?

We currently mock CSS Modules with [identity-obj-proxy](https://www.npmjs.com/package/identity-obj-proxy), which simply echoes back the property name.

As a result, invalid classes can slip through undetected, e.g. `styles.this_class_does_not_exist` will return `'this_class_does_not_exist'`.

It also means you can't enumerate classes in your app code because `Object.keys(styles)` returns an empty array.

When it comes to `.css.js` files, we don't do any mocking at all, which means class names resolve to `undefined` in tests since they don't exist on the exported object.

## Development Notes

I've also added tests to ensure that `sku test` gets the expected mocks when consuming styles.


markdalgleish:
This change is likely to cause consumers' test suites to fail, particularly if they're using snapshot tests. Do we consider this a breaking change?

mengtzu:
Make it breaking I'd say, the prompt to actually read release notes will probably be necessary for people to understand why their snapshots broke and have confidence hitting `u`

markdalgleish:
Chatted with @michaeltaranto, and we're leaning towards non-breaking, but with a clear migration guide. In the vast majority of cases, this will just require consumers to resnapshot, which is pretty painless. This mirrors our strategy with Prettier, since it's also easy to update when tests break.

PR #164, opened 2018-09-13
feat(srcPaths): Support custom source paths other than 'src'

markdalgleish:
## Description

By default, sku expects your source code to be in a directory named `src` in the root of your project. If your source code needs to be arranged differently, you can provide a `srcPaths` array.

## Example Usage

 ```js
module.exports = {
  ...,
  srcPaths: [
    'src',
    'docs/src'
  ]
};
```

### CODE REVIEW

File: scripts/format.js [hunk 1]
```diff
@@ -1,15 +1,22 @@
...
 
 const args = require('../config/args').argv;
 
 console.log(chalk.cyan('Formatting source code with Prettier'));
 
-const filePattern = args.length === 0 ? defaultPath : args;
+const filePattern =
+  args.length === 0
+    ? builds[0].paths.src.map(
+        srcPath => `${path.relative(process.cwd(), srcPath)}/**/*.js`
```
mbfisher: Do we want to limit prettier to just JS files?

test/test-cases/custom-src-paths/custom-src-paths.test.js
```diff
@@ -0,0 +1,55 @@
+const { promisify } = require('es6-promisify');
```
mbfisher: 👏👊

scripts/format.js
markdalgleish: Hmmmm, good point. We weren't limiting this previously. This wasn't a conscious decision.

markdalgleish: @mbfisher Fixed! 😄

PR #165, opened 2018-09-14
feat(webpack): Add names to webpack configs

mbfisher:
This makes it really easy to target either the `client` or `render` config in a webpack decorator.


mbfisher:
Whilst this isn't a documented part of the config schema, it's used in the canonical example [multi compiler example](https://github.com/webpack/webpack/blob/master/examples/multi-compiler/webpack.config.js), so I assume it's legit.

PR #167, opened 2018-09-17
feat(webpack): Use webpack-node-externals on render build

mbfisher:
This will hopefully fix the memory related issues we've been having since upgrading to Webpack 4.

If you'd like to try it out you can install it with:

```
$> npm install 'https://github.com/seek-oss/sku#beta-render-node-externals-5'
```

This has grown quite large and deserves some explanation:

The title change is to add `webpack-node-externals` to the prerender webpack build. This reduces build time in my app from 165s to 8s (😵), and I have a good feeling that it could fix the memory issues we've been having since webpack 4. Trying to transpile **every** node import used in the render phase will use a lot of memory.

This change requires adding `seek-style-guide` and any user-specified `compilePackages` to `webpack-node-external`'s whitelist, or they don't get transpiled or bundled. We currently resolve `compilePackages` to absolute paths when parsing `sku.config.js`, but `webpack-node-externals` expects package names e.g. `node_modules/${packageName}`. So, this branch introduces 
 a `resolvePackage` function in the webpack configs to resolve package names to paths when needed. This has nice side-effects, like fixing `packageNameToClassPrefix`.

Which brings me to `utils`. Most recent sku PRs involving webpack changes have involved copy-pasting changes from `webpack.config.js` to `webpack.config.ssr.js`. This new `resolvePackage` function tipped me over the edge, so I've centralised anything that was identical between the two. This turned out to be all the `make*Loaders` functions and a few others.


config/webpack/webpack.config.js
```diff
@@ -231,10 +232,11 @@ const buildWebpackConfigs = builds.map(
         name: 'render',
         mode: webpackMode,
         entry: {
-          render:
-            paths.renderEntry || path.join(__dirname, '../server/server.js')
+          render: paths.renderEntry
         },
         target: 'node',
+        // Don't bundle or transpile anything in node_modules
+        externals: [nodeExternals()],
```
markdalgleish: We'll probably need exceptions here for `seek-style-guide` and libraries defined in the `compilePackages` option.

mbfisher:
This needs testing on a real app that doing actual pre-rendering before merging!

config/webpack/webpack.config.js
mbfisher: 👍

mbfisher:
This looks like `seek-style-guide` isn't being transpiled, which we thought would be fixed by adding `compilePackages` to the `wepback-node-externals` whitelist, right? Anyone else had a chance to test this? @jahredhope

markdalgleish:
My guess is that the `whitelist` arg on `webpack-node-externals` doesn’t support full paths, which is what we're passing in here.

mbfisher:
@markdalgleish correct, good spot

`An array for the externals to whitelist, so they will be included in the bundle. Can accept exact strings ('module_name'), regex patterns (/^module_name/), or a function that accepts the module name and returns whether it should be included. `

mbfisher:
To clarify, the issue here is that we're [resolving `compilePackages` into full paths](https://github.com/seek-oss/sku/blob/0b3ccf85dc95399da1584d4ce3b94bd0c2c4f27e/config/builds.js#L66). We need a solution to be able to pass the unmodified package names into `webpack-node-externals`.

mbfisher:
Think I have a fix for this, and been on a dedup rampage 🚜

config/webpack/utils.js
```diff
@@ -0,0 +1,138 @@
...
+ */
+const resolvePackage = packageName => {
+  try {
+    // First, try to leverage Node's require algorithm to find
+    // the package.
+    // We add `/package.json` to prevent `require.resolve` treating
+    // the directory as a package and returning the `main` file.
+    // We then use path.dirname to remove `/package.json` from the result.
+    // Clever, right?
+    return path.dirname(require.resolve(`${packageName}/package.json`));
```
mbfisher: @markdalgleish I've removed `{ paths: [cwd] }` from this, because it makes the tests pass. Are they covering the use-case that lead you to add it in the first place?

markdalgleish: My particular use case should still work—this would break `npm link` though, right?

mbfisher: Yes it does break `npm link`. I have a new version, just finishing off tests.

config/webpack/utils/resolvePackage.js
```diff
@@ -0,0 +1,78 @@
...
+};
+
+/**
+ * Resolve a package name to an absolute path.
+ * e.g. my-package -> /Users/me/code/my-project/node_modules/my-package
+ *
+ * Throws if a package is listed in the project's dependencies and cannot be resolved.
+ */
+const createPackageResolver = (fs, resolve) =>
+  memoize(function resolvePackage(packageName) {
```
markdalgleish: Nice use of memoize—was going to suggest this 👏

test/test-cases/zero-config/zero-config.test.js
```diff
@@ -0,0 +1,54 @@
+const { promisify } = require('es6-promisify');
+const rimrafAsync = promisify(require('rimraf'));
+const dirContentsToObject = require('../../utils/dirContentsToObject');
+const runSkuScriptInDir = require('../../utils/runSkuScriptInDir');
+const waitForUrls = require('../../utils/waitForUrls');
+const fetch = require('node-fetch');
+
+describe('zero-config', () => {
```
markdalgleish: 👏

config/webpack/utils/resolvePackage.js
```diff
@@ -0,0 +1,78 @@
...
+      debug(`Resolved ${packageName} to ${localPackage}`);
+      return localPackage;
+    }
+
+    // If we don't find a local copy we use Node's require algorithm to find one.
+    // This branch is used when sku or seek-style-guide have been `npm link`ed into a project,
+    // and also by the seek-style-guide.test.js sku test.
+    // Doing this can produce hard-to-debug build issues, so we log a warning.
+    const resolvedPackage = path.dirname(
+      resolve(`${packageName}/package.json`)
```
markdalgleish: Will this throw if it doesn't exist? Are we catching this error?

mbfisher: Yes we throw. I know the existing behaviour is to always return a path. My rationale here is that if you've listed the package in your dependencies and we can't resolve it, we should throw. Webpack is probably going to throw anyway. It felt weird to return a path when we know it doesn't exist.

Thoughts?

markdalgleish: I originally ran into this issue because we were trying to resolve `seek-style-guide`, since it's included as a default, regardless of whether a sku consumer has it installed. Would this still happen?

mbfisher: I've expanded the comments to make it clearer!

mbfisher:
I've fixed @adenj's issue, where `seek-style-guide` was not being properly whitelisted by `webpack-node-externals`, and updated the `seek-style-guide` test case to catch that issue in the future.

In doing so I've unearthed a new issue! Packages that aren't added to `compilePackages` because they are compiled before publishing to npm aren't added to the whitelist, and so can't be found in the render build. This is occurring in internal projects that use `@seek/*` npm modules.

Not sure what to do about that yet.

mbfisher:
Turns out that last issue was because I was `npm link`ing sku! Not sure what's going on there, think it's something to do with `require-like`, but don't think it should hold this up.

PR #168, opened 2018-09-17
feat(webpack): Add bundle analyser to build

michaeltaranto:
Provides an analysis report to help consumers identify and refine their dependencies in order to optimise their bundle at build time.

After running a `sku build` the report can be found at `dist/report/<bundle_name>.html`, where the bundle name will be either `client`, `render` or `server`.

### DIFF
--- a/config/webpack/plugins/bundleAnalyzer.js
+++ b/config/webpack/plugins/bundleAnalyzer.js
@@ -1,0 +1,8 @@
+const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
+
+module.exports = ({ name }) =>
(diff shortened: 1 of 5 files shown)


mbfisher:
Any chance the output file could be configurable? It's common to ship the contents of `dist/`, and I'd like to opt out of shipping the report.

michaeltaranto:
In hindsight (but pre-merge 😉) I feel like we should generate this at the root and add `report/` to the `.gitignore`. Thoughts? Im shooting for a zero config for this, and introduce that complexity if absolutely needed.

mbfisher:
Love it 👍🙌

PR #169, opened 2018-09-17
feat(mjs): Add support for .mjs files in node_modules

georgespyropoulos:
Some node modules are being published with the mjs extension (module js) and webpack is currently
unable to handle them. This change adds support for mjs modules/files

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -196,6 +196,11 @@ const buildWebpackConfigs = builds.map(
               )
             },
             {
+              test: /\.mjs$/,
+              include: /node_modules/,
+              type: 'javascript/auto'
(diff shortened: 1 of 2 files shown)


markdalgleish:
This is only configured to work on `node_modules`, but won't have any effect on your app code. Is this by design?

georgespyropoulos:
Yeah the issue we all had was a .mjs file in a module that wasn't understood by webpack so this PR fixes that issue. That said, perhaps we can apply this rule to all files. Can anyone think of a reason why that would be a bad idea?

markdalgleish:
I think this makes sense as an initial step. There's probably a bigger discussion around how we want to handle .mjs files in our apps, if at all.

PR #171, opened 2018-09-17
fix: Fix build error when seek-style-guide isn’t a dependency

markdalgleish:
## Why?

Builds without `seek-style-guide` were broken in https://github.com/seek-oss/sku/pull/166 because we call `require.resolve('seek-style-guide/package.json')`, which throws an error if it can't be found.

To fix this, we now wrap the require call in a try/catch block, falling back to the legacy behaviour if a package can't be found to maintain backwards compatibility.

## Development Notes

- Added a `.prettierrc` so Prettier editor extensions can pick it up.
- Added `report/` to `.gitignore`, so we don't commit reports generated during test runs.

### DIFF
--- a/config/builds.js
+++ b/config/builds.js
@@ -63,14 +63,24 @@ const builds = buildConfigs
       src: (buildConfig.srcPaths || ['src']).map(srcPath =>
         path.join(cwd, srcPath)
       ),
-      compilePackages: [
-        path.dirname(
-          require.resolve('seek-style-guide/package.json', { paths: [cwd] })
-        ),
-        ...compilePackages.map(package =>
-          path.join(cwd, 'node_modules', package)
-        )
-      ],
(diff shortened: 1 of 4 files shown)


config/builds.js
```diff
@@ -63,14 +63,24 @@ const builds = buildConfigs
...
+      compilePackages: ['seek-style-guide', ...compilePackages].map(
+        packageName => {
+          try {
+            // First, try to leverage Node's require algorithm to find
+            // the package relative to the current directory.
+            return path.dirname(
+              require.resolve(`${packageName}/package.json`, { paths: [cwd] })
+            );
+          } catch (err) {
+            // If a `package.json` file can't be resolved, maintain
```
mbfisher: Comments 👍👍👍👍👍👍👍

PR #175, opened 2018-09-26
fix(init): Create directory with the provided name, not 'init'

mbfisher:
#156 has broken `sku init`. It seemed easier to just ditch `commander` and do some simple manual args inspection, but I'm happy to be challenged!


scripts/init.js
```diff
@@ -1,47 +1,44 @@
...
-    projectName = name;
-  })
-  .option('--verbose', 'print additional logs')
-  .parse(process.argv);
+const projectName = args.argv[0];
 
-if (typeof projectName === 'undefined') {
-  console.error('Please specify the project directory:');
+const usage = () => {
+  console.error('Usage:');
```
markdalgleish: This probably shouldn't be a `console.error` call, since we display the usage guide in valid cases like `sku init --help`.

markdalgleish:
Just renamed the PR to more clearly call out what's being fixed.

PR #178, opened 2018-10-03
feat(build): Wipe dist folders on build

chardos:
Currently, sku is not clearing the `dist` folder when running `sku build`. This PR implements that.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,12 +1,12 @@
 // First, ensure the build is running in production mode
 process.env.NODE_ENV = 'production';
 
-const Promise = require('bluebird');
-const webpackPromise = Promise.promisify(require('webpack'));
+const { promisify } = require('es6-promisify');
+const webpackPromise = promisify(require('webpack'));
 const webpackConfigs = require('../config/webpack/webpack.config');
 const fs = require('fs-extra');
 const builds = require('../config/builds');
-const rimraf = require('rimraf');
+const rimraf = promisify(require('rimraf'));
 
 const runWebpack = config => {
   return webpackPromise(config).then(stats => {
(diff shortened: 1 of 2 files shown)


scripts/build.js
```diff
@@ -30,6 +31,23 @@ const runWebpack = config => {
...
+const compileAll = async () => {
+  webpackConfigs.forEach(async config => {
+    await runWebpack(config);
+  });
+};
+
+const cleanDistFolders = async () => {
+  builds.forEach(async ({ paths }) => {
+    console.log(`Clearing ${paths.dist}`);
+    await rimraf(paths.dist, err => {
```
markdalgleish: This `rimraf` call isn't returning a promise, is it? Looks like it's just dealing in regular callbacks here.

scripts/build.js
```diff
@@ -30,6 +31,23 @@ const runWebpack = config => {
   });
 };
 
+const compileAll = async () => {
```
markdalgleish: Using async/await means we no longer support Node below 7.6. Might be worth sanity checking if this is a problem.

scripts/build.js
```diff
@@ -30,6 +31,23 @@ const runWebpack = config => {
...
+  webpackConfigs.forEach(async config => {
+    await runWebpack(config);
+  });
+};
+
+const cleanDistFolders = async () => {
+  builds.forEach(async ({ paths }) => {
```
markdalgleish: For this function to wait until all async work is done, it would need to do something like this:

```js
await Promise.all(
  builds.map(async () => {
    ...
  })
);
```

scripts/build.js
chardos: Good catch. Will convert it using bluebirds `Promisify`?

markdalgleish: Yep. We already depend on es6-promisify, you import it like this:

```js
const { promisify } = require('es6-promisify');
````

File: scripts/build.js [hunk 4]
```diff
@@ -30,6 +31,14 @@ const runWebpack = config => {
   });
 };
 
+const cleanDistFolders = async () => {
+  await Promise.all(builds.map(({ paths }) => rimraf(paths.dist)));
+};
+
+const compileAll = async () => {
```
mbfisher: We can just return `Promise.all` here; this doesn't need to be an `async` function.

PR #181, opened 2018-10-05
chore(package): Update node version in package.json

chardos:
This PR (https://github.com/seek-oss/sku/pull/178) has added code using async/await. This PR updates the recommended node version in the package.json to be the latest LTS version


chardos:
@mbfisher The `engines` property is advisory only. There used to be a `enginesStrict` flag, but that got removed in Npm 3.

Happy to make it >=, but I feel like recommending a single version will only lead to more consistency across the business.

PR #182, opened 2018-10-05
docs(README): Static rendering multiple pages

chardos:
Some documentation about how to statically render multiple files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,32 @@ Changes to formatting are considered non-breaking. Please ensure you run `sku fo
 
 Generate static HTML files via a webpack-compiled render function that has access to your application code. For example, when building a React application, you can pre-render to static HTML with React's [renderToString](https://facebook.github.io/react/docs/react-dom-server.html#rendertostring) function.
 
+By default, sku is set up to render a single `index.html` page.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,6 +108,32 @@ Changes to formatting are considered non-breaking. Please ensure you run `sku fo
...
 
+By default, SKU is set up to render a single `index.html` page.
+
+If you need to statically pre-render multiple files, you can return an object whose keys are the path names, and values are the rendered mark up.
+
+In your `src/render.js`
+```
+export default ({ publicPath }) => {
+  return {
+    '/foo': <div>foo</div>,
```
markdalgleish: These should be strings, rather than React elements.

README.md
```diff
@@ -108,6 +108,32 @@ Changes to formatting are considered non-breaking. Please ensure you run `sku fo
 
 Generate static HTML files via a webpack-compiled render function that has access to your application code. For example, when building a React application, you can pre-render to static HTML with React's [renderToString](https://facebook.github.io/react/docs/react-dom-server.html#rendertostring) function.
 
+By default, SKU is set up to render a single `index.html` page.
```
markdalgleish: Should be "sku", not "SKU" 😄

PR #184, opened 2018-10-09
feat(storybook): Add Storybook

markdalgleish:
This adds [Storybook](https://storybook.js.org) to our standard toolkit, to make it as easy as possible to set up a local component explorer.

Running `sku storybook` will open up a local Storybook server, displaying all component instances declared in files named `*.stories.js`, for example:

```js
import { storiesOf } from 'sku/storybook';
import React from 'react';
import Button from './Button';

storiesOf('Button', module)
  .add('Primary', () => (
    <Button variant="primary">
      Primary
    </Button>
  ))
  .add('Secondary', () => (
    <Button variant="secondary">
      Secondary
    </Button>
  ));
```

_**NOTE:** To access the Storybook API, you should import from `sku/storybook`, since your project isn't depending on Storybook directly._

By default, Storybook runs on port `8081`. If you'd like to use a different port, you can provide it via the `storybookPort` option in `sku.config.js`:

```js
module.exports = {
  storybookPort: 9000
};
```

## Development Notes

I'm hoping to add more functionality to this, but this is very much an MVP to get a sense of how useful it is, and what changes we'd need.

It's a pretty difficult thing to test automatically, so I've opted to leave tests out for now. I've added a story to the test suite so it can at least be tested manually by running `sku storybook` in that folder.


config/storybook/config.js
```diff
@@ -0,0 +1,22 @@
+import { configure } from '@storybook/react';
+
+const reqs = [
```
mbfisher: I think a comment linking to the explanation of this in `config/storybook/webpack.config.js` would be helpful.

config/storybook/webpack.config.js
```diff
@@ -0,0 +1,49 @@
...
+  {
+    // We don't want to apply the entire webpack config,
+    // mainly because it configures entries and outputs,
+    // which would break the Storybook build.
+    module: clientWebpackConfig.module,
+    plugins: clientWebpackConfig.plugins
+  },
+  {
+    plugins: [
+      new webpack.DefinePlugin({
```
mbfisher: Does this work because `DefinePlugin` replaces the `__SKU_SRC_PATHS*` tokens before webpack examines the `require` expressions?

markdalgleish: Yep, exactly.

mbfisher: Learned a thing 👍

config/storybook/config.js
markdalgleish: Done 👍

PR #185, opened 2018-10-09
fix(build): Remove contents of dist only when building

jahredhope:
A recent change to sku deleted the `dist` folder on build.
This `dist` folder can sometimes be a mounted volume and can not be deleted.
This change ensure only the contents of the dist folder and not the folder itself is deleted.


etaoins:
It might also be a good idea to just `catch` the error and `console.warn`. I don't think failing to clean the dist should be a fatal error

PR #187, opened 2018-10-12
feat(typescript): Add support for typescript

mattsjones:
Opening this up for review because there's a lot here. Pretty happy with the outcome though. The test suite has actually been amazing so 👍 for that.

~Also, note this PR is against the babel7 branch as it's required for this to work. Might make another branch before actual merge.~


scripts/lint.js
```diff
@@ -1,53 +1,90 @@
...
+  if (report.errorCount > 0) {
     process.exit(1);
-  });
+  }
+
+  console.log(chalk.cyan('Checking Prettier format rules'));
+
+  const filePattern =
+    args.length === 0
+      ? builds[0].paths.src.map(srcPath => `${srcPath}/**/*.js`)
```
mbfisher: Should we add `.ts` files to this?

scripts/lint.js
```diff
@@ -1,53 +1,90 @@
...
+    files => files.length > 0
+  );
 
-const cli = new EslintCLI({
-  baseConfig: eslintConfig,
-  useEslintrc: false
-});
+  if (hasTSFiles) {
+    try {
+      await runTsc();
```
mbfisher: Not sure about `tsc` here. The TypeScript projects I've worked on follow a pattern I've found common with other compiled languages where you compile first, which catches compile errors, then lint. Running `tsc` as a lint tool would mean that a prod build does the Typescript compilation twice.

That said, the boundary between `tsc` and `(t|e)slint` is confusing; for example it's `tsc` that catches things like null safety and the other `strict` settings, but I tend to think of those as linting checks.

mattsjones: We're taking a different approach here. Babel is still our compilation tool and we are using (@babel/preset-typescript)[https://babeljs.io/docs/en/babel-preset-typescript] to handle the typescript part. We run `tsc` as part of the lint phase to ensure our types are correct. We will probably add `tslint` to this as well but didn't want to overwhelm this PR.

mattsjones: To be clear, `tsc` has no part in our compilation/bundling. It's only used to validate types. This is great though as we can have the best of both worlds.

scripts/lint.js
mattsjones: Good catch 👍

scripts/lint.js
mbfisher: Didn't realise babel doesn't do type checking, so in that context it makes a lot of sense to run `tsc` here 👍

mattsjones:
Closing to create new PR against `beta` branch

PR #189, opened 2018-10-17
fix: Clean render.js from output directory after build

markdalgleish:
## Why?

The static site generation phase creates an unnecessary `render.js` file in the output directory, which consumers may accidentally deploy to production.

## Migration guide

If your project includes a script to manually delete `render.js`, you can now remove it:

```diff
-"build": "sku build && rm dist/render.js"
+"build": "sku build"
```


markdalgleish:
Since we're no longer writing `render.js` to disk, our Node externals assertions were failing because they were inspecting the contents of this file.

To address this, I've introduced a custom webpack plugin that writes the list of Node externals to `externals.json`, which then get picked up by our snapshot tests.

PR #192, opened 2018-10-21
feat(typescript): Add support for TypeScript

mattsjones:


test/utils/dirContentsToObject.js
```diff
@@ -1,18 +1,20 @@
...
 
   const handleFile = (err, content, filePath, next) => {
     if (err) throw err;
 
     const relativeFilePath = relative(dirname, filePath);
     
-    files[relativeFilePath] = /\.js$/.test(relativeFilePath)
-      ? 'CONTENTS IGNORED IN SNAPSHOT TEST'
-      : content;
+    if (!includeExtensions || includeExtensions.includes(extname(relativeFilePath))) {
```
mattsjones: This was added to include other files into the snapshot. Specifically the `.css.js.d.ts`files as they are generated to the `src` directory.

mattsjones: This was added to include other files into the snapshot. Specifically the `.css.js.d.ts`files as they are generated to the `src` directory.

mattsjones: This was added to include other files into the snapshot. Specifically the `.css.js.d.ts`files as they are generated to the `src` directory.

config/webpack/utils/loaders.js
```diff
@@ -30,7 +31,20 @@ const makeCssLoaders = (options = {}) => {
     ...makeJsLoaders({ target: 'node' })
   ];
 
+  const cssModuleToTypeScriptLoader =
+    isTypeScript && !server
```
markdalgleish: Would be good to comment this one, I think.

package.json
```diff
@@ -105,6 +108,8 @@
     "svgo": "^1.0.5",
     "svgo-loader": "^2.1.0",
     "traverse": "^0.6.6",
+    "typescript": "^3.1.2",
+    "typings-for-css-modules-loader": "^1.7.0",
```
markdalgleish: This should be deleted now, right?

test/test-cases/typescript-css-modules/app/sku.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  entry: {
+    client: 'src/client.tsx',
+    render: 'src/render.tsx',
+    server: 'src/server.tsx'
```
markdalgleish: Why do we have all three here? This isn't really something we'd want to recommend.

michaeltaranto: For brevity it serves the test to ensure all paths are covered rather than having to build separate apps/configs in the test.

PR #195, opened 2018-10-26
feat(prettier): Support ignore files and control config in project root

michaeltaranto:
Provide editor compliance for prettier and honour a consumers `.prettierignore` file in the root of their project.

This PR also enforces the `tabWidth: 2` and runs both check and format across all `js`, `ts` and `tsx`, not just those within the target directory.

### DIFF
--- a/lib/runPrettier.js
+++ b/lib/runPrettier.js
@@ -1,14 +1,19 @@
+const fs = require('fs');
+const path = require('path');
+const { promisify } = require('util');
 const chalk = require('chalk');
 const runBin = require('./runBin');
 
-const runPrettier = ({ write, listDifferent, config, filePattern }) => {
-  const prettierConfig = config || {};
+const exists = promisify(fs.access);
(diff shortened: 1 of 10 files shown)


lib/runPrettier.js
```diff
@@ -1,22 +1,36 @@
+const fs = require('fs');
+const path = require('path');
+const { promisify } = require('util');
 const chalk = require('chalk');
 const runBin = require('./runBin');
 
-const runPrettier = ({ write, listDifferent, config, filePattern }) => {
-  const prettierConfig = config || {};
+const exists = promisify(fs.access);
+const prettierIgnorePath = path.join(process.cwd(), '.prettierignore');
```
mattsjones: Should we share this between `configure` and here?

michaeltaranto: As discussed, leaving as is in lieu of a refactor of `config/builds` making paths like these more available around the code base.

PR #196, opened 2018-10-29
feat: TypeScript, controlled configuration and better Prettier support

michaeltaranto:
Here is our first PR back to `master` from `beta`  🎉

### Features
In this PR we have added the following:
- **Support for TypeScript**
Allows partial adoption by adding `.ts` and `.tsx` files in existing project. Includes TSLint and editor integration (https://github.com/seek-oss/sku/pull/192, https://github.com/seek-oss/sku/pull/194)
- **Added `configure` step**
Manage configuration and control ignore files, such as `.gitignore` (https://github.com/seek-oss/sku/pull/193)
- **Improvements to how we manage Prettier**
Includes better editor support and the ability to ignore files using a `.prettierignore` in the project root (https://github.com/seek-oss/sku/pull/195).

#### New `configure` step

sku now has the concept of &ldquo;controlled configurations&rdquo;, which refers to configuration files that are dynamically generated in the project root, but managed by sku. Adding these files improves editor integration for tools such as [prettier](https://prettier.io) and [TypeScript](typescriptlang.org), while adding them to `.gitignore` keeps them a platform concern rather than that of a project.

The `configure` step will run as a post-install and will re-organise the `.gitignore` file in your project using [ensure-gitignore](https://github.com/seek-oss/ensure-gitignore). All sku controlled ignores will now appear in a block, eg:

```bash
my_project_specific_ignore/

# managed by sku
...
ignores_controlled_by_sku/
...
# end managed by sku
```

Any changes made within the bounds of the comment block will be re-written on the next run of `configure` (i.e. every `npm install`).

Please assert that the changes in this file make sense.

#### Prettier improvements

If you have previously added `.prettierrc` to your project it is recommended you remove it from version control. sku will now generate a controlled config file for you, ensuring it&rsquo;s always consistent with the `sku format`/`sku lint` commands.

In addition, previously there was no way to control what files Prettier ran on. The format and lint commands now run on all supported file formats (eg. `js`,`ts`,`tsx`,`md`,`less` and `css`), not just the files in your `src` folder.

If you want to opt out of any specific patterns you can do so by adding them to a `.prettierignore` file in the root of your project.

```bash
$ git rm .prettierrc
$ npm run format

# then stage & commit the change
```


mbfisher:
Just chatting with @michaeltaranto. Would be nice for generated `.prettierrc`, `tsconfig.json` etc files to have `DO NOT EDIT` comments in them to avoid confusion.

scripts/configure.js
```diff
@@ -16,18 +16,23 @@ const tslintConfig = require('../config/typescript/tslint.json');
 const prettierConfig = require('../config/prettier/prettierConfig');
 
 const addSep = p => `${p}${path.sep}`;
+const prependBanner = str =>
+  `/** THIS FILE IS GENERATED BY SKU, MANUAL CHANGES WILL BE DISGUARDED **/\n${str}`;
```
markdalgleish: Should be `discarded`.

michaeltaranto: Touchè

lib/runPrettier.js
```diff
@@ -1,40 +1,36 @@
...
+  try {
+    const ignoreExists = await exists(prettierIgnorePath);
+    if (ignoreExists) {
+      prettierArgs.push('--ignore-path', prettierIgnorePath);
+    }
+  } catch (err) {
+    // don't error if `.prettierignore` not found
+  }
+
+  prettierArgs.push('**/*.{js,ts,tsx,md,less,css}');
```
mbfisher: 🎉

scripts/configure.js
```diff
@@ -0,0 +1,84 @@
...
+    gitIgnorePatterns.push(tsConfigFileName);
+
+    // Generate TSLint configuration
+    const tslintConfigFileName = 'tslint.json';
+    await writeFileToCWD(tslintConfigFileName, tslintConfig);
+    gitIgnorePatterns.push(tslintConfigFileName);
+  }
+
+  // Generate Prettier configuration
+  // NOTE: We are not generating a banner as prettier does not support the `JSON
```
mbfisher: Annoying, but agreed.

scripts/format.js
```diff
@@ -1,25 +1,3 @@
...
 
-prettierWrite(filePattern, prettierConfig)
-  .then(() => {
-    console.log(chalk.cyan('Prettier format complete'));
-  })
-  .catch(exitCode => {
-    console.error('Error: Prettier exited with exit code', exitCode);
-    process.exit(1);
-  });
+(async () => await prettierWrite())();
```
mbfisher: Is this any different to just calling `prettierWrite()`? Node will wait for the event loop to be empty before exiting, and you'll get an `UnhandledPromiseRejection` error if anything throws.

Not a big deal, just curious!

mbfisher:
image

scripts/format.js
michaeltaranto: No i believe you are correct, that's essentially a relic of refactoring as the file became smaller and smaller 👏

PR #203, opened 2018-11-05
style(eslint): Setup for internal codebase

michaeltaranto:
Implements `eslint` on the internal codebase. This PR goes with the [`eslint-config-seek`](https://github.com/seek-oss/eslint-config-seek) rules, but turns off a few relating to console, file system access and process management. We can revisit these later if required, but keen to get this through as it will help catch some problems earlier.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,21 @@
+{
+  "extends": "seek",
+  "rules": {
+    "no-console": 0,
+    "no-process-exit": 0,
+    "no-sync": 0
+  },
+  "globals": {
+    "__SKU_PUBLIC_PATH__": true,
+    "__SKU_SRC_PATHS_0__": true,
+    "__SKU_SRC_PATHS_1__": true,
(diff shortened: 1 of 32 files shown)


.eslintrc
```diff
@@ -0,0 +1,21 @@
...
+    "__SKU_SRC_PATHS_0__": true,
+    "__SKU_SRC_PATHS_1__": true,
+    "__SKU_SRC_PATHS_2__": true,
+    "__SKU_SRC_PATHS_3__": true,
+    "__SKU_SRC_PATHS_4__": true,
+    "__SKU_SRC_PATHS_5__": true,
+    "__SKU_SRC_PATHS_6__": true,
+    "__SKU_SRC_PATHS_7__": true,
+    "__SKU_SRC_PATHS_8__": true,
+    "__SKU_SRC_PATHS_9__": true
```
mattsjones: Lol

PR #211, opened 2018-11-20
feat: Support library builds

markdalgleish:
If you need to build a UMD library instead of a web site, you can provide a `library` entry and `libraryName` option instead:

```js
modules.exports = {
  entry: {
    library: 'src/library.js',
    render: 'src/render.js'
  },
  libraryName: 'MyAwesomeLibrary'
};
```

Your `library` entry must export its public API via a default export:

```js
export default () => {
  console.log('Hello from my library!');
};
```

Note that, in this scenario, the `render` entry is only used to provide a development environment. No HTML will be generated when running `sku build`.


context/index.js
```diff
@@ -12,12 +12,23 @@ const appSkuConfig = fs.existsSync(appSkuConfigPath)
   : {};
 const skuConfig = merge(defaultSkuConfig, appSkuConfig);
 
+// Validate config
+if (skuConfig.entry.library && !skuConfig.libraryName) {
+  console.log(
+    chalk.red(
+      "Error: In your sku config, you've provided 'entry.library' without a corresponding 'libraryName' option."
```
michaeltaranto: I feel like this could be more to the point and link to the readme for details, eg.
```
"Error: No `libraryName` provided. For more information refer to the `Building a Library` documentation https://github.com/seek-oss/sku#building-a-library"
```

markdalgleish: Done 👍

README.md
```diff
@@ -367,6 +367,30 @@ module.exports = {
 
 Any `node_modules` passed into this option will be compiled through webpack as if they are part of your app.
 
+### Building a Library
```
mattsjones: Thoughts on moving this to an actual docs file?

markdalgleish: Yeah, we should, but I think we'll do this as part of a broader docs pass.

PR #213, opened 2018-11-22
feat: Transform ‘compilePackages’ code in tests

markdalgleish:
BREAKING CHANGE: `braid-design-system` is no longer auto-mocked in tests.

### DIFF
--- a/config/jest/jestConfig.js
+++ b/config/jest/jestConfig.js
@@ -1,9 +1,12 @@
+const escapeRegex = require('escape-string-regexp');
 const { paths } = require('../../context');
+const slash = '[/\\\\]'; // Cross-platform path delimiter regex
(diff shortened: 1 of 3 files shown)


config/jest/jestConfig.js
```diff
@@ -1,25 +1,27 @@
+const escapeRegex = require('escape-string-regexp');
 const { paths } = require('../../context');
+const slash = '[/\\\\]'; // Cross-platform path delimiter regex
```
michaeltaranto: Could we use `path.sep` for this use case? Not sure how it copes inside a regex/pattern like this.

markdalgleish: Spoke about this, decided to leave it for now.

PR #214, opened 2018-11-22
feat: Generate ESLint config files

markdalgleish:
To support ESLint editor integrations, we now generate config files in the root of the project directory.

Similar to Prettier, we now run ESLint across the whole project, ignoring files with `.eslintignore` which is generated automatically.

Any modifications to the ESLint config in `dangerouslySetESLintConfig` are written to disk as part of your `.eslintrc`.

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -57,7 +57,6 @@ module.exports = async () => {
 
   // Generate ESLint configuration
   const eslintConfigFilename = '.eslintrc';
-  console.log('eslintConfig', eslintConfig);
   await writeFileToCWD(eslintConfigFilename, eslintConfig);
   gitIgnorePatterns.push(eslintConfigFilename);
 
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -14,6 +14,7 @@ const {
 } = require('../config/webpack/plugins/bundleAnalyzer');
 const tslintConfig = require('../config/typescript/tslint.json');
 const prettierConfig = require('../config/prettier/prettierConfig');
+const eslintConfig = require('../config/eslint/eslintConfig');
 
 const addSep = p => `${p}${path.sep}`;
 const prependBanner = str =>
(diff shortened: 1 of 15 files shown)


lib/configure.js
```diff
@@ -54,6 +55,12 @@ module.exports = async () => {
     gitIgnorePatterns.push(tslintConfigFileName);
   }
 
+  // Generate ESLint configuration
+  const eslintConfigFilename = '.eslintrc';
+  console.log('eslintConfig', eslintConfig);
```
mattsjones: Console.log slip through ?

markdalgleish: Oops. Yep.

markdalgleish: Fixed 👍

test/test-cases/typescript-css-modules/app/sku.config.js
```diff
@@ -3,5 +3,13 @@ module.exports = {
     client: 'src/client.tsx',
     render: 'src/render.tsx',
     server: 'src/server.tsx'
+  },
+  // Use 'dangerouslySetESLintConfig' so we can
+  // check that overrides are written to disk.
```
michaeltaranto: Is there a corresponding test for the written output?

lib/configure.js
```diff
@@ -54,6 +55,11 @@ module.exports = async () => {
     gitIgnorePatterns.push(tslintConfigFileName);
   }
 
+  // Generate ESLint configuration
+  const eslintConfigFilename = '.eslintrc';
+  await writeFileToCWD(eslintConfigFilename, eslintConfig);
```
michaeltaranto: Confirming, can `.eslintrc` handle the comments in json? By default the comment banner is prepended so just making sure.

markdalgleish: Yep, it can.

test/test-cases/typescript-css-modules/app/sku.config.js
markdalgleish: Just added a test 👍

markdalgleish: Let me know what you think.

PR #217, opened 2018-11-26
fix(storybook): Add decorator to stories

michaeltaranto:
### Problem
When applying a CSS transform to the root element of a component screenshot services (such a Chromatic) would render the element but not at the translated position.

### Solution
Wrap every story in a `<div>` using a [global storybook decorator](https://storybook.js.org/addons/introduction/#storybook-decorators).


markdalgleish:
Nice. Love that this fixes the issue for everyone.

PR #240, opened 2018-12-11
fix: Ensure publicPath always ends in slash

michaeltaranto:
Currently when `publicPath` does not end with a `/` we append it when creating the assets for the compiled chunks, however when css image resources exceed the size limit and fall back to file loader the images are not being resolved when `publicPath` does not have a trailing slash.

This PR hoists the slash appending to operate on the `publicPath` available on context so it works for all use cases.


config/webpack/plugins/htmlRenderPlugin.js
```diff
@@ -13,11 +13,7 @@ const {
   defaultClientEntry
 } = require('../../../context');
 
-const createPublicUrl = (publicPath, asset) => {
-  const host = publicPath.endsWith('/') ? publicPath : `${publicPath}/`;
-
-  return `${host}${asset}`;
-};
+const createPublicUrl = (publicPath, asset) => `${publicPath}${asset}`;
```
markdalgleish: Is this function even useful anymore?

michaeltaranto: We concatenate in multiple places so figured it's worth keeping. Happy to inline if you prefer?

PR #241, opened 2018-12-11
chore: Change configure test to better handle generated files

michaeltaranto:
Adding generated files from the configure test that are leaving the working directory dirty when running tests locally.


mattsjones:
How about we just add `/test/test-cases/configure-test/app` and `/test/test-cases/configure-test/typescript-app` to `.gitignore`?

PR #243, opened 2018-12-21
fix(validation): Add error for incorrect entry extensions

mattsjones:
Example:
image


mattsjones:
Closing due to epic bike shed

PR #250, opened 2019-01-03
fix(validateConfig): Stop validate destroying object env

hscgavin:
Evil `validate` destroys our beloved `env` object even is part of the schema.

skuConfig:
```
env: {foo: "bar"}
...
```
after call `validateConfig(skuConfig)`, env in skuConfig is modified and empty `{}`.

https://github.com/eivindfjeldstad/validate/issues/54
https://github.com/eivindfjeldstad/validate/issues/42

Fix: set `strip = false`  on the options to disable `validate` muting original object for no reason.


hscgavin:
@mattsjones  `env` is part of the schema https://github.com/seek-oss/sku/blob/master/context/configSchema.js#L66

I guess `validate` mutates all `type: Object` to empty`{}`. If you use `validate` somewhere else,  it's worth double checking to make sure nothing gets destroyed 😈

PR #251, opened 2019-01-06
fix(configure): Add coverage directory to .prettierignore

fmr:
`sku test --coverage` generates a coverage directory, but the `sku lint` runs prettier, which still checks the coverage directory.

This commit adds the `coverage/` directory so prettier proceeds without an error.


.prettierignore
```diff
@@ -2,6 +2,7 @@ dist-build/
 dist-start/
 
 # managed by sku
+coverage/
```
markdalgleish: This portion of the file is actually programatically generated, it shouldn't be manually updated. It'll need to be updated in the `configure` script.

fmr: Thanks for the heads up. I've updated the configure script and related tests now.

fmr:
@mattsjones Done!

PR #253, opened 2019-01-07
fix: Upgrade css loader related dependencies

jahredhope:
Upgrades css-loader and it's related dependencies.

By upgrading to new versions we are able to remove some temporary fixes as-well as making future Webpack upgrades easier.

# Changes
- Use `cssnano` to replace removed functionality from `css-loader`
- Remove `string-replace-loader` hack. This is now supported in `css-loader`
- Changes to version has resulted in slight changes to CSS, resulting in content hashes changing requiring snapshot updates

**Note:** This change is being made a `fix` to ensure consumers see the new versions.


config/webpack/utils/loaders.js
```diff
@@ -51,34 +51,30 @@ const makeCssLoaders = (options = {}) => {
...
       }
     },
     {
       loader: require.resolve('postcss-loader'),
       options: {
-        plugins: () => [require('autoprefixer')(supportedBrowsers)]
+        plugins: () => [
+          require('autoprefixer')(supportedBrowsers),
+          // Minimize CSS on production builds
+          ...(isProductionBuild ? [require('cssnano')] : [])
```
michaeltaranto: The `postcss-loader` documentation indicates that you need to invoke the function returned from `cssnano`, eg. `require('cssnano')()`, not true? See https://github.com/postcss/postcss-loader#plugins

config/webpack/utils/loaders.js
```diff
@@ -51,34 +51,30 @@ const makeCssLoaders = (options = {}) => {
...
     },
-    {
-      // Hacky fix for https://github.com/webpack-contrib/css-loader/issues/74
-      loader: require.resolve('string-replace-loader'),
-      options: {
-        search: '(url\\([\'"]?)(.)',
-        replace: '$1\\$2',
-        flags: 'g'
-      }
-    },
```
michaeltaranto: 😍

config/webpack/utils/loaders.js
jahredhope: Good call. Fixed.

PR #254, opened 2019-01-07
refactor: Apply HTML plugin directly in webpack.config.js

jahredhope:
Apply `html-render-webpack-plugin` directly in Webpack config.
This API became available with `v1.1.0` and removes the need to apply the plugin seperately.

This won't result in any behaviour changes.


mattsjones:
Can we make this a `refactor` rather then a `fix`. If there is zero behaviour difference we shouldn't perform a release.

test/test-cases/custom-src-paths/.gitignore
```diff
@@ -1,6 +1,7 @@
 # managed by sku
 .eslintrc
 .prettierrc
+coverage/
```
mattsjones: Realise this has nothing to do with your PR but pretty weird it's showing in this PR.

markdalgleish: It's because the Renovate PR didn't run `sku configure`.

jahredhope: This will go away once I rebase off the changes from https://github.com/seek-oss/sku/pull/253/

jahredhope: This is now gone with merge from master

PR #255, opened 2019-01-08
feat(start): Assign free port if requested one is unavailable

mattsjones:
I have chosen not to implement this for SSR projects as they have two ports making it more complex and likely to cause problems. Open to adding it, just want input from the SSR teams before doing so.


mattsjones:
Closing for now as this breaks hot reloading. Will open again with SSR and hot reloading working.

PR #258, opened 2019-01-10
test: Add (watch) typeahead plugin for Jest

fmr:
Adds typeahead support to Jest's watch mode

From jest-community/jest-watch-typeahead

watch


fmr:
Testing from a consumer:

sku-test-watch

PR #266, opened 2019-01-16
fix(startssr): Make SKU_PORT runtime configurable

michaeltaranto:
Previously `SKU_PORT` was baked into ssr the same way we currently support `env` in the sku config, but critically it didnt support changing the port at runtime. Now we honour `SKU_PORT` from the environment as an override for the `serverPort` that is specified in the sku config file.

This supports more advanced use cases, like starting the server once per cpu core using pm2 with a different port number for each — preventing port collisions.

### DIFF
--- a/config/server/index.js
+++ b/config/server/index.js
@@ -1,6 +1,6 @@
 import http from 'http';
 import app from './server';
-const port = process.env.SKU_PORT || 8080;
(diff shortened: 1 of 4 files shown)


michaeltaranto:
Going a slightly different way with this, new PR coming shortly

PR #268, opened 2019-01-16
fix(storybook): Bypass htmlRender plugin

markdalgleish:
Refactors to our webpack config logic broke the Storybook integration. This PR fixes it, and also introduces a Storybook test (long overdue—sorry 🙇🏻‍♂️).

Note, I've also had to improve our child process cleanup logic a little bit because Storybook was leaving child processes running in our test suite.


markdalgleish:
@michaeltaranto Sounds good 👍

PR #269, opened 2019-01-17
fix(configure): Run configure before each script

mattsjones:


bin/sku.js
```diff
@@ -20,10 +29,9 @@ switch (script) {
   case 'start':
   case 'start-ssr':
   case 'storybook':
-  case 'chromatic':
-  case 'configure': {
-    const scriptPath = require.resolve(`../scripts/${script}`);
```
michaeltaranto: Any idea why this was previously doing `require.resolve`? Wary of breaking a use case we're unaware of.

test/test-cases/custom-src-paths/custom-src-paths.test.js
```diff
@@ -8,10 +8,6 @@ const skuConfig = require('./sku.config');
 const targetDirectory = `${__dirname}/dist`;
 
 describe('custom-src-paths', () => {
-  beforeAll(async () => {
-    await runSkuScriptInDir('configure', __dirname);
-  });
-
```
markdalgleish: Nice cleanup 🎉

PR #270, opened 2019-01-17
fix(webpack): Clean up webpack API deprecation warnings

mattsjones:
This PR moves the [css-in-js-loader](https://github.com/naistran/css-in-js-loader) to be an internal file of sku as it is unmaintained and using deprecated webpack APIs. If we continue to use this loader long term we should move it to its own repo, this at least fixes the warnings for now. Also, our test suite does a decent job of catching issues with it.  

The only changes to the loader source code are to update its usage of webpacks plugin API causing the warnings.

### DIFF
--- a/config/webpack/utils/cssInJsLoader.js
+++ b/config/webpack/utils/cssInJsLoader.js
@@ -1,4 +1,5 @@
 /* eslint-disable */
+// Original source: https://github.com/naistran/css-in-js-loader/blob/master/index.js
 var path = require('path');
 var fs = require('fs');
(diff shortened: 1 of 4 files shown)


config/webpack/utils/cssInJsLoader.js
```diff
@@ -0,0 +1,122 @@
+/* eslint-disable */
```
michaeltaranto: If we are opting out of linting because the source is from elsewhere, might be handy to have a link back to the original source?

PR #273, opened 2019-01-21
docs: Add information about bundle analyzer in README

chardos:
Added webpack-bundle-analyzer to one of our projects and then realised sku already generated reports for us. That's awesome, will be nice to surface this information.


README.md
```diff
@@ -420,6 +420,10 @@ Last but not least, please note that commands for SSR are different to the ones
 - Use `sku build-ssr` to build your production assets. You can then run `node ./dist/server.js`. Your server will run at `http://localhost:xxxx`, where `xxxx` is `serverPort`.
 - Use `sku test-ssr` to test your application
 
+## Bundle Analysis
```
mattsjones: Maybe reference the actual plugin with a link. Also, can we move this to the `features` section?

chardos: Done.

PR #277, opened 2019-01-22
test(manual): Fix manual test script

mattsjones:
Finding myself using the test cases quite often and this should help iterate on new features. 
I'm aware this is not an ideal fix but it at least gets the manual script working again. I think an ideal solution would be a larger refactor of the test directory structure.


markdalgleish:
Thanks for fixing this—I was feeling the pain of this recently, was hoping to fix it soon myself.

PR #281, opened 2019-01-24
feat: Add support for rendering dynamic imports

mattsjones:
- Render all dynamic imports through [loadable-components](https://github.com/smooth-code/loadable-components)
- All scripts are now loaded async for concurrent downloads
- Render`link rel="preload"` tags for all tags
- Above features allow efficient braid theming across multiple brands

Note: most of this PR is test updates

BREAKING CHANGE: Render and Server API now need to use the new SkuProvider component for correct script/style rendering.


config/babel/babelConfig.js
```diff
@@ -28,13 +28,12 @@ module.exports = ({ target, lang = 'js' }) => {
...
     ],
-    require.resolve('babel-plugin-macros')
+    require.resolve('babel-plugin-macros'),
+    require.resolve('@loadable/babel-plugin')
   ];
 
   if (isBrowser) {
     plugins.push(require.resolve('babel-plugin-seek-style-guide'));
-  } else {
-    plugins.push(require.resolve('babel-plugin-dynamic-import-node'));
```
markdalgleish: 👋👋👋

context/defaultClientEntry.js
```diff
@@ -0,0 +1 @@
+module.exports = 'main';
```
markdalgleish: Why do we need this? Feel like this could use a comment.

entry/render/buildRenderEntry.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable import/no-unresolved */
```
markdalgleish: This is operating on the whole file. Should we scope it to the `'__sku_alias__renderEntry'` import?

context/defaultClientEntry.js
mattsjones: Well it's shared in 3 places now. Felt like it should be centrally defined.

test/test-cases/custom-src-paths/__snapshots__/custom-src-paths.test.js.snap
```diff
@@ -6,13 +6,17 @@ Object {
     <meta charset=\\"UTF-8\\">
     <title>hello-world</title>
     <meta name=\\"viewport\\" content=\\"width=device-width, initial-scale=1\\">
-    
+    <link data-chunk=\\"main\\" rel=\\"preload\\" as=\\"script\\" href=\\"http://localhost:4002/runtime-27f87d9f0d19bf7deb87.js\\">
```
markdalgleish: This is so cool 😍

markdalgleish: We're losing `crossorigin="anonymous"`—is this a problem?

markdalgleish: ☝️ Oops, commented on the wrong line—but the point still stands.

mattsjones: Working on that atm. That's what the loadable-component PR is for

test/test-cases/typescript-css-modules/typescript-css-modules.test.js
```diff
@@ -65,7 +65,7 @@ describe('typescript-css-modules', () => {
     });
 
     it('should generate the expected files', async () => {
-      const files = await dirContentsToObject(distDir);
+      const files = await dirContentsToObject(distDir, ['.js', '.css']);
```
markdalgleish: Why is this being limited to JS and CSS now?

test/test-cases/typescript-css-modules/app/src/server.tsx
```diff
@@ -4,23 +4,34 @@ import { renderToString } from 'react-dom/server';
...
+
     res.send(`
       <!DOCTYPE html>
       <html>
         <head>
           <meta charset="UTF-8">
           <title>My Awesome Project</title>
           <meta name="viewport" content="width=device-width, initial-scale=1">
-          ${headTags}
+          ${getHeadTags()}
```
markdalgleish: Why is this now a function call?

test/test-cases/zero-config/zero-config.test.js
```diff
@@ -1,50 +0,0 @@
...
-    });
-  });
-
-  describe('lint', () => {
-    it('should lint successfully', async () => {
-      const { childProcess } = await runSkuScriptInDir('lint', __dirname);
-      expect(childProcess.exitCode).toEqual(0);
-    });
-  });
-});
```
markdalgleish: 👋👋👋

test/test-cases/typescript-css-modules/typescript-css-modules.test.js
mattsjones: We now spit out a `webpackStats.json` file which contains info that will be specific to the machine that created it so it's not a good candidate for snapshots. The function currently only has a whitelist so I just whitelisted everything else.

test/test-cases/typescript-css-modules/app/src/server.tsx
mattsjones: Previously we created a list of assets for a build. The assets are now specific to each render so a function is required.

test/test-cases/typescript-css-modules/typescript-css-modules.test.js
markdalgleish: Do we only output this file for server rendering?

mattsjones: Yes

markdalgleish: 😎

PR #288, opened 2019-01-29
feat(webpack): Add feature to enable source maps in production

jahredhope:
Some projects may want to publish source maps for their compiled code.
Currently we only create source maps for the development environment.

Builds source maps could be seen to have a negative effect:
 - Increase in build time
 - Increase visibility of source code

Therefore adding source maps by default could be considered a breaking change. This change adds source maps behind a configuration feature toggle.

A later major release could then change the default.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -78,9 +78,9 @@ An array of routes for the app. Each route must specify a name and a route corre
 Example:
 
 ```js
-{
-  routes: [{ name: 'home', route: '/' }];
-}
+const config = {
+  routes: [{ name: 'home', route: '/' }]
+};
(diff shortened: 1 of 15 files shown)


docs/configuration.md
```diff
@@ -246,3 +247,16 @@ Example:
...
+## `sourceMapsProd` [`boolean`]
+
+By default source maps will be generated only for development builds.
+Set to `true` to enable source maps in production.
+
+Example:
+
+```js
+{
+  sourceMapsProd: true;
```
michaeltaranto:
```suggestion
  sourceMapsProd: true
```

PR #289, opened 2019-01-30
feat: Serialize config from render to client

mattsjones:
This PR also ensures `loadableReady` has fired before calling client code.


entry/client/index.js
```diff
@@ -0,0 +1,11 @@
+// '__sku_alias__clientEntry' is a webpack alias
+// pointing to the consuming apps client entry
+import client from '__sku_alias__clientEntry';
+
+import { loadableReady } from '@loadable/component';
+
+import configKey from '../configKey';
+
+loadableReady(() => {
+  client(window[configKey] || {});
```
michaeltaranto: Should we be deleting the configKey off window after its passed back into client function?

entry/client/index.js
```diff
@@ -0,0 +1,18 @@
...
+import { loadableReady } from '@loadable/component';
+
+import clientContextKey from '../clientContextKey';
+
+loadableReady(() => {
+  let clientContext = {};
+
+  if (window[clientContextKey]) {
+    clientContext = window[clientContextKey];
+    delete window[clientContextKey];
```
markdalgleish: 👏👏👏

entry/render/buildRenderEntry.js
```diff
@@ -1,21 +1,3 @@
...
-  const app = await render.renderApp({ ...renderParams, SkuProvider });
-
-  return await render.renderDocument({
-    ...renderParams,
-    app,
-    bodyTags: getBodyTags(),
-    headTags: getHeadTags()
-  });
-};
+export default async renderParams => render(renderParams);
```
michaeltaranto: `await render`?

mattsjones: Tbh this is just a passthrough function now so it doesn't even need to be async really.

PR #290, opened 2019-01-31
refactor(validation): Change config validation library

mattsjones:
I've made this change as the previous library doesn't support variant types for config which will likely be introduced in v8.


context/configSchema.js
```diff
@@ -1,153 +1,121 @@
...
+    type: 'array',
+    items: {
+      type: 'object',
+      props: {
+        route: {
+          type: 'string'
+        },
+        name: {
+          type: 'string',
+          optional: true
```
michaeltaranto: Is it a deliberate change to make `name` optional? Previously `entry` was the only optional key.

context/configSchema.js
```diff
@@ -1,153 +1,121 @@
-const Schema = require('validate');
+const Validator = require('fastest-validator');
```
michaeltaranto: sounds quick 😉

mattsjones: Haha I know. The name is a bit lame but it was the only fit for purpose one I found.

context/configSchema.js
mattsjones: Yes, this was deliberate. It's technically not required and probably shouldn't have been in the first place.

PR #292, opened 2019-02-05
feat(ssr): Support onStart callback for server entry

michaeltaranto:
Provide the ability to pass a callback to the server entry of your app that will be invoked after the server starts. The callback receives the app instance as the first and only parameter.

An example server entry: 
```diff
export default () => ({
  renderCallback: () => { ... },
  middleware: [ ... ],
+ onStart: app => {
+   console.log('My app started 👯‍♀️!');
+   app.keepAliveTimeout = 20000;
+ }  
})
```

### DIFF
--- a/config/server/index.js
+++ b/config/server/index.js
@@ -1,6 +1,6 @@
 import http from 'http';
 import commandLineArgs from 'command-line-args';
-import app from './server';
+import { app, onStart } from './server';
 
 const { port } = commandLineArgs(
   [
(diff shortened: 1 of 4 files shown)


markdalgleish:
Looks good. Do we need to update the readme?

michaeltaranto:
Happy to, given yesterday's discussion around the server rendering api re-think I wasnt sure if we wanted to commit to documenting this. Thoughts?

markdalgleish:
Yeah, I hear that—but I think it's still a good practice to document what's there in the meantime.

PR #298, opened 2019-02-08
fix: Move loadable component to a normal dependency

mattsjones:


@loadable/component.ts
```diff
@@ -0,0 +1,3 @@
+// This is provided so consumers can import `sku/loadable/component`,
```
markdalgleish:
```suggestion
// This is provided so consumers can import `sku/@loadable/component`,
```

PR #303, opened 2019-02-11
fix: Use loadable plugin in static apps

mattsjones:
This PR makes a few changes
- Use the loadable webpack plugin on static apps (I was wrong about it not being needed)
- Only preload scripts (not styles) and hoist styles above preloads. I feel this is the optimal way to load the dependent resources for now. 
- Update our internal version of React (as I refused to change a function to a class 😉)
- Makes the `react-css-modules` test case self update to ensure client apps are running correctly


test/test-cases/react-css-modules/app/src/App.js
```diff
@@ -2,13 +2,21 @@ import React from 'react';
...
-    <div
-      className={`${lessStyles.nested} ${jsStyles.nested}`}
-      data-automation-text
-    >
-      Hello World
+export default () => {
+  const [message, setMessage] = React.useState('Initial render');
+
+  setTimeout(() => {
+    setMessage('Updated render');
```
markdalgleish: Would be worth adding a comment here to explain why—otherwise it looks a bit odd.

test/test-cases/react-css-modules/app/src/App.js
```diff
@@ -2,13 +2,21 @@ import React from 'react';
...
 
-export default () => (
-  <div className={`${lessStyles.root} ${jsStyles.root}`}>
-    <div
-      className={`${lessStyles.nested} ${jsStyles.nested}`}
-      data-automation-text
-    >
-      Hello World
+export default () => {
+  const [message, setMessage] = React.useState('Initial render');
```
markdalgleish: 😍

test/test-cases/library-build/__snapshots__/library-build.test.js.snap
```diff
@@ -14,9 +14,8 @@ Object {
     <meta charset=\\"UTF-8\\">
     <title>My Awesome Project</title>
     <meta name=\\"viewport\\" content=\\"width=device-width, initial-scale=1\\">
-    <link data-chunk=\\"main\\" rel=\\"preload\\" as=\\"style\\" href=\\"/MyLibrary.css\\">
+    <link data-chunk=\\"main\\" rel=\\"stylesheet\\" href=\\"/MyLibrary.css\\">
```
markdalgleish: 👍

PR #304, opened 2019-02-12
 feat(storybook): Add 'knobs' add-on

markdalgleish:
We now bundle [@storybook/addon-knobs](https://www.npmjs.com/package/@storybook/addon-knobs) to support editable props within the Storybook interface.

## Example Usage

```js
import React from 'react';
import Button from './Button';

import { storiesOf } from 'sku/@storybook/react';
import { text } from 'sku/@storybook/addon-knobs';

storiesOf('Button', module)
  .add('Primary', () => (
    <Button variant="primary">{text('Children', 'Primary')}</Button>
  ))
  .add('Secondary', () => (
    <Button variant="secondary">{text('Children', 'Secondary')}</Button>
  ));
```

## Breaking Change

You must now import from `sku/@storybook/react`

```diff
-import { storiesOf } from 'sku/storybook';
+import { storiesOf } from 'sku/@storybook/react';
```


package.json
```diff
@@ -83,9 +83,11 @@
     "@loadable/component": "^5.6.0",
     "@loadable/server": "^5.6.0",
     "@loadable/webpack-plugin": "^5.5.0",
+    "@storybook/addon-knobs": "^4.1.11",
     "@storybook/react": "^4.0.0-alpha.24",
```
michaeltaranto: Worth looking at getting off the alpha while we're here? `4.1.11` is the latest atm.

markdalgleish: Consumers are already downloading latest because of the version range, but we should still do this anyway.

PR #305, opened 2019-02-13
test: Improve snapshots quality

mattsjones:
Happy to discuss these changes but the idea is to make snapshots easier to review and check for regressions. 

Updates made: 
- Added a CSS snapshot serializer based on prettier
- Added an HTML serializer
- Added custom serializer for app snapshots which includes git diffing for client hydrates against the source HTML


test/test-cases/library-build/__snapshots__/library-build.test.js.snap
```diff
@@ -2,39 +2,50 @@
...
-}
+</html>
+
+POST HYDRATE DIFFS: 
+     <script>
+       window.MyLibrary();
+     </script>
++    <div class="styles__root___1ys27hL">
++      This element was generated by the library
++    </div>
```
mattsjones: Library fix here

markdalgleish: 😍😍😍

PR #306, opened 2019-02-13
fix(library): Fix development HTML in library mode

markdalgleish:
The development environment for libraries is broken, because loadable-components is rendering the library script tag as async. To fix this, we now bypass loadable-components during the render phase and provide the `link` and `script` tags manually.


test/test-cases/library-build/__snapshots__/library-build.test.js.snap
```diff
@@ -14,13 +14,11 @@ Object {
...
-    <link data-chunk=\\"main\\" rel=\\"stylesheet\\" href=\\"/MyLibrary.css\\">
-<link data-chunk=\\"main\\" rel=\\"preload\\" as=\\"script\\" href=\\"/MyLibrary.js\\">
+    <link rel=\\"stylesheet\\" type=\\"text/css\\" href=\\"/MyLibrary.css\\">
   </head>
   <body>
-    <script>window.__LOADABLE_REQUIRED_CHUNKS__ = [];</script>
-<script async=\\"\\" data-chunk=\\"main\\" src=\\"/MyLibrary.js\\"></script>
-    <script>window.MyLibrary();</script>
+    <script src=\\"/MyLibrary.js\\"></script>
+    <script>window.MyLibrary();</script><div class=\\"styles__root___1ys27hL\\">This element was generated by the library</div>
```
markdalgleish: ☝️  This was the line that we missed.

PR #307, opened 2019-02-14
docs: Fix spacing on migration guide

chardos:
Spacing is slightly off.

image


michaeltaranto:
You should be fine to make this a `docs` only change, we dont need to publish this.

PR #318, opened 2019-02-19
chore(renovate): Update a few settings

michaeltaranto:
Making a few updates to the renovate config:

#### prCreation: "not-pending"
Only open the PR once all CI checks have run (pass or fail)

#### lockFileMaintenance: true
Will update the lock file to take the the latest versions of deps compatible with specified semver. Uses the default schedule of "before 5am on monday" to keep the lock file fresh.

#### packageRules
Apply a schedule to style guide dependencies to update once a month

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,13 @@
 {
-  "extends": [
-    "seek"
+  "extends": ["seek"],
+  "prCreation": "not-pending",
+  "lockFileMaintenance": {
+    "enabled": true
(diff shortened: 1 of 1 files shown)


michaeltaranto:
Sure. Was initially focusing on a change set that was just for internal, but happy to look at webpack and/or babel in this PR if we agree?

mattsjones:
I'd like to trial it personally.

PR #319, opened 2019-02-20
chore(renovate): Experiment with scheduled package updates

michaeltaranto:
We are going to try implementing a couple of package rule sets to keep our dependencies up to date — starting with some obvious sets in `webpack` and `babel`. In addition we have a `tooling` group that is a collection of dependencies for the `test`/`lint`/`format` commands.

Deliberately starting with a higher cadence to allow us to review the behaviour (not necessarily merging PRs).

Also fixes the regex for the `seek-asia-style-guide` which was incorrect.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -23,7 +23,7 @@
     },
     {
       "groupName": "Tooling maintenance",
-      "packagePatterns": ["jest|puppeteer|tslint|eslint|prettier"],
+      "packagePatterns": ["jest|puppeteer|tslint|eslint"],
       "schedule": ["before 8am on Thursday"]
     }
   ]
--- a/renovate.json
+++ b/renovate.json
@@ -8,7 +8,7 @@
   "packageRules": [
     {
       "groupName": "Style guide maintenance",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -8,8 +8,23 @@
...
+      "schedule": ["before 8am on Thursday"]
+    },
+    {
+      "groupName": "Babel maintenance",
+      "packagePatterns": ["^@?babel[/-]"],
+      "schedule": ["before 8am on Thursday"]
+    },
+    {
+      "groupName": "Tooling maintenance",
+      "packagePatterns": ["jest|puppeteer|tslint|eslint|prettier"],
```
jahredhope: I'm less interested in prettier. Given the requirement for consumers to run commands to update when prettier updates I'm more happy for it to go stale and get major leaps forward.

If we want it, I'd suggest adding it as a seperate group and scheduling it every 1-3 months.

michaeltaranto: Yep im happy with that. Will remove prettier from this for now.

PR #342, opened 2019-04-01
fix(test): Support dynamic imports in tests

markdalgleish:
When running `sku test`, we need to ensure that dynamic `import()` calls are compiled correctly for node, which we do with [babel-plugin-dynamic-import-node](https://github.com/airbnb/babel-plugin-dynamic-import-node).


test/test-cases/multiple-routes/__snapshots__/multiple-routes.test.js.snap
```diff
@@ -19,25 +19,25 @@ SOURCE HTML:
     <link data-chunk="handlers-Home"
           rel="preload"
           as="script"
-          href="http://localhost:4004/handlers-Home-9299312b46da1275b24f.js"
+          href="http://localhost:4004/handlers-Home-d4afda529b3b1a8b8da7.js"
```
kosanna: huh?

markdalgleish: Because the script contents changed, it's changed the hash in the snapshot tests.

PR #355, opened 2019-05-08
fix: Remove sku prefix from Loadable Components import

michaeltaranto:
BREAKING CHANGE: Loadable Components import is no longer prefixed with `sku/`, see [v9 Migration guide](./migration-guides/v9.0.0).

#### Dynamic imports using `@loadable`

If your application features code splitting, you will need to update your [Loadable Components](https://github.com/smooth-code/loadable-components) imports to remove the `sku/` prefix. For example:

```diff
-import loadable from 'sku/@loadable/component';
+import loadable from '@loadable/component';
```

#### Braid Design System consumers

If you are consuming `braid-design-system`, you will need to ensure that you are on v0.0.63 or newer as it depends on this release.


michaeltaranto:
Closing this PR as we are changing our approach a little. May not require a breaking version. Stay tuned.

PR #367, opened 2019-05-28
feat(playroom): Add Playroom support

markdalgleish:
This PR adds [Playroom](https://github.com/seek-oss/playroom) support via the `sku playroom` and `sku build-playroom` commands.


context/defaultSkuConfig.js
```diff
@@ -1,5 +1,6 @@
 const supportedBrowsers = require('browserslist-config-seek');
 const path = require('path');
+const isTypeScript = require('../lib/isTypeScript');
```
mattsjones: `isTypeScript` is actually a function. However, it probably should be a value because otherwise, we may call it multiple times and it's not zero cost.

test/test-cases/react-css-modules/__snapshots__/react-css-modules.test.js.snap
```diff
@@ -36,9 +36,9 @@ SOURCE HTML:
     </script>
     <script src="runtime~main.6ac4c33d73ae9704bbb9.bundle.js">
     </script>
-    <script src="vendors~main.de7408c173c18a6a9a78.bundle.js">
+    <script src="vendors~main.2ed72cf0f594567d550e.bundle.js">
     </script>
-    <script src="main.9e6db395536d80e4abd3.bundle.js">
+    <script src="main.8643fa625f3b6b9dae1a.bundle.js">
```
mattsjones: Any idea why this updated?

mattsjones: Might be related to lock file changes.

PR #388, opened 2019-07-15
feat: Allow support for pre-rendering head tags

jahredhope:
Currently `getHeadTags` is designed to be called once. At which time it'll return all head tags. Re-calling the function will return all head tags again, including ones that had been added since the previous call.

This change allows the consumer to call `flushHeadTags` multiple times, each time only new tags will be returned. This allows the consumer to call flushHeadTags whenever they are ready to respond, not just after all rendering is complete.

Additionally getHeadTags and flushHeadTags supports limiting the types of tags that are returned:
```js
getHeadTags({excludeJs: false, excludeCss: false}) // default
flushHeadTags({excludeJs: false, excludeCss: false}) // default
```
```js
getHeadTags({excludeJs: true}) // Useful if preloading JS is not necessary
flushHeadTags({excludeJs: true}) // Useful if preloading JS is not necessary
```


entry/makeExtractor.js
```diff
@@ -19,17 +27,48 @@ export default (stats, publicPath) => {
...
+  let previouslyReturnedCssHeadTags = '';
+
+  const getJsHeadTags = () =>
+    extractor
+      .getLinkTags(extraScriptTagAttributes)
+      .split('\n')
+      .filter(tag => tag.includes('as="script"'))
+      .join('\n');
+
+  const getCssHeadTags = () => extractor.getStyleTags();
```
mattsjones: Feel like the code would be much cleaner if `getJsHeadTags` and `getCssHeadTags` returned an array instead of a string?

jahredhope: I kept the current pattern to optimise for the predominant use-case: a single call to `getHeadTags`. By keeping these as strings we don't have to split `getStyleTags` if you only ever call `getHeadTags` once.
It's not hard to implement either way, so if you still think an array is better I'll refactor.

mattsjones: I think I prefer what I suggested but that's just my opinion. Will leave it up to you.

docs/docs/server-rendering.md
```diff
@@ -49,6 +49,46 @@ export default () => ({
...
+import { initialResponseTemplate, followupResponseTemplate } from './template';
+import middleware from './middleware';
+
+export default () => ({
+  renderCallback: ({ SkuProvider, getBodyTags, getHeadTags }, req, res) => {
+    res.status(200);
+    // Call `getHeadTags` early to retrieve whatever tags are available.
+    res.write(initialResponseTemplate({ headTags: getHeadTags() }));
+    res.flush();
+    await Promise.resolve();
```
mattsjones: I think this is confusing. Would rather just have a comment saying something like `Perform required data fetching` or similar.

entry/makeExtractor.js
ashin: I wonder if we are making things overly complex by trying to jam more logic into a single method.

Would it be easier to seperate the use cases of returning a concatted string, and partially returning the tags as an array into seperate methods?

docs/docs/server-rendering.md
jahredhope: It's actually a required behaviour to allow Node to release the event loop and actually perform the flush that it said it has flushed. You'll need to do something like this for any flush call.
The other option is to use
```js
await res.flush()
```

but that's a bit deceitful, as it wouldn't be obvious that res.flush doesn't return a promise and your just saying 'seriously Node, why are you so single minded'

entry/makeExtractor.js
jahredhope: Good call. Separated out into `flushHeadTags`

PR #392, opened 2019-07-19
fix(deps): Upgrade to Storybook v5

michaeltaranto:
Upgrades the storybook ecosystem to v5. Worth noting this PR moves to `~5.0.11` rather than `^5.1.9`, this is due to [a few](https://github.com/storybookjs/storybook/issues/6984) [issues](https://github.com/storybookjs/storybook/issues/6204) that have come through since the `5.1` release.

They have built in types too now so this removes our dependency on the unofficial types package.


michaeltaranto:
Closing to sort through some dependency of dependency issues 🕸

PR #403, opened 2019-07-30
fix(deps): Upgrade storybook to v5

michaeltaranto:
Upgrades the storybook ecosystem to v5. Worth noting this PR moves to `~5.0.11` rather than `^5.1.9`, this is due to [a few](https://github.com/storybookjs/storybook/issues/6984) [issues](https://github.com/storybookjs/storybook/issues/6204) that have come through since the `5.1` release.

They have built in types too now so this removes our dependency on the unofficial types package.


test/test-cases/configure-test/configure.test.js
```diff
@@ -69,13 +69,13 @@ describe('configure', () => {
...
     it('should not generate tsconfig config', async () => {
-      expect(readJsonC(appFolder, 'tsconfig.json')).rejects.toThrow(
+      return expect(readJsonC(appFolder, 'tsconfig.json')).rejects.toThrow(
         /ENOENT: no such file or directory, open \'.*\/tsconfig\.json\'/,
       );
     });
 
     it('should not generate tslint config', async () => {
-      expect(readJsonC(appFolder, 'tslint.json')).rejects.toThrow(
+      return expect(readJsonC(appFolder, 'tslint.json')).rejects.toThrow(
```
mattsjones: Pretty sure we should `await` here or remove the `async`.

michaeltaranto:
Closing while i upgrade & test `storybook-chromatic` as well

PR #415, opened 2019-08-12
fix(format): Allow paths to be passed to sku format

jahredhope:
To support patterns such as [lint-staged](https://github.com/okonet/lint-staged#configuration) we need to pass additional arguments to `sku format` to the underlying prettier command.

### DIFF
--- a/lib/runPrettier.js
+++ b/lib/runPrettier.js
@@ -13,7 +13,7 @@ const prettierConfigPath = path.join(
   '../config/prettier/prettierConfig.js',
 );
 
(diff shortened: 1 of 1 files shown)


michaeltaranto:
Looks good to me, for bonus points do you have time to pass the [`pathsToCheck`](https://github.com/seek-oss/sku/blob/bf037308cc423ad542acf0614ea2665ae7d05dee/scripts/lint.js#L29) through in the [`lint` script](https://github.com/seek-oss/sku/blob/bf037308cc423ad542acf0614ea2665ae7d05dee/scripts/lint.js#L41)?

PR #418, opened 2019-08-16
feat(pre-commit): Provide script to check code before CI

michaeltaranto:
This PR makes some changes to `sku format` & `sku lint` to support the new `sku pre-commit` script which can be used in development to runs checks on the code that is staged for committing. Details of the changes are as follows:

### `sku format` / `sku lint`
Changes have been made to both these scripts to more consistently support accepting paths as an argument. This is namely to support tools like [`lint-staged`](https://www.npmjs.com/package/lint-staged) as a performance optimisation to run scripts over the part of the code base that has been changed.

*Note: For Typescript consumers, when providing file paths to `sku lint`, if the paths do not contain Typescript files we skip running `tsc` as a performance optimisation. This is to support workflows like `pre-commit` described below, however `tsc` will still execute when `sku lint` is operating on the entire project.*

### `sku pre-commit`
This new script can be used in development to run checks on the code that is staged for committing. Under the hood it is using [`lint-staged`](https://www.npmjs.com/package/lint-staged) with two steps:
- Run `sku format` & `sku lint` over the staged code.
- If it's a `yarn` project and the `package.json` or `yarn.lock` has changed, run `yarn check-integrity`.

#### Usage
It is recommended to run this using git hooks with a library like [`husky`](https://www.npmjs.com/package/husky). You can install it as follows:

```shell
yarn add husky --dev
# or
npm install husky --save-dev
```

The recommended configuration is then as simple as adding the following to your `package.json`:
```js
"husky": {
  "hooks": {
    "pre-commit": "sku pre-commit"
  }
}
```

#### Migrating from existing lint-staged implementation
If you already had a `lint-staged` setup for this purpose, you can uninstall `lint-staged` and delete it's configuration in favour of calling `sku pre-commit`

```diff
"husky": {
  "hooks": {
-    "pre-commit": "lint-staged"
+    "pre-commit": "sku pre-commit"
  }
},
-"lint-staged": {
-  "**/*.js": [
-    "eslint --fix",
-    "git add"
-  ],
-  "**/*.{js,ts,tsx,md,less,css}": [
-    "sku format",
-    "git add"
-  ]
-},
```
If you are doing anything above and beyond formatting and linting please let us know as it may have value to others as well.


scripts/lint.js
```diff
@@ -1,42 +1,40 @@
...
-    useEslintrc: false,
-  });
+      await runTSLint(pathsToCheck);
 
-  const pathsToCheck = args.length === 0 ? ['.'] : args;
-
-  const formatter = cli.getFormatter();
-  console.log(chalk.gray(`eslint ${pathsToCheck.join(' ')}`));
-  const report = cli.executeOnFiles(pathsToCheck);
+      if (!hasPaths || pathsIncludeTS) {
```
markdalgleish: JSON files should probably trigger a `tsc` run too, since TypeScript treats them as typed files.

config/lintStaged/lintStagedConfig.js
```diff
@@ -0,0 +1,20 @@
...
+// Yarn lock integrity check
+if (isYarn) {
+  steps['+(package.json|yarn.lock)'] = [
+    'yarn check-integrity',
+    'git add package.json yarn.lock',
+  ];
+}
+
+// Format & lint
+steps['**/*.{js,jsx,ts,tsx,md,less,css}'] = [
```
markdalgleish: Should we include JSON files too?

docs/docs/extra-features.md
```diff
@@ -43,3 +43,17 @@ module.exports = {
 ## Bundle analysis
 
 `sku` comes with bundle analysis built in via [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer). A report is generated in the `/report` directory when `sku build` is run.
+
+## Pre-commit hooks
+
+To speed up the feedback loop on linting and formatting errors `sku` provides `pre-commit` script that can be run to catch simple problems before CI. To make use of this it is recommended that you install [husky](https://www.npmjs.com/package/husky) as a development dependency and configure it in `package.json` as follows:
```
markdalgleish:
```suggestion
To speed up the feedback loop on linting and formatting errors, `sku` provides a `pre-commit` script that can be run to catch simple problems before CI. To make use of this, it's recommended that you install [husky](https://www.npmjs.com/package/husky) as a development dependency and configure it in `package.json` as follows:
```

config/lintStaged/lintStagedConfig.js
michaeltaranto: I think we should, the reason i didnt was to keep it in sync with the [defaults in our `runPrettier` script](https://github.com/seek-oss/sku/blob/f4cfb5ac540f0cd0acdd6e5f1383877d1b577303/lib/runPrettier.js#L40). Up for grabs is whether we should sync these and maybe import one from the other.

Thoughts?

scripts/lint.js
michaeltaranto: Happy to add that, excluding obvious files like `package.json` and `package-lock.json` though right?

context/configSchema.js
```diff
@@ -149,4 +149,5 @@ module.exports = validator.compile({
   },
   playroomPort: { type: 'number', optional: true },
   playroomTitle: { type: 'string', optional: true },
+  isYarn: { type: 'boolean', optional: true },
```
mattsjones: This isn't required as it's not publically configurable.

michaeltaranto: Agreed. Moved from context to standalone script.

PR #425, opened 2019-09-04
fix(storybook): Support hot reloading

michaeltaranto:
Add hot reloading to the storybook setup to better support development iteration cycle while working on components.

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -25,17 +25,20 @@ const {
   polyfills,
   isLibrary,
   libraryName,
-  isStartScript,
   sourceMapsProd,
   supportedBrowsers,
 } = config;
 
 // port is only required for dev builds
-const makeWebpackConfig = ({ isIntegration = false, port = 0 } = {}) => {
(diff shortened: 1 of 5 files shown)


config/webpack/webpack.config.js
```diff
@@ -73,7 +76,7 @@ const makeWebpackConfig = ({ isIntegration = false, port = 0 } = {}) => {
 
   const createEntry = entry => [
     ...resolvedPolyfills,
-    ...(isStartScript ? devServerEntries : []),
+    ...(hot ? devServerEntries : []),
```
mattsjones: 🤔 Not sure we should use `hot` instead of `isStartScript` in some of these scenarios. `sku start` currently doesn't support hot reloading right?

michaeltaranto: Yeah happy to bikeshed this. `isStartScript` is suffering a couple of problems at the minute:
1. It's not triggered when using the `test-manual` script,
2. Should be true outside of `start`, eg. `storybook`, `playroom` etc.

PR #442, opened 2019-10-18
fix(ssr): Include public path in image urls

michaeltaranto:
Server rendering images that were larger than the base64 limit would result in only the filename being rendered—not including the full public path as required. Also extended the test to capture that the full public path is rendered correctly.

Other fixes include:
#### `start-ssr`
Now executes the `onStart` handler in dev and production mode

#### `start`/`start-ssr`
The logs output by webpack dev server were not configured to correctly reflect what was running. Both commands now output configured host (the first resolved host from the sku config file), and also the public path where assets are served from, most often `/` in dev mode.
```diff
-｢wds｣: Project is running at http://localhost:5000/
-｢wds｣: webpack output is served from undefined
+｢wds｣: Project is running at http://dev.seek.com.au:5000/
+｢wds｣: webpack output is served from /
｢wds｣: Content not from webpack is served from /Users/mtaranto/projects/seek/cpp-react/public
｢wds｣: 404s will fallback to /index.html
```

#### `init`
Updated the default app template that is rendered when initialising a new repo, adopting the new `Stack` api from `braid-design-system`.


test/test-cases/braid-design-system/app/sku.config.js
```diff
@@ -1,10 +1,7 @@
 const ListExternalsPlugin = require('../../../utils/ListExternalsWebpackPlugin');
 
 module.exports = {
-  sites: [
-    { name: 'seekAnz', host: 'dev.seek.com.au' },
-    { name: 'jobStreet', host: 'dev.jobstreet.com' },
-  ],
+  sites: [{ name: 'seekAnz' }, { name: 'jobStreet' }],
```
mattsjones: Any specific reason for this change?

scripts/start.js
```diff
@@ -64,6 +64,8 @@ const localhost = '0.0.0.0';
 
   const devServer = new WebpackDevServer(parentCompiler, {
     contentBase: paths.public,
+    publicPath: paths.publicPath,
+    host: appHosts[0],
```
mattsjones: Just double checking this doesn't impact using the other hosts?

test/test-cases/braid-design-system/app/sku.config.js
michaeltaranto: Yes these hosts arent setup in CI (or even guaranteed to be setup on a dev env). Given the test doesnt exercise these explicitly it made sense to remove them.

scripts/start.js
michaeltaranto: Any example you can think of? Happy to test further.

PR #453, opened 2019-12-30
feat: Add ‘provideDefaultChromaticViewports’ config option

markdalgleish:
Currently sku has hard-coded widths of 320px and 1200px for [Chromatic](https://www.chromaticqa.com) screenshots. This PR adds the ability to disable this by setting `provideDefaultChromaticViewports` to `false`.


markdalgleish:
After chatting with @jahredhope about this, decided to change this from `screenshotWidths` to `provideDefaultChromaticViewports` since we only want to allow this to be disabled, not further customised. In the future we'd like to remove the defaults entirely, but this avoids a breaking change for now.

PR #455, opened 2020-02-04
feat(playroom): Support snippets

michaeltaranto:


context/index.js
```diff
@@ -126,6 +129,5 @@ module.exports = {
     port: skuConfig.playroomPort,
     widths: skuConfig.playroomWidths,
     title: skuConfig.playroomTitle,
-    themes: skuConfig.playroomThemes,
```
markdalgleish: Why is this being removed? Was it redundant?

michaeltaranto: Yeah `themes` for playroom are under `paths`, along with `components`, `snippets`, etc. This was unreferenced and caused confusion.

PR #456, opened 2020-02-06
fix(playroom): Update webpack config

michaeltaranto:


config/playroom/makePlayroomConfig.js
```diff
@@ -38,5 +38,9 @@ module.exports = () => ({
     module: clientWebpackConfig.module,
     resolve: clientWebpackConfig.resolve,
     plugins: clientWebpackConfig.plugins,
+    optimization: {
+      ...clientWebpackConfig.optimization,
```
mattsjones: I think we should let playroom drive the rest of optimisation section.

PR #463, opened 2020-03-24
Update to Prettier v2

mattsjones:
Prettier v2 requires node > 10.13.0 so also bumping minimum supported node version, making this PR a breaking change.


package.json
```diff
@@ -7,7 +7,7 @@
     "sku": "./bin/sku.js"
   },
   "engines": {
-    "node": ">=8.6"
+    "node": ">=10.13.0"
```
michaeltaranto: Might be worth doing a lock file upgrade on the back of this. Thoughts?

PR #468, opened 2020-03-24
Update to Jest 25

mattsjones:


.prettierignore
```diff
@@ -9,4 +9,5 @@ dist-start/
 dist-storybook/
 report/
 template/
-.changeset/*.md
\ No newline at end of file
+.changeset/*.md
+CHANGELOG.md
```
michaeltaranto: Whats the reason for ignoring this?

mattsjones: It's auto-generated by changesets and it breaks our rules.

PR #470, opened 2020-03-25
Remove support for .css.js files

mattsjones:


test/test-cases/ssr-hello-world/src/App.js
```diff
@@ -1,11 +1,10 @@
 import React from 'react';
 import lessStyles from './lessStyles.less';
-import jsStyles from './jsStyles.css.js';
 import logo from './logo.png';
 
 export default () => (
-  <div className={`${lessStyles.root} ${jsStyles.root}`}>
```
markdalgleish: This has made me realise that code like this won't break the build, and values like this one will resolve as `undefined`. Instead, we should probably throw an error if any `.css.js` files are detected, otherwise we run the risk of silently breaking styles when upgrading.

mattsjones: https://github.com/seek-oss/sku/pull/470/commits/eab86323fb9c06ba31a48eea4baa75dc90c5aa44

PR #471, opened 2020-03-25
Remove deprecated react-treat re-exports from sku/treat

mattsjones:


mattsjones:
This uses the new import/export types in typescript 3.8. To make sure this works correctly for everyone I've force updated all babel and typescript deps.

PR #474, opened 2020-03-29
Update html-render-webpack-plugin

mattsjones:


scripts/start.js
```diff
@@ -81,20 +72,17 @@ const localhost = '0.0.0.0';
...
+        htmlRenderPlugin
+          .renderWhenReady({
+            route: matchingRoute.route,
+            routeName: matchingRoute.name,
+            site: getSiteForHost(req.hostname),
+            environment: environments.length > 0 ? environments[0] : undefined,
+          })
+          .then((html) => res.send(html))
+          .catch((renderError) => {
+            const webpackStats = renderError.webpackStats.toJson();
```
jahredhope: I'm worried that errors aren't a happy time, and weird things happen. Can we check if webpackStats exists? It's not an expected use-case, but errors rarely work as expected :)

mattsjones: Good shout 👍

PR #495, opened 2020-04-16
Add support for sku-telemetry

mattsjones:


.changeset/lovely-buttons-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+Add metrics for core scripts
```
markdalgleish, line 5: Might be worth adding a note here to install the package, rather than letting them find out the hard way?

mattsjones: Was just adding this haha. Is what I added enough or should we add more reasoning?

markdalgleish: I'd say go a step further and add the commands for both yarn and npm here, so it's literally just a copy and paste.

scripts/build-ssr.js
```diff
@@ -24,10 +27,20 @@ const { port } = require('../context');
     await run(webpack(serverConfig));
     await copyPublicFiles();
 
-    console.log(green('Sku build complete!'));
-  } catch (e) {
-    console.error(red(e));
+    const timeTaken = performance.now();
+    track.timing('build_ssr', timeTaken, { status: 'success' });
```
jahredhope, line 31: Using a seperate key for build_ssr will make it hard to report on 'how many builds of any kind have happened'.
Would it make sense to call the key 'build' and move ssr into a tag?
```suggestion
    track.timing('build', timeTaken, { status: 'success', type: 'ssr' });
```

mattsjones: Addressed this for both build and start

PR #502, opened 2020-05-06
Add CSP support

mattcompiles:


docs/docs/csp.md
```diff
@@ -0,0 +1,56 @@
...
+
+If you need to allow scripts that are only known client side (e.g. scripts loaded by tag managers) you can add their URLs to the `cspExtraHosts` array in `sku.config.js`.
+
+### Extra SSR Setup
+
+As sku doesn't handle the returned HTML in SSR apps, any extra scripts (scripts not created by sku) must be registered.
+
+In the `renderCallback` function, register all extra script tags (inline and external) via the `registerScript` function.
+
+> If you are using multi-part reponses via the `flushHeadTags` API, all scripts must be registered before sending the the initial response.
```
markdalgleish, line 19:
```suggestion
> If you are using multi-part responses via the `flushHeadTags` API, all scripts must be registered before sending the the initial response.
```

entry/client/index.js
```diff
@@ -30,9 +30,9 @@ if (process.env.NODE_ENV === 'development') {
 loadableReady(() => {
   let clientContext = {};
 
-  if (window[clientContextKey]) {
-    clientContext = window[clientContextKey];
-    delete window[clientContextKey];
+  const dataElement = document.getElementById(clientContextKey);
+  if (dataElement) {
+    clientContext = JSON.parse(dataElement.textContent);
```
markdalgleish, line 35: Love this 👏

entry/render/index.js
```diff
@@ -1,16 +1,18 @@
...
 
 import render from '__sku_alias__renderEntry';
 
 const libraryName = SKU_LIBRARY_NAME;
 const publicPath = __SKU_PUBLIC_PATH__;
+const csp = __SKU_CSP__;
 
 export const serializeConfig = (config) =>
-  `<script>window.${clientContextKey} = ${serializeJavascript(
+  `<script id="${clientContextKey}" type="application/json">${serializeJavascript(
```
markdalgleish, line 13: Now that this is JSON rather than JavaScript, might need to be mindful that [serialize-javascript](https://www.npmjs.com/package/serialize-javascript) supports a superset of JSON, so we might need to ensure that only strict JSON is allowed here.

entry/csp.js
```diff
@@ -0,0 +1,95 @@
...
+    if (src) {
+      addScriptUrl(src);
+    } else if (
+      !scriptTypeIgnoreList.includes(scriptNode.getAttribute('type'))
+    ) {
+      addScriptContents(scriptNode.firstChild.rawText);
+    }
+  };
+
+  const registerScript = (script) => {
```
jahredhope, line 42: Is there any way for this script to fail if headers have already been flushed once? If I understand correctly right now the new script will just not be returned in the second head flush, leading to any registredScripts to not be white-listed silently.
```ts
if(csp.hasReturned) {
  throw new Error(`Unable to register script. Content Security Policy already sent. Try registering scripts before calling flushHeadTags. Script: ${script.substr(0, 30)}`);
}
```

entry/csp.js
```diff
@@ -0,0 +1,95 @@
...
+      policies.join(' '),
+      '">',
+    ].join('');
+  };
+
+  const handleHtml = (html) => {
+    const root = parse(html, { script: true });
+
+    if (!root.valid) {
+      throw new Error('Invalid HTML');
```
jahredhope, line 80: How leniant is `node-html-parser`? It seems like this would be the first time we are validating a consumer's end HTML, which means that this could pick up a few typos in some project's big template strings.
If so, is it worth making this error message a little more verbose?
```suggestion
      throw new Error(`Unable to parse HTML in order to create CSP tag. Check the output of renderDocument for invalid HTML. Error: ${root.error}`);
```

entry/csp.js
```diff
@@ -0,0 +1,95 @@
...
+      inlineCspShas.push(`'sha256-${sha}'`);
+    }
+
+    const scriptSrcPolicy = [
+      `'self'`,
+      ...hosts.values(),
+      ...inlineCspShas,
+    ].join(' ');
+
+    policies.push(`script-src ${scriptSrcPolicy};`);
```
markdalgleish, line 67: Is this the only item in the array? Does it even need to be an array?

entry/csp.js
```diff
@@ -0,0 +1,102 @@
...
+    if (tagReturned) {
+      throw new Error(
+        `Unable to register script. Content Security Policy already sent. Try registering scripts before calling flushHeadTags. Script: ${script.substr(
+          0,
+          30,
+        )}`,
+      );
+    }
+
+    const root = parse(script, { script: true });
```
jahredhope, line 53: I'm just thinking now... this is on one of the hotest/critical paths of a server render. And most scripts will just be static strings re-used for every render.
It would be nice to have some way to pre-parse this HTML so we don't need to redo it everytime. I'm not sure memoization would work here as only some scripts may be static, some may change with every render.

I don't think there is anything to do now, but perhaps a future feature may allow you to pre-parse scripts like:
```ts
interface CsPValidAnalyticsScript {
  src?: string,
  content?: string,
}
const cspValidAnalyticsScript = generateCspValidScript(script);

registerScript(cspValidAnalyticsScript);
```

entry/csp.js
```diff
@@ -0,0 +1,102 @@
...
+
+    return `<meta http-equiv="Content-Security-Policy" content="${scriptSrcPolicy};">`;
+  };
+
+  const handleHtml = (html) => {
+    const root = parse(html, { script: true });
+
+    if (!root.valid) {
+      throw new Error(
+        `Unable to parse HTML in order to create CSP tag. Check the following output of renderDocument for invalid HTML.\n${html}`,
```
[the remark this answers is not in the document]
mattcompiles: I hear you, but how else can they debug the issue?

entry/csp.js
```diff
@@ -0,0 +1,102 @@
+import { createHash } from 'crypto';
+import { parse } from 'node-html-parser';
+import { URL } from 'url';
+
+const scriptTypeIgnoreList = ['application/json', 'application/ld+json'];
+
+const defaultBaseName = 'http://relative-url';
+
+const hashScriptContents = (scriptContents) =>
+  createHash('sha256').update(scriptContents).digest('base64');
```
zsims, line 10: awesome 👏

test/test-cases/multiple-routes/app/sku.config.js
```diff
@@ -4,4 +4,6 @@ module.exports = {
   environments: ['production'],
   port: 8202,
   publicPath: '/static/place',
+  cspEnabled: true,
+  cspExtraHosts: ['https://error-tracking.com', 'https://fb-tracking.com'],
```
zsims, line 8: this is brilliant :-)

.changeset/unlucky-bags-exist.md
```diff
@@ -0,0 +1,7 @@
+---
+'sku': minor
+---
+
+Add Content Security Policy generation
```
zsims, line 5: should we clarify that this is just for `script-src`?

config/webpack/webpack.config.js
```diff
@@ -29,6 +29,8 @@ const {
   sourceMapsProd,
   supportedBrowsers,
   displayNamesProd,
+  cspEnabled,
+  cspExtraHosts,
```
zsims, line 33: Similar to the above question, is it worth clarifying this is for scripts, e.g. `cspExtraScriptSrcHosts` ? Bit of a mouthful though

entry/csp.js
mattcompiles: Yeah I thought about this as well. Happy to add something later to improve the perf.

entry/csp.js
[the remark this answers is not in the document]
jahredhope: Could we do the same as the script error message, put a reasonable limit on the HTML. Just enough to know what the HTML roughtly looks like but by pointing them to renderDocument they know if they need more info they can log out their renderDocument function.
```
html.length > 400 ? `${html.subStr(0, 200)}...` : html;
```
The limit can be quite high because it's just about not completely spamming the console out.

PR #507, opened 2020-05-12
Fix ReferenceError in compilePackages peer validation

72636c:
```
Error validating peer dependencies
ReferenceError: compilePackage is not defined
    at module.exports (/workdir/node_modules/sku/lib/validatePeersDeps.js:90:30)
```


72636c:
🍴 #508

PR #510, opened 2020-05-17
Detect Buildkite as CI

mattcompiles:


lib/isCI.js
```diff
@@ -0,0 +1 @@
+module.exports = Boolean(process.env.CI || process.env.BUILDKITE_BUILD_ID);
```
georgespyropoulos, line 1: Perhaps I'm misunderstanding this but buildkite already sets a `CI` env var (https://buildkite.com/docs/pipelines/environment-variables). Also, if you were trying to detect buildkite specifically, would the `BUILDKITE` env var be more appropriate here?

mattcompiles: The issue is that the Buildkite docker plugin doesn't forward any of those vars by default and most teams don't set it up. However, `BUILDKITE_BUILD_ID ` is forwarded by default.

PR #522, opened 2020-06-19
feat: 🎸 HTTPS for devServer and expose devServer middleware

willhackett:
New authentication flows and tooling surrounding authentication require HTTPS which makes local development tricky when using custom domains.

This feature plans to introduce `https` support for the Webpack Dev Server but also expose the `express` middleware within to allow further extension.


willhackett:
Closed in favour of: https://github.com/seek-oss/sku/pull/523

PR #523, opened 2020-06-24
feat: 🎸 HTTPS for devServer and expose devServer middleware

willhackett:
New authentication flows and tooling surrounding authentication require HTTPS which makes local development tricky when using custom domains.

This feature plans to introduce https support for the Webpack Dev Server but also expose the express middleware within to allow further extension.


entry/server/index.js
```diff
@@ -22,15 +23,36 @@ const startCallback = () => {
...
-    server.on('request', app);
-    currentApp = app;
-  });
-} else {
-  app.listen(port, startCallback);
-}
+(async () => {
+  if (module.hot) {
+    if (__SKU_DEV_MIDDLEWARE_ENABLED__) {
+      console.log(`PATH IS ${__SKU_DEV_MIDDLEWARE_PATH__}`);
```
mattcompiles, line 29: Left a console.log

lib/certificate.js
```diff
@@ -0,0 +1,108 @@
...
+      },
+      {
+        name: 'subjectAltName',
+        altNames,
+      },
+    ],
+  });
+};
+
+const generateCertificate = async (certificatePath, certificateDirPath) => {
```
mattcompiles, line 69: Might be worth adding a timing metric around this so we can see who's using the feature and how long it's taking.

context/configSchema.js
```diff
@@ -164,4 +164,8 @@ module.exports = validator.compile({
     type: 'array',
     items: { type: 'string' },
   },
+  useHttpsDevServer: {
+    type: 'boolean',
+    optional: true,
```
mattcompiles, line 169: Because we validate the merged default config this is actually required.

context/configSchema.js
```diff
@@ -164,4 +164,8 @@ module.exports = validator.compile({
     type: 'array',
     items: { type: 'string' },
   },
+  useHttpsDevServer: {
```
mattcompiles, line 167: Need to add a schema for `devServerMiddleware` as well

context/index.js
```diff
@@ -130,4 +139,7 @@ module.exports = {
   orderImports: Boolean(skuConfig.orderImports),
   cspEnabled: skuConfig.cspEnabled,
   cspExtraScriptSrcHosts: skuConfig.cspExtraScriptSrcHosts,
+  useHttpsDevServer: skuConfig.useHttpsDevServer,
+  devServerMiddleware,
```
mattcompiles, line 143: This is available via the `paths` key already. Can we remove this one?

context/defaultSkuConfig.js
```diff
@@ -48,4 +48,6 @@ module.exports = {
   orderImports: false,
   cspEnabled: false,
   cspExtraScriptSrcHosts: [],
+  useHttpsDevServer: false,
+  devServerMiddleware: null,
```
mattcompiles, line 52: Should this be `dev-middleware.js`? I don't anywhere else that it does the defaulting.

context/index.js
```diff
@@ -63,7 +63,16 @@ const publicPath = skuConfig.publicPath.endsWith('/')
   ? skuConfig.publicPath
   : `${skuConfig.publicPath}/`;
 
+const devServerMiddleware =
+  skuConfig.devServerMiddleware &&
+  getPathFromCwd(skuConfig.devServerMiddleware);
+
+const useDevServerMiddleware = devServerMiddleware
```
mattcompiles, line 70: Maybe we should throw an error if the dev middleware file doesn't exist? Otherwise people might be confused why it's not working.

entry/server/server.js
```diff
@@ -10,7 +10,7 @@ const csp = __SKU_CSP__;
 
 const serverOptions = serverExports({ publicPath });
 
-const { renderCallback, middleware, onStart } = serverOptions;
+const { middleware, onStart, renderCallback } = serverOptions;
```
mattcompiles, line 13: No biggy, but these changes are redundant.

lib/certificate.js
```diff
@@ -0,0 +1,114 @@
...
+const { hosts } = require('../context');
+const { performance } = require('perf_hooks');
+const track = require('../telemetry');
+
+const exists = promisify(fs.exists);
+const mkdir = promisify(fs.mkdir);
+const unlink = promisify(fs.unlink);
+const writeFile = promisify(fs.writeFile);
+const stat = promisify(fs.stat);
+const readFile = promisify(fs.readFile);
```
mattcompiles, line 16: I would suggest just importing from `fs-extra` as it support promises natively.

lib/certificate.js
```diff
@@ -0,0 +1,114 @@
...
+    ],
+  });
+};
+
+const generateCertificate = async (certificatePath, certificateDirPath) => {
+  const startTime = performance.now();
+
+  const pems = await createSelfSignedCertificate();
+
+  const certificateDirExists = await exists(certificateDirPath);
```
mattcompiles, line 76: You should use the sync version of exists as otherwise the file may/may not exist by the next tick. https://nodejs.org/api/fs.html#fs_fs_exists_path_callback

lib/certificate.js
```diff
@@ -0,0 +1,114 @@
...
+  return pems.private + pems.cert;
+};
+
+const getCertificate = async (certificateDirName = '.ssl') => {
+  const certificateDirPath = getPathFromCwd(`./${certificateDirName}`);
+  const certificatePath = getPathFromCwd(
+    `./${certificateDirName}/self-signed.pem`,
+  );
+
+  const certificateExists = await exists(certificatePath);
```
mattcompiles, line 95: Same here

test/test-cases/sku-with-https/sku-with-https.test.js
```diff
@@ -0,0 +1,100 @@
...
+    afterAll(async () => {
+      await process.kill();
+    });
+
+    it('should start a development server', async () => {
+      const snapshot = await getAppSnapshot(url);
+      expect(snapshot).toMatchSnapshot();
+    });
+    it('should support the supplied middleware', async () => {
+      const snapshot = await getAppSnapshot(`${url}/test-middleware`);
```
mattcompiles, line 52: Not sure `getAppSnapshot` is right for this use case. Maybe just validate it returns a 200?

context/defaultSkuConfig.js
willhackett: I'd like them to manually specify it for it to be activated.

context/index.js
willhackett: Where's the best place to throw an error?

```
if (devServerMiddleware && fs.existsSync(devServerMiddleware) === false) {
    throw new Error(`${devServerMiddleware} does not exist. Please create it and try again.`);
}
```

context/defaultSkuConfig.js
mattcompiles: > Place a `dev-middleware.js` file in the root of the project. The file must export a function that will receive the express server.

Docs seems to say that's the default file though. Maybe just make the docs clearer that you need to supply a path to a file then.

test/test-cases/sku-with-https/sku-with-https.test.js
willhackett: I just wanted to make sure the body was present — perhaps not necessary?

context/index.js
mattcompiles: The context file is fine. This seems fine to me 👍

test/test-cases/sku-with-https/sku-with-https.test.js
mattcompiles: Happy to leave it for now then.

PR #527, opened 2020-07-28
Add hot reloading

mattcompiles:
See changeset for details


config/webpack/plugins/createTreatPlugin.js
```diff
@@ -1,6 +1,18 @@
...
+  'jobStreet',
+  'jobStreetClassic',
+  'jobsDb',
+  'seekAnz',
+  'seekBusiness',
+  'seekUnifiedBeta',
+  'occ',
+  'catho',
+  'docs',
+  'wireframe',
```
mattcompiles, line 14: This is a bonus change to update the braid theme hashing

package.json
```diff
@@ -143,6 +144,7 @@
     "prettier": "2.0.2",
     "pretty-ms": "^6.0.1",
     "raw-loader": "^4.0.0",
+    "react-refresh": "^0.8.3",
     "react-treat": "^1.0.2",
```
michaeltaranto, line 148: Worth bumping the baseline versions of all treat deps i'd say.

scripts/start.js
```diff
@@ -61,6 +68,8 @@ const localhost = '0.0.0.0';
     stats: 'errors-only',
     allowedHosts: appHosts,
     serveIndex: false,
+    hot,
+    clientLogLevel: 'warn',
```
michaeltaranto, line 72: Was this debugging, or for better visibility of something specific?

mattcompiles: This actually hides logging rather than adds it. HMR is super noisy in the console, classic webpack DX.

.changeset/giant-sloths-yawn.md
```diff
@@ -0,0 +1,25 @@
...
+
+We're considering adding lint rules for these scenarios in future. 
+
+**NOTE:** React >16.9 is required for fast-refresh to work
+
+In some cases a change cannot be hot reloaded, in these situations sku should fallback to performing a full page refresh. You should **never** need to manually refresh your browser.
+
+If your app is not hot reloading when you would expect it to or you are being forced to manually refresh the page, please contact #sku-support.
+
+You can disable HMR by setting `SKU_HOT=false`.
```
michaeltaranto, line 24: Should this have config support as well?

mattcompiles: My thinking is not for now. I've put it there as an escape hatch in case we hit unexpected issues or you need to disable it for testing scenarios. Project wide disabling shouldn't need to be happening.

PR #535, opened 2020-09-21
Update Typescript and ESLint

mattcompiles:


mattcompiles:
Seems the `@typescript-eslint/explicit-module-boundary-types` rules is going to give a lot of trouble to react projects if they haven't been written that way from the beginning. It's only a warning as well which is mostly useless. Thoughts on disabling at the sku level? Or sshould we raise in `eslint-config-seek`?

jahredhope:
I can see the benefit of explicitly declaring your return types.
Does the rule allow you to cast the function to a `React.FunctionalComponent`?
```ts
import React, { FunctionComponent } from 'react';

const App: FunctionalComponent<props> = () => {}
```

If so, and we believe it's a good pattern I'd suggest changing from warning to error in `eslint-config-seek`.
Alternatively I don't mind disabling the rule for sku as React projects as it's common pattern to assume a React component boundary is a React component.

Basically anything other than "leave it a warning". IMO warnings should only be used to temporarily downgrade an error in a project whilst in the process of migrating to the new pattern.

72636c:
I find `@typescript-eslint/explicit-module-boundary-types` draconian even outside of React, and it's [sneakily disabled for skuba projects](https://github.com/seek-oss/eslint-config-skuba/blob/17146a43e21d9339521d5237fae306daf274660a/index.js#L42). I think I snuck that one in behind @etaoins though, who is the senior principal advocate for it

etaoins:
I'm fine disabling `@typescript-eslint/explicit-module-boundary-types` for Sku. It's pretty full on.

markdalgleish:
I vote to remove it, seems a bit heavy handed.

jahredhope:
If skuba and sku both disable then lets disable in `eslint-config-seek`?

mattcompiles:
PR updated to use `eslint-config-seek@7.0.5`

mattcompiles:
This PR is ready to go now. CI finally passing.

72636c (approved):
🤩

markdalgleish (approved):

PR #560, opened 2021-01-05
Compile vocab files before lint and test commands

jahredhope:


markdalgleish:
It seems like this runs even if your project isn't configured for multiple languages—is that right?

markdalgleish:
I'm wrong. Ignore me. Just saw that we bail out of `getVocabConfig` if languages aren't configured.

markdalgleish (approved):

PR #572, opened 2021-02-01
Replace $language in route when using dev server

jahredhope:
Explanation in changeset


lib/language-utils.js
```diff
@@ -81,4 +81,12 @@ function getLanguageFromRoute(req, route) {
   return supportedLanguagesForRoute[0];
 }
 
-module.exports = { getLanguageFromRoute, getValidLanguagesForRoute };
+function getRouteWithLanguage(routePath, language) {
+  return routePath.replace('$language', language);
```
markdalgleish, line 85: Should this be a global replace?

mattcompiles (approved):

Issue #594, opened 2021-05-28
Update Braid and remove deprecated components from template


tadhglewis:
When setting up a project using sku, the template is using BulletList and Bullet, these are deprecated - https://seek-oss.github.io/braid-design-system/components/BulletList

Todo
- [x] update braid-design-system in sku
- [x] Follow migration guide for https://seek-oss.github.io/braid-design-system/components/List/releases to replace BulletList

jahredhope:
Hi @tadhglewis , great issue and thank-you for calling this out.

I agree the template should be updated, and probably needs a good looking over for anything else that might have changed.

Given the `Todo` list in this Issue have you already got a PR in mind? Otherwise I'll work on a change myself.

tadhglewis:
@jahredhope I was gonna work on a PR although as you mentioned, needs a good look over.

I can put through a PR with anything I find and if I get stuck with any other changes that need doing with the latest Braid version I'll open up a draft PR for someone to continue on.

Thoughts?

jahredhope:
Sounds good, thanks

PR #595, opened 2021-05-28
Update Braid and remove deprecated components from template

tadhglewis:
https://github.com/seek-oss/sku/issues/594


tadhglewis:
@markdalgleish thanks for adding that. i had a bit of a look and that should be all that's required for updating braid.

I was having trouble running unit tests and had a fair amount failing. Seems like there are a lot of flaky tests? any tips?

jahredhope:
There is one known flakey test. But only one and only in CI, the one that tests the `init` script. Which is my bad, something about CI doesn't like spawning and writing to STIDIN.

If you are seeing flakey tests failing outside of the `init` script that's curious.
I wouldn't expect any others to be flakey. We'll pull it down the branch today and see whats going on.

mattcompiles (approved):
Thanks for the PR 🍻

tadhglewis:
@mattcompiles @markdalgleish thanks for finishing that one off

PR #609, opened 2021-07-19
Move to automatic React runtime

mattcompiles:


config/eslint/eslintConfig.js
```diff
@@ -6,6 +6,10 @@ const baseConfig = {
   extends: orderImports
     ? [coreConfig, require.resolve('./importOrderConfig')]
     : coreConfig,
+  rules: {
+    'react/jsx-uses-react': 'off',
+    'react/react-in-jsx-scope': 'off',
```
mattcompiles, line 11: Currently disabling these rules inside sku itself, but we should probably lift this into `eslint-config-seek` once these changes have stabilised.

michaeltaranto (approved):

PR #610, opened 2021-07-20
Fix static rendering of vanilla-extract classes in production

mattcompiles:


config/webpack/webpack.config.js
```diff
@@ -291,6 +291,9 @@ const makeWebpackConfig = ({
         library: 'static',
         libraryTarget: 'umd2',
       },
+      optimization: {
+        nodeEnv: process.env.NODE_ENV,
```
mattcompiles, line 295: This ensures the `NODE_ENV` is set correctly. Otherwise it defaults to the webpack mode which is always `development` for the render config.

michaeltaranto (approved):

jahredhope (approved):

PR #616, opened 2021-08-04
Upgrade to webpack 5 

mattcompiles:


config/webpack/plugins/createTreatPlugin.js
```diff
@@ -17,11 +17,9 @@ const braidThemes = [
 module.exports = ({
   target,
   isProductionBuild,
-  include,
```
michaeltaranto, line 20: This is still being passed through at the [call site](https://github.com/seek-oss/sku/pull/616/files#diff-4e4806abd6e6dec03e7f50c209f69b659d0ceeca5fae03a88e3c40876b6e1de5L193)

config/webpack/plugins/sku-webpack-plugin/index.js
```diff
@@ -163,28 +162,42 @@ class SkuWebpackPlugin {
...
-      compiler.options.resolve.extensions.push('.ts');
-    }
+    if (!compiler.options.resolve.extensions) {
+      compiler.options.resolve.extensions = [
+        '.mjs',
+        '.js',
+        '.json',
+        '.ts',
+        '.tsx',
+      ];
```
michaeltaranto, line 192: I know this is pretty standard, but would it be better to share this list between the webpack config and here to make sure they are in-sync?

config/webpack/webpack.config.js
```diff
@@ -154,13 +144,15 @@ const makeWebpackConfig = ({
       optimization: {
         nodeEnv: process.env.NODE_ENV,
         minimize: isProductionBuild,
+        minimizer: [new TerserPlugin({ extractComments: false })],
```
michaeltaranto, line 147: GIven their api is a little misleading, worth adding a comment that this is to preserve inline license comments?

michaeltaranto (approved):
Massive piece of work 👏 !

jahredhope (approved):

PR #632, opened 2021-08-17
Fix svgo plugin config

benjervis:
Fixes the logging issues that were coming up with new sku.

We don't need to have svgo installed anymore, so it can just be a subdependency for svgo-loader.

### DIFF
--- a/.changeset/stupid-cars-push.md
+++ b/.changeset/stupid-cars-push.md
@@ -1,0 +1,5 @@
+---
+'sku': patch
+---
+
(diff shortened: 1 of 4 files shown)


jahredhope (approved):

mattcompiles (approved):
Nice one. We might need to get some people to lockfile refresh as we can't control what version they have now and it's possible they'll get a version prior to v2.4.0.

benjervis:
I'll add that guidance to the changeset

PR #634, opened 2021-08-19
docs: Fix link to v11 release notes

jackhurley23:
Tag syntax is different from the v10 release. point migration guide link to v11.0.0 tag


mattcompiles (approved):
Thanks bud

PR #656, opened 2021-11-30
docs: Fix languages type in config docs

jackhurley23:
Config doc has incorrect type for the `languages` field

Valid config is shown here:
https://github.com/seek-oss/sku/blob/c303ec3705eccbb68c70d5add2d15a1667cf6079/context/configSchema.js#L8-L14

It's also valid to pass `{ name: 'en' }` as an option in `languages`

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -201,7 +201,7 @@ The browser URL to open when running `sku start` or `sku start-ssr`. It will def
 
 ## languages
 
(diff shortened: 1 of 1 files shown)


zendy (approved):

benjervis:
Hey Jack,

Thanks for the PR! It clashes slightly with another change that's on the way though, so we've rolled your changes into #657

PR #670, opened 2022-02-22
Fix static rendering for routes using `:language`

mattcompiles:


jahredhope (approved):

.changeset/old-tips-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+Fix static rendering for routes using ':language'
```
jahredhope, line 5:
```suggestion
Support static rendering for routes using ':language'
```

PR #675, opened 2022-04-06
Generate pseudo language

askoufis:


test/test-cases/translations/__snapshots__/translations.test.js.snap
```diff
@@ -2,9 +2,9 @@
 
 exports[`translations should render en 1`] = `
```
mattcompiles, line 3: Maybe worth adding a test for `en-PSEUDO`?

mattcompiles (commented):
Looks good mate. Maybe just add a test for the new language?

mattcompiles (approved):

PR #687, opened 2022-05-20
`sku serve` fix

benjervis:
See changeset

Also fixed a variable typo


test/test-cases/multiple-routes/app/sku.config.js
```diff
@@ -5,14 +5,14 @@ export default {
       host: 'dev.seek.com.au',
       routes: [
         { route: '/', name: 'home' },
-        { route: '/details/$id', name: 'details' },
+        { route: '/details/:id', name: 'details' },
```
askoufis, line 8: Should we keep these as `$`?

test/test-cases/multiple-routes/__snapshots__/multiple-routes.test.js.snap
```diff
@@ -227,7 +227,7 @@ SOURCE HTML: <!DOCTYPE html>
     </script>
   </body>
 </html>,
-  "production/nz/nz/details/$id/index.html": SCRIPTS: Array [
+  "production/nz/nz/details/:id/index.html": SCRIPTS: Array [
```
askoufis, line 230: This too.

test/test-cases/multiple-routes/app/sku.config.js
askoufis: Actually, maybe worth keeping one as `$` and the other as `:` so this config tests both at once.

.changeset/breezy-wolves-joke.md
```diff
@@ -0,0 +1,10 @@
+---
+'sku': patch
+---
+
+Allow `:` to be used in dynamic paths again.
+
+Previously, dynamic paths were declared using the standard `:param` syntax, but this had been deprecated in favour of `$`.
+
+This has now been updated to allow for both.
+This should allow `sku serve` to work for projects using colon syntax.
```
askoufis, line 10: Feel like this is a bit misleading.`:` still works, and always has, but now sku serve supports it, instead of throwing an error.

benjervis: [The documentation](https://seek-oss.github.io/sku/#/./docs/static-rendering?id=dynamic-routes) is pretty clear that the syntax is `$`. It might have worked by accident before, but I don't think there's anything wrong with saying "this is now expressly permitted".

askoufis (approved):

PR #702, opened 2022-09-20
Add `delete-unused-keys` flag to `translations push` command

askoufis:
- Expose the feature added in https://github.com/seek-oss/vocab/pull/93 via `sku translations push --delete-unused-keys`.
- Handle the case where no sub-command is passed to `sku translations` by throwing an error with a message containing all available commands

The CLI arg parsing is kinda janky for `sku translations`. If it gets any more complicated it might be worth using yargs or something.

### DIFF
--- a/.changeset/polite-pigs-sing.md
+++ b/.changeset/polite-pigs-sing.md
@@ -1,0 +1,11 @@
+---
+'sku': minor
+---
+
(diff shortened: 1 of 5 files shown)


askoufis:
Tested a snapshot release of this branch in the `candidate-demo-app`. Deleted only keys that no longer exist (one of the translations moved folders, but as far as Phrase is concerned this is a new key).
image

michaeltaranto (approved):

PR #703, opened 2022-09-25
Add changesets snapshot workflow

askoufis:


.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Set up Node.js 12.x
+        uses: actions/setup-node@v3
+        with:
+          node-version: 12.x
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Publish
+        uses: seek-oss/changesets-snapshot@v0
```
mrm007, line 26: Should we release a `v1` branch now that it's widely used?

mrm007 (approved):

.github/workflows/snapshot.yml
askoufis: I guess we could. Feels weird to go to v1 just on API stability though, rather than new features.

PR #709, opened 2022-10-07
Update `jest`, `typescript`, `babel`, `eslint-config-seek` and minimum node version requirement

askoufis:
See changesets.

Note: `eslint` was not updated because `eslint-plugin-cypress` is not compatible with `eslint` 8. [It's being moved to the cypress repo soon](https://github.com/cypress-io/eslint-plugin-cypress/issues/114), so will hopefully be updated to support after that's complete.


mrm007:
Looks like one snapshot needs to be updated.

Unrelated, seems like we're running **Validate / Lint & Test** on both `pull_request` and `push`. I've seen this solved with 

```yaml
on:
  pull_request:
  push:
    branches: [master]
```

with the caveat it doesn't run on push _before_ opening a PR. I don't necessarily see this as a downside — if you want actions to run, open a draft PR.

package.json
```diff
@@ -152,7 +153,7 @@
     "traverse": "^0.6.6",
     "treat": "^2.0.4",
     "tree-kill": "^1.2.1",
-    "typescript": ">=4.1.3 <4.5.0",
+    "typescript": "^4.5.0",
```
mrm007, line 156: This [goes](https://github.com/seek-oss/sku/blob/b03adb48391fafb44fd7cfa8c34d73dd952f5d4f/yarn.lock#L16975-L16976) straight to latest, `4.8.4`. Do we want to use `~4.5.0` to stay on `4.5.x`?

askoufis: Skuba does this, though I'm not exactly sure of the benefit in the case of sku. If anything, I feel like we'd want to restrict the upper limit, since typescript doesn't adhere to semver, just in case something breaks, but I'm not sure about restricting to a specific minor version just yet.

michaeltaranto (approved):
👏

PR #710, opened 2022-10-11
More dependency updates

askoufis:
This should be the last batch of updates before we can release. I've updated the node version changeset (not snapshot as the commit message says).

Updated packages:
- storybook
- webpack
- a bunch of webpack loaders
- lodash
- a few babel plugins

The webpack updates have shuffled around some files in the snapshots, but the changes LGTM.

I also de-duplicated the lockfile.


.changeset/swift-sloths-relax.md
```diff
@@ -5,4 +5,5 @@
 Drop support for Node v12
 
 Sku now only supports Node v14.15 and above.
-Although sku itself does not depend on any Node v14 APIs, Node v12 is [no longer officially supported](https://github.com/nodejs/Release#end-of-life-releases), and many of sku's dependencies no longer support it either.
\ No newline at end of file
+Although sku itself does not depend on any Node v14 APIs, Node v12 is [no longer officially supported](https://github.com/nodejs/Release#end-of-life-releases), and many of sku's dependencies no longer support it either.
+Please ensure you are using a supported version of Node both locally (e.g. via a `.nvmrc` file) and in CI (check your docker images).
```
mrm007, line 9:
```suggestion
Please ensure you are using a supported version of Node both locally (e.g. via a `.nvmrc` file) and in CI (check your Docker images).
```

test/test-cases/braid-design-system/__snapshots__/braid-design-system.test.js.snap
```diff
@@ -2284,19 +2284,19 @@ svg.jhxt971g {
...
-  "url",
-  "util"
+  "external node-commonjs \\"crypto",
+  "external node-commonjs \\"fs",
+  "external node-commonjs \\"os",
+  "external node-commonjs \\"path",
+  "external node-commonjs \\"stream",
+  "external node-commonjs \\"tty",
+  "external node-commonjs \\"url",
+  "external node-commonjs \\"util"
```
mrm007, line 2294: Might be nothing, but these look like unterminated strings.

mrm007 (approved):

test/test-cases/braid-design-system/__snapshots__/braid-design-system.test.js.snap
askoufis: Found the culprit https://github.com/seek-oss/sku/blob/master/test/utils/ListExternalsWebpackPlugin.js.
I think something has changed, will try tweak it.

.changeset/swift-sloths-relax.md
askoufis: #shitoberfest

PR #714, opened 2022-10-21
Fix transpilation of typescript-specific class keywords

askoufis:
Something in `@babel/babel-plugin-transform-typescript@7.18.0` broke transpilation of typescript-specific class keywords. Users could run into this issue when updating to `sku@11.6.0` because the version range of `@babel/babel-plugin-transform-typescript` was updated.

Something like:
```ts
export class MyClass {
  constructor(protected foo: number) {}
}
```

is being transpiled to:
```js
var MyClass = /*#__PURE__*/_createClass(function MyClass(protected foo) {
//                               not valid javascript -> ^^^^^^^^^
  _classCallCheck(this, MyClass);
});
```
`protected` is a typescript-specific keyword, so this is invalid code.

I'm unsure as to why this has suddenly become an issue in that version of the babel plugin, but it turns out that [babel presets are executed in reverse](https://babeljs.io/docs/en/presets#preset-ordering), and we had `preset-typescript` before `preset-env` in the `presets` array, meaning `preset-tyepscript` was executing _after_ `preset-env`, but `preset-env` doesn't understand the `protected` keyword (because it's typescript), so 💥 .

This PR fixes the ordering of presets so `preset-typescript` executes before `preset-env`, and adds a test class in one of the test cases with a typescript file because there isn't really a good place to put it (suggestions welcome) and I didn't want to make the test run any longer than it already is.

### DIFF
--- a/.changeset/tiny-walls-wink.md
+++ b/.changeset/tiny-walls-wink.md
@@ -1,0 +1,5 @@
+---
+'sku': patch
+---
+
+Fix transpilation of typescript-specific class keywords
--- a/config/babel/babelConfig.js
+++ b/config/babel/babelConfig.js
@@ -58,6 +58,13 @@ module.exports = ({
   }
 
   const presets = [
+    [
+      require.resolve('@babel/preset-env'),
+      {
+        targets: browserslist,
+        shippedProposals: true,
+      },
+    ],
     lang === 'ts'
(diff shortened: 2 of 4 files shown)


mrm007 (approved):
Nice work 🕵️

PR #727, opened 2022-12-09
Add package.json configuration flags for disabling specific functionality when running scripts

askoufis:
Braid no longer relies on sku for its configuration files (e.g. eslintrc, tsconfig, etc.), and currently [patches sku](https://github.com/seek-oss/braid-design-system/blob/5ab32f96f9ee144383f5926773ca3e3656c5f82d/patches/sku%4011.6.2.patch) to skip the sku configuration step, as well as peer dep validation.

This PR adds checks for configuration present in `package.json` that allows you to disables these steps. We do this already for [skipping post install](https://github.com/seek-oss/sku/blob/2e8a088336ff66fa272602b524fa7a78313bafaa/scripts/postinstall.js#L26), and I felt that these configurations also fit better in `package.json` rather than in the sku config.
```jsonc
// package.json
{
  "skuSkipConfigure": true,
  "skuSkipPeerDepValidation": true
}
```


bin/sku.js
```diff
@@ -13,9 +15,38 @@ if (args.debug) {
...
+}
+
 const configureProject = async () => {
-  const configure = require('../lib/configure');
-  await configure();
+  if (!skipConfigure) {
+    const configure = require('../lib/configure');
+    await configure();
+  } else {
+    log(`"skuSkipConfigure" set in ${packageJson}, skipping sku cofiguration`);
```
michaeltaranto, line 37:
```suggestion
    log(`"skuSkipConfigure" set in ${packageJson}, skipping sku configuration`);
```

bin/sku.js
```diff
@@ -13,9 +15,38 @@ if (args.debug) {
...
+  }
+};
+
+const validatePeerDeps = () => {
+  if (!skipValidatePeerDeps) {
+    // Intentionally not awaiting async function as it's just for logging warnings
+    _validatePeerDeps();
+  } else {
+    log(
+      `"skipValidatePeerDeps" set in ${packageJson}, skipping sku peer dep validation`,
```
michaeltaranto, line 47:
```suggestion
      `"skuSkipValidatePeerDeps" set in ${packageJson}, skipping sku peer dependency validation`,
```

michaeltaranto (approved):

bin/sku.js
```diff
@@ -13,9 +15,38 @@ if (args.debug) {
...
+
 const configureProject = async () => {
-  const configure = require('../lib/configure');
-  await configure();
+  if (!skipConfigure) {
+    const configure = require('../lib/configure');
+    await configure();
+  } else {
+    log(`"skuSkipConfigure" set in ${packageJson}, skipping sku configuration`);
+  }
```
mrm007, line 38: Not a fan of early exits [like over here](https://github.com/seek-oss/sku/blob/aeeb5a7e59c3ce75670911c6c5216aab57dddc47/config/vocab/vocab.js#L6-L9)?

```suggestion
  if (skipConfigure) {
    log(`"skuSkipConfigure" set in ${packageJson}, skipping sku configuration`);
    return;
  }
  const configure = require('../lib/configure');
  await configure();
```

bin/sku.js
```diff
@@ -13,9 +15,38 @@ if (args.debug) {
...
+
+const validatePeerDeps = () => {
+  if (!skipValidatePeerDeps) {
+    // Intentionally not awaiting async function as it's just for logging warnings
+    _validatePeerDeps();
+  } else {
+    log(
+      `"skuSkipValidatePeerDeps" set in ${packageJson}, skipping sku peer dependency validation`,
+    );
+  }
```
mrm007, line 49: Same early exit suggestion:

```suggestion
  if (skipValidatePeerDeps) {
	  log(
	    `"skuSkipValidatePeerDeps" set in ${packageJson}, skipping sku peer dependency validation`,
  	);
  }
  // Intentionally not awaiting async function as it's just for logging warnings
  _validatePeerDeps();
```

PR #730, opened 2023-01-27
Proof of concept: support storybook middleware

tadhglewis:
The problem: I need a api endpoint to generate auth tokens for local development. With sku I can simply use dev server middleware however this isn't possible when running storybook.

Storybook actually [supports a router middleware](https://github.com/storybookjs/storybook/pull/435) since 2016 however it isn't currently documented as per https://github.com/storybookjs/storybook/issues/15300#issuecomment-866469521 so it 
might be debatable whether we want to support it in sku.

There is probably a larger discussion to be had around a better implementation for storybook as configuring storybook through `.storybook/*` directory isn't supported in sku projects - talked to @askoufis [here](https://seekchat.slack.com/archives/CDL5VP5NU/p1674088704705369?thread_ts=1674064947.112759&cid=CDL5VP5NU)

Todo
- [ ] fix broken tests
- [ ] check if this is something we should actually support
- [ ] make sure it won't break existing storybook projects

### DIFF
--- a/config/storybook/start/middleware.js
+++ b/config/storybook/start/middleware.js
@@ -1,0 +1,3 @@
+const { paths } = require('../../../context');
+
+module.exports = require(paths.devServerMiddleware);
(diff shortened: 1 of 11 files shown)


tadhglewis:
Closing in favor of supporting supporting config dir or dangerously override

PR #732, opened 2023-02-09
Upgrade webpack deps in `yarn.lock`

askoufis:
Upgraded webpack deps with `yarn upgrade webpack`. Slowly updating parts of the lockfile because doing it in one go results in the weird "content hash changed but no content changed, only on CI" bug that Matt ran into months ago, so trying to isolate which dependency is causing that.


michaeltaranto (approved):

test/test-cases/storybook-config/app/.eslintrc
```diff
@@ -0,0 +1,7 @@
+/** THIS FILE IS GENERATED BY SKU, MANUAL CHANGES WILL BE DISCARDED **/
+{
+  "extends": [
+    "/Users/askoufis/code/fep/sku/node_modules/eslint-config-seek/.eslintrc.js",
```
mrm007, line 4: You might want to `.gitignore` this file

askoufis: Oh woops, this is a file from another branch. Will fix it in the other PR.

PR #735, opened 2023-02-10
Refresh a bunch of deps in `yarn.lock`

askoufis:
Ran `yarn upgrade --pattern "webpack|babel|@vocab|css|@vanilla-extract|eslint|postcss" && npx yarn-deduplicate`

### DIFF
--- a/test/test-cases/lint-rules/__snapshots__/import-order.test.js.snap
+++ b/test/test-cases/lint-rules/__snapshots__/import-order.test.js.snap
@@ -27,9 +27,9 @@ import c from '../c';
 import a from './a';
 import b from './b';
 
-import cTreat from '../c.treat';
 import aTreat from './a.treat';
 import bTreat from './b.treat';
+import cTreat from '../c.treat';
 
 `;
 
@@ -39,8 +39,8 @@ import c from '../c';
 import a from './a';
 import b from './b';
 
-import cStyle from '../c.css';
 import aStyle from './a.css';
 import bStyle from './b.css';
+import cStyle from '../c.css';
 
 `;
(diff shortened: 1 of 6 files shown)


test/test-cases/lint-rules/__snapshots__/import-order.test.js.snap
```diff
@@ -27,9 +27,9 @@ import c from '../c';
 import a from './a';
 import b from './b';
 
-import cTreat from '../c.treat';
 import aTreat from './a.treat';
 import bTreat from './b.treat';
+import cTreat from '../c.treat';
```
mrm007, line 32: Looks like we need to change _something_ in `eslint-config-seek` 🤔

askoufis: Isn't this likely a result of a change in our eslint config? https://github.com/seek-oss/sku/pull/735/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR7341
https://github.com/seek-oss/eslint-config-seek/releases/tag/v10.1.3

mrm007: Import ordering is a sku feature: 
https://github.com/seek-oss/sku/blob/cf67d47f0e109dafb0541a05c50311f56bd5baa9/config/eslint/importOrderConfig.js#L17-L39

askoufis: Weird that this changed then, unless something changed in how the eslint rule itself behaves.

mrm007: I think this is fine.

mrm007 (approved):

PR #741, opened 2023-02-24
Fix root resolution within `.css.ts` files during tests

askoufis:
Replacing the [VE babel plugin](https://github.com/seek-oss/sku/pull/738/files#diff-10b70ee3d83bd30c0159313993c58ca5a8595009c84491159ac8f9aa589a4a44L27) with the [VE jest transform](https://github.com/seek-oss/sku/pull/738/files#diff-cd013def59a6ab8f5f61b6f4af62a0ae3a6986fce54db6bdcca140108109002cR37-R38) resulted in `.css.ts` files no longer going through our babel transformation stack during jest tests. This ended up breaking root resolution (importing from `src/*`) within `.css.ts` files during tests.

Luckily, jest has a config value just for this: [`modulePaths`](https://jestjs.io/docs/configuration#modulepaths-arraystring). Setting this to the result of `cwd` when `rootResolution` is true (like we do in [the babel transform stack](https://github.com/seek-oss/sku/blob/f1c41a126ee1a131a8cbdf6ce2cf0eac82fd370a/config/babel/babelConfig.js#L20)) fixes the issue.

### DIFF
--- a/config/jest/jestConfig.js
+++ b/config/jest/jestConfig.js
@@ -1,5 +1,6 @@
 const escapeRegex = require('escape-string-regexp');
-const { paths } = require('../../context');
+const { cwd } = require('../../lib/cwd');
+const { paths, rootResolution } = require('../../context');
 const slash = '[/\\\\]'; // Cross-platform path delimiter regex
 const compilePackagesRegex = paths.compilePackages
   .map((pkg) => `.*${escapeRegex(pkg)}`)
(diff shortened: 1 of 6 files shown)


benjervis (approved):

config/jest/jestConfig.js
```diff
@@ -10,6 +11,7 @@ module.exports = {
   testEnvironment: 'jsdom',
   setupFilesAfterEnv: paths.setupTests,
   prettierPath: require.resolve('prettier'),
+  modulePaths: rootResolution ? [cwd()] : undefined,
```
mrm007, line 14: Does this work?

```suggestion
  modulePaths: rootResolution ? ['<rootDir>'] : undefined,
```

michaeltaranto (approved):

config/jest/jestConfig.js
askoufis: It probably does, but I wanted to keep it consistent with the babel transform root, which we get from the result of `cwd`. I think if you're running `sku test` via your package manger, this will end up being your project root anyway.

PR #744, opened 2023-03-08
Migrate to PNPM, make the repo a monorepo

askoufis:
Sku does weird stuff to set up dependencies for tests. Each test case is almost like its own test fixture, except the test setup creates a sort of "fake" `node_modules` to get everything working. A monorepo with fixture packages is a natural fit for these kinds of e2e tests, so I wanted to see how much effort it would be to convert sku to use PNPM, and in the process become a monorepo.

It turned out to be not as much as effort I thought.

Notes:
- The `braid-design-system` and `sku-webpack-plugin` tests have been migrated to their own fixture package within the `fixtures` folder. They were migrated because they were having issues with the node modules copying/symlinking shenanigans that were going on in those tests. Turning them into proper fixture packages with explicit dependencies (and hence their own `node_modules`) has fixed this completely, and allowed me to remove the setup code they had. The actual tests themselves are untouched.
- All other tests worked without issue, so they've been left as-is for now
- The `virtual-resource-loader` package needs to be hoisted because `treat` can't resolve it correctly unless it's hoisted. This was fixed quite easily [in vocab](https://github.com/seek-oss/vocab/releases/tag/%40vocab%2Fwebpack%401.1.3), so we could push a fix to `treat`, but it's deprecated so I'd rather leave it alone. Regardless, I'm happy that we didn't need to hoist everything.

Future work:
- Migrate all tests to fixtures
- Move test utils into their own package
- Potentially remove the `test-manual` script and the `runSkuScriptInDir` function as you can just run `sku <script>` in the fixture folder
- I've tested running both of fixture tests at the same time locally, and they pass without issue. This means there's a potential opportunity to remove `--detectOpenHandles --runInBand` from the test script, meaning we might be able to test sku _way_ faster soon :tm:.


michaeltaranto (approved):
Epic! 👏

PR #746, opened 2023-03-09
Convert some more tests to fixtures

askoufis:
See commit messages.


tests/library-build.test.js
```diff
@@ -0,0 +1,41 @@
...
+const dirContentsToObject = require('../test/utils/dirContentsToObject');
+const waitForUrls = require('../test/utils/waitForUrls');
+const runSkuScriptInDir = require('../test/utils/runSkuScriptInDir');
+const { getAppSnapshot } = require('../test/utils/appSnapshot');
+const appDir = path.dirname(
+  require.resolve('@fixtures/library-build/sku.config.js'),
+);
+const distDir = path.resolve(appDir, 'dist');
+
+describe('library-build', () => {
```
askoufis, line 11: The only thing that's different between the new and [old version](https://github.com/seek-oss/sku/pull/746/files#diff-6df529ed30cac7e069286e3e3ce7d44400fcaee7c7052c6f5dc1e83c2c1046d6) of this file is the change to how we derive `appDir` and the removal of setup code that's no longer needed.

michaeltaranto (approved):

PR #747, opened 2023-03-13
test: Convert more test cases to fixtures

askoufis:
I have converted a few more tests to fixtures.

EDIT: Tests have started to take quite a while on CI, to the point that the github actions runner times out (we set a 30m limit somewhere). I've now made two separate test scripts, `test:concurrent` and `test:sequential`, the main difference between them being the absence of `--runInBand` and `--detectOpenHandles` on `test:concurrent`, as well as the tests they each run.

TL;DR: `test:sequential` is running the remaining test cases in `/test`, `test:concurrent` is running everything else. This is a temporary measure until everything is a fixture.

The result of this is that tests are quite fast on my macbook, but linux runners [only have 2 cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) (maybe we could swap to macOS runners for an extra core, or potentially [large runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#machine-specs-for-larger-runners) if we have access), so there's not much benefit on CI (relative to before I started migrating these test cases to fixtures), other than it not taking 30 minutes (more like 15-20). This might improve once the remaining test cases are converted to fixtures, though I'm not holding my breath.


fixtures/custom-src-paths/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@fixtures/custom-src-paths",
+  "private": true,
+  "main": "./main.js",
```
michaeltaranto, line 4: Can this one be removed too? Think we are only requiring the sku config by path

askoufis: Ah, missed this one.

fixtures/typescript-css-modules/src/render.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { renderToString } from 'react-dom/server';
 import dedent from 'dedent';
 
-import { Render } from '../../../../../sku-types';
+import { Render } from 'sku';
```
michaeltaranto, line 5: Did the type-only import rules not get promoted to the general eslint config? Would have expected this to be:
```suggestion
import type { Render } from 'sku';
```

tests/typescript-css-modules.test.js
```diff
@@ -144,7 +146,7 @@ describe('typescript-css-modules', () => {
       await page.goto(storybookUrl, { waitUntil: 'networkidle2' });
 
       const content = await page.evaluate(async () => {
-        const element = await window.document
+        const element = window.document
```
michaeltaranto, line 149: Nit-pick: Removing `await` here makes it no longer an `async` function

fixtures/typescript-css-modules/src/render.tsx
askoufis: `eslint-config-seek` doesn't actually enforce this rule. Probably worth updating, I've building a list of eslint rules I think we should add to it, I'm sure Remus has a few more too.

michaeltaranto (approved):

PR #750, opened 2023-03-20
Move fixture packages to `@sku-fixtures` scope

askoufis:
We own the `@sku-fixtures` scope, so it's safer to put private packages in this scope. All the classnames in the snapshots changed, hence the large diff.


askoufis:
I've tried to update [the failing snapshot](https://github.com/seek-oss/sku/actions/runs/4464390658/jobs/7840492292#step:11:569) locally, but it says there's no changes. :(

askoufis:
> I've tried to update [the failing snapshot](https://github.com/seek-oss/sku/actions/runs/4464390658/jobs/7840492292#step:11:569) locally, but it says there's no changes. :(

Clearing the jest cache with `jest --clearCache` fixed this.

michaeltaranto (approved):

PR #753, opened 2023-03-21
Package dependency cleanup, move docs deployment script to docs package

askoufis:
- Renamed the docs package to `@sku-private/docs`, and made it private. Also moved the deployment script to that package so it now owns its own deployment.
- Cleaned up the top-level gitignore, the fixtures themselves now handle their own gitignores, so there's no need to keep the top-level sku-configured stuff. `dist-ssr` needs to be explicitly added for one of the fixtures that outputs both to `dist` and `dist-ssr`.
- Removed a few top-level dependencies, moving them to the actual packages they're used in (`@testing-library/react` was already declared as a dependency elsewhere)
- Removed the `test/` entry from `.npmignore` as that folder no longer exists

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,16 +3,9 @@ node_modules/
 package-lock.json
 .idea
 entry/render/startConfig.json
-dist-start/
+dist/
 dist-build/
+dist-ssr/
+dist-start/
 .eslintcache
 yarn.lock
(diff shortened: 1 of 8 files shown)


mrm007 (approved):
🧹🧹🧹

PR #754, opened 2023-03-24
Move sku to a package within the monorepo

askoufis:
Currently the `sku` package is also the monorepo root. This is generally a weird pattern as far as monorepos go. In addition, pnpm's symlinked `node_modules` causes an infinite glob loop when [validating peer dependencies](https://github.com/seek-oss/sku/blob/master/lib/validatePeerDeps.js#L26) in fixtures (because fixtures link to sku which contains fixtures which link to sku...). This infinite glob loop was the reason we had to set `SKU_FORCE_EXIT=true` when running tests, because the validation is [intentionally not awaited](https://github.com/seek-oss/sku/blob/275752bd3066e52885d461ef5bcd953aaac3bfff/bin/sku.js#L50).

This doesn't really cause any issues on an M1 Pro macbook, but the build agents, with their measly 2 cores, don't like it at all. A max depth can be set on this glob, but I'm not sure what a good value for the depth is, as nested `node_modules` _are_ valid, to a certain extent. Ultimately, fixing the monorepo structure by moving `sku` to its own package within the repo is a better solution, and that's what this PR does.

Consequently, both local and CI tests have gotten significantly faster (about 33% faster on my macbook, probably more on CI (and _even more_ if we get a jest cache hit!)), both due to the structural change _and_ the fact that we can now set `--maxWorkers=2` on CI when running tests.

The primary changes in this PR are:
- Moving all the sku files to `packages/sku` (this is the bulk of the changes)
- Splitting out sku's dependencies into its own `package.json`
- Updating references to sku code from tests and test-utils. Arguably these packages could depend on sku, but in practice it doesn't really matter, and it just causes lots of sku's peer dep validation warnings to appear, which is just noise in the case of this repo.
- Hoist a few dependencies so things work 😢 (this could potentially be fixed in the future)
- Update snapshots
- Add a few missing dependencies to some fixtures as they were depending on top-level repo deps (which have moved now that the repo structure has changed)


packages/sku/README.md
```diff
@@ -0,0 +1 @@
+../../README.md
```
askoufis, line 1: This is a symlink to the root readme. However, this results in the readme no being present in the `sku` package. I'm not too fussed about this, but if we want to keep the readme in the built package, we could reverse the symlink, just copy paste, or maybe have different readmes bewteen the package and the monorepo root.

mrm007: Given that sku [sub-package] is the main character, I think we should reverse the symlink.

.changeset/config.json
```diff
@@ -4,5 +4,6 @@
   "commit": false,
   "linked": [],
   "access": "public",
+  "ignore": ["@sku-fixtures/*", "@sku-private/*"],
```
mrm007, line 7: dark side

packages/sku/README.md
askoufis: The logo link is relative, so it'll only work in one place, not both 😢. Could just hardcode it to the raw URL, but that's not ideal.

askoufis: Actually if it's just relative it just needs to change to be from the sku package instead. 🤦

askoufis: Github doesn't know how to resolve it I guess
image

mrm007: I'd be ok with having the full path to the file i.e. `https://github.com/seek-oss/sku/raw/master/docs/logo/logo.png?raw=true`

askoufis:
To make sure the re-location didn't cause issues with the built package I tested a snapshot in a few SEEK apps and they all worked fine.

jahredhope (changes requested):
README links aren't correct. (Please dismiss once changed)

kjots (approved):

packages/sku/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+```bash
+$ npm install -g npx
+```
+
+## [Documentation](https://seek-oss.github.io/sku)
+
+## Contributing
+
+Refer to [CONTRIBUTING.md](./CONTRIBUTING.md). If you're planning to change the public API, please [open a new issue](https://github.com/seek-oss/sku/issues/new) and follow the provided RFC template in the [GitHub issue template](.github/ISSUE_TEMPLATE.md).
```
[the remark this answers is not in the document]
askoufis: https://github.com/seek-oss/sku/pull/754/commits/1c9a3ec47887d6ac1e88247a3cbfbc6cac63755d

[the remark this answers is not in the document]
jahredhope: Is it worth a snapshot just to make sure it works well on npm?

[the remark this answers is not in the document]
askoufis: https://github.com/seek-oss/sku/actions/runs/4548218626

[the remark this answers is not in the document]
askoufis: Pushed to a separate branch to avoid having to clean up a test changeset in this branch.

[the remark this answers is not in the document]
askoufis: https://www.npmjs.com/package/sku/v/0.0.0-sku-package-snapshot-test-20230328222730

PR #755, opened 2023-03-30
Add `libraryFile` configuration option

kjots:
- Add a new option, `libraryFile`, to allow the file name of a library to be specified in the sku configuration.
- Fallback to the existing behaviour of using `libraryName` for file name of the library if `libraryFile` is not specified.
- Update types and documentation for new configuration option.
- Add test case and fixture for new configuration option.


fixtures/library-file/src/styles.treat.ts
```diff
@@ -0,0 +1,9 @@
+// @ts-ignore
+import { style } from 'sku/treat';
```
askoufis, line 2: I don't think ts-ignore is needed anymore. If it was copy-pasted from another fixture, it's probably not needed there anymore either.

packages/sku/config/webpack/webpack.config.js
```diff
@@ -99,7 +100,7 @@ const makeWebpackConfig = ({
     // Libraries should always have the same file name
     // for the main chunk unless we're building for storybook
     if (isLibrary && isMainChunk) {
-      return libraryName;
+      return libraryFile ?? libraryName;
```
askoufis, line 103: Could probably remove/reword the comment here as it's not longer true.

.changeset/many-toys-search.md
```diff
@@ -0,0 +1,26 @@
+---
+'sku': minor
+---
+
+Add `libraryFile` configuration option
+
+This option allows the file name of the library to be specified in the sku
+configuration. If this option is not specified then the `libraryName` option
+will be used for this purpose instead (note that this is the previously
+existing behaviour).
```
askoufis, line 10: Putting each sentence on its own line here results in better release note formatting IMO.

```suggestion
This option allows the file name of the library to be specified in the sku configuration.
If this option is not specified then the `libraryName` option will be used for this purpose instead (note that this is the previously existing behaviour).
```

fixtures/library-file/src/styles.treat.ts
kjots: Yes, the `library-file` fixture was copied from `library-build` with minimal changes.

.changeset/many-toys-search.md
```diff
@@ -0,0 +1,24 @@
...
+import type { SkuConfig } from 'sku';
+
+const skuConfig: SkuConfig = {
+  libraryEntry: 'src/library.js',
+  renderEntry: 'src/render.js',
+  libraryName: 'MyAwesomeLibrary',
+  libraryFile: 'my-awesome-library',
+};
+
+export default skuConfig;
```
mrm007, line 23: Do we want to start encouraging people to use the [(new-ish) `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html)?

```suggestion
export default {
  libraryEntry: 'src/library.js',
  renderEntry: 'src/render.js',
  libraryName: 'MyAwesomeLibrary',
  libraryFile: 'my-awesome-library',
} satisfies SkuConfig;
```

docs/docs/libraries.md
```diff
@@ -12,6 +12,18 @@ modules.exports = {
 };
 ```
 
+By default the file name of the library will be based on the `libraryName` option. A distinct library file name may be specified by providing a `libraryFile` option:
+
+```js
+modules.exports = {
```
mrm007, line 18:
```suggestion
module.exports = {
```

fixtures/library-file/src/library.js
```diff
@@ -0,0 +1,12 @@
+import styles from './styles.less';
+import { customBox } from './styles.treat';
```
mrm007, line 2: We should probably stop adding fixtures with LESS and treat, to make it easier to migrate the future.

packages/sku/entry/libraryRender/index.js
```diff
@@ -1,14 +1,17 @@
 import render from '__sku_alias__renderEntry';
 
 const libraryName = SKU_LIBRARY_NAME;
+const libraryFile = SKU_LIBRARY_FILE;
 const publicPath = __SKU_PUBLIC_PATH__;
```
mrm007, line 5: Kinda weird that we have both `SKU_BAR` and `__SKU_BAR__` names. What do you think about normalising to `__SKU_BAR__`?

.changeset/many-toys-search.md
kjots: Agreed to look into this at a later date.

fixtures/library-file/src/library.js
kjots: Agreed to keep this for now and look into updating the fixtures to be more concise/fit for purpose across the board at a later date.

packages/sku/config/webpack/webpack.config.js
kjots: In this case having the same file name referrers to using a name from the config rather than allowing WebPack to generate it, potentially with a content hash, so this comment is still relevant.

askoufis (approved):

mrm007 (approved):

PR #763, opened 2023-04-11
test: Use explicit ipv4 address instead of `localhost` when waiting on URLs

askoufis:
All our fixture tests use `localhost` in URLs. Some tests were failing locally for me because of this. This has to do with node 17+ preferring ipv6 over ipv4. See https://github.com/jeffbski/wait-on/issues/133 for more info. We're pinning node 16 in sku though, so it's odd that this is even an issue.

Ideally we'd make webpack dev server listen on both ipv4 localhost and ipv6 localhost, but I'm not sure how to configure that. Others are welcome to try. https://webpack.js.org/configuration/dev-server/#devserverhost

### DIFF
--- a/test-utils/waitForUrls.js
+++ b/test-utils/waitForUrls.js
@@ -6,7 +6,13 @@ const waitForUrls = async (...urls) => {
 
   try {
     return await waitOnAsync({
-      resources: urls.map((url) => url.replace(/http(s?)\:/, 'http$1-get:')),
+      resources: urls.map((url) =>
+        url
+          .replace(/http(s?)\:/, 'http$1-get:')
+          // As of node 17, ipv6 is preferred, so explicitly use ipv4
+          // See https://github.com/jeffbski/wait-on/issues/133
+          .replace(/localhost/, '127.0.0.1'),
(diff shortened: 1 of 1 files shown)


michaeltaranto (approved):

test-utils/waitForUrls.js
```diff
@@ -6,7 +6,13 @@ const waitForUrls = async (...urls) => {
 
   try {
     return await waitOnAsync({
-      resources: urls.map((url) => url.replace(/http(s?)\:/, 'http$1-get:')),
+      resources: urls.map((url) =>
+        url
+          .replace(/http(s?)\:/, 'http$1-get:')
+          // As of node 17, ipv6 is preferred, so explicitly use ipv4
+          // See https://github.com/jeffbski/wait-on/issues/133
+          .replace(/localhost/, '127.0.0.1'),
```
mrm007, line 14: Does it work with this?

```suggestion
          .replace(/localhost/, '0.0.0.0'),
```

askoufis: Addressed in #764

PR #770, opened 2023-04-14
Remove some usage of `promisify`, replace `fs-extra` usage with plain `fs`

askoufis:
With recent, and not so recent, dependency updates (e.g. #768), some packages now export async versions of their functions, removing the need to use `promisify` in some situations.

Also, some `fs-extra` usage can be replaced with plain `fs` or `fs/promises`. ~There are some `fs-extra` functions that don't have equivalents in plain `fs`, particularly `ensureExists`, but they could be recreated if need be, and we could potentially remove our dependency on `fs-extra`, if there's any value in doing that.~

After addressing [the comment below](https://github.com/seek-oss/sku/pull/770#discussion_r1166241956), `fs-extra` has been removed as a dependency 🎉 .

### DIFF
--- a/packages/sku/lib/certificate.js
+++ b/packages/sku/lib/certificate.js
@@ -75,7 +75,7 @@ const generateCertificate = async (certificatePath, certificateDirPath) => {
   try {
     await access(certificateDirPath);
   } catch {
-    await mkdir(certificateDirPath, { force: true, recursive: true });
+    await mkdir(certificateDirPath, { recursive: true });
   }
 
   await writeFile(certificatePath, `${pems.private}${pems.cert}`);
--- a/packages/sku/lib/certificate.js
+++ b/packages/sku/lib/certificate.js
@@ -1,7 +1,7 @@
 const selfsigned = require('selfsigned');
 const { blue } = require('chalk');
 const {
-  access: exists,
+  access,
   mkdir,
   unlink,
   writeFile,
@@ -70,10 +70,10 @@ const createSelfSignedCertificate = () => {
 const generateCertificate = async (certificatePath, certificateDirPath) => {
   const startTime = performance.now();
 
-  const pems = await createSelfSignedCertificate();
+  const pems = createSelfSignedCertificate();
 
   try {
-    await exists(certificateDirPath);
+    await access(certificateDirPath);
   } catch {
     await mkdir(certificateDirPath, { force: true, recursive: true });
   }
@@ -92,7 +92,7 @@ const getCertificate = async (certificateDirName = '.ssl') => {
   );
 
   try {
-    await exists(certificatePath);
+    await access(certificatePath);
   } catch {
     return generateCertificate(certificatePath, certificateDirPath);
   }
--- a/packages/sku/lib/certificate.js
+++ b/packages/sku/lib/certificate.js
@@ -72,10 +72,10 @@ const generateCertificate = async (certificatePath, certificateDirPath) => {
 
   const pems = await createSelfSignedCertificate();
 
-  const certificateDirExists = await exists(certificateDirPath);
-
-  if (!certificateDirExists) {
-    await mkdir(certificateDirPath);
+  try {
+    await exists(certificateDirPath);
+  } catch {
+    await mkdir(certificateDirPath, { force: true, recursive: true });
   }
 
   await writeFile(certificatePath, `${pems.private}${pems.cert}`);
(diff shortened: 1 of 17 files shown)


packages/sku/scripts/init.js
```diff
@@ -193,7 +193,7 @@ const getTemplateFileDestinationFromRoot =
       const destination = getTemplateFileDestination(file);
 
       // Ensure folders exist before writing files to them
-      await fs.mkdirp(path.dirname(destination));
+      await fs.ensureDir(path.dirname(destination));
```
jahredhope, line 196: does `ensureDir` offer any value over native `mkdir`?

Since [Node v12](https://nodejs.org/api/fs.html#fspromisesmkdirpath-options) (possibly earlier) mkdir with recursive true will create any dir if it doesn't exist, and won't error if it does exist.

packages/sku/lib/certificate.js
```diff
@@ -1,13 +1,13 @@
 const selfsigned = require('selfsigned');
 const { blue } = require('chalk');
 const {
-  existsSync,
+  access: exists,
```
kjots, line 4: I don't think we should rename the `access` import to `exists` since the two operations have different ways of reporting success or failure (i.e. throw an exception vs return a boolean).

If we wanted to keep using the term `exists` in the code we could create a new function to wrap `access`, i.e.:

```typescript
const exists = async (path: string) => {
  try {
    await access(path);

    return true;
  } catch {
    return false;
  }
};
```

Alternatively, we should just use `access` as-is.

askoufis: That's a good point. I'm happy to remove the rename and just use `access` as-is.

packages/sku/scripts/init.js
askoufis: Swapped to using `mkdir`, and consequently removed the `fs-extra` dep from sku entirely 🎉

packages/sku/lib/certificate.js
```diff
@@ -70,12 +70,12 @@ const createSelfSignedCertificate = () => {
...
+  const pems = createSelfSignedCertificate();
 
-  const certificateDirExists = existsSync(certificateDirPath);
-
-  if (!certificateDirExists) {
-    await mkdir(certificateDirPath);
+  try {
+    await access(certificateDirPath);
+  } catch {
+    await mkdir(certificateDirPath, { force: true, recursive: true });
```
jahredhope: Jedi Mind Tricks don't work on me.

```suggestion
    await mkdir(certificateDirPath, { recursive: true });
```

packages/sku/scripts/init.js
```diff
@@ -148,7 +148,7 @@ const getTemplateFileDestinationFromRoot =
   };
   const packageJsonString = JSON.stringify(packageJson, null, 2);
 
-  fs.writeFileSync(path.join(root, 'package.json'), packageJsonString);
+  fs.writeFile(path.join(root, 'package.json'), packageJsonString);
```
jahredhope, line 151: Should this be awaited?

```suggestion
await  fs.writeFile(path.join(root, 'package.json'), packageJsonString);
```

mrm007, line 151:
```suggestion
  await fs.writeFile(path.join(root, 'package.json'), packageJsonString);
```

packages/sku/scripts/serve.js
```diff
@@ -33,9 +33,9 @@ const preferredSite = args.site;
 (async () => {
   track.count('serve');
 
-  const targetFolderExists = fs.existsSync(paths.target);
-
-  if (!targetFolderExists) {
+  try {
+    await access(paths.target);
+  } catch {
```
mrm007, line 38: Might be useful to have an `exists` util that returns `true`/`false`.

askoufis: Yeah actually having to try/catch everywhere is kinda gross.

askoufis: https://github.com/seek-oss/sku/pull/770/commits/564b44e98abbb575d720132cb8d745f1e0e5f414

packages/sku/scripts/init.js
askoufis: https://github.com/seek-oss/sku/pull/770/commits/1f137e4372bcab5dc0546fa6da91272f496da813

mrm007 (approved):

PR #771, opened 2023-04-16
Add missing fixture dev dep, small code removal

askoufis:
- Added a missing `@vanilla-extract/css` dev dep to the `braid-design-system` fixture. It needs the dep because it has a `css.ts` file.
- Remove duplicate `relativeTarget` entry
- Removed a CI check in the storybook webpack config that was used to remove some noise in CI. There is now a `--quiet` flag, as mentioned in [the issue in the comment](https://github.com/storybookjs/storybook/issues/1260#issuecomment-641866661), which suppress  suppresseses this output, and we already set it for the [`storybook`](https://github.com/seek-oss/sku/blob/1acc7c960d9a3383b0fde071ac0f7ed04a6cb5ef/packages/sku/scripts/storybook.js#L11) and [`storybook-build`](https://github.com/seek-oss/sku/blob/1acc7c960d9a3383b0fde071ac0f7ed04a6cb5ef/packages/sku/scripts/build-storybook.js#L19) commands, so this code can be removed.

### DIFF
--- a/fixtures/braid-design-system/package.json
+++ b/fixtures/braid-design-system/package.json
@@ -7,5 +7,8 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "sku": "workspace:*"
+  },
+  "devDependencies": {
+    "@vanilla-extract/css": "^1.0.0"
   }
 }
--- a/packages/sku/config/storybook/storybookWebpackConfig.js
+++ b/packages/sku/config/storybook/storybookWebpackConfig.js
@@ -1,22 +1,12 @@
-const webpack = require('webpack');
 const { paths } = require('../../context');
 const find = require('lodash/find');
 const { merge: webpackMerge } = require('webpack-merge');
-const isCI = require('../../lib/isCI');
 const makeWebpackConfig = require('../webpack/webpack.config');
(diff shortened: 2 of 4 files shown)


fixtures/braid-design-system/package.json
```diff
@@ -7,5 +7,8 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "sku": "workspace:*"
+  },
+  "devDependencies": {
+    "@vanilla-extract/css": "^1.0.0"
```
mrm007, line 12: I'm confused about this one being a devDep vs dep. `sku` is in `dependencies` but `@vanilla-extract/css` is in `devDependencies`?! Shouldn't all of them be in `dependencies`?

mrm007 (approved):

fixtures/braid-design-system/package.json
askoufis: Good point. Haven't been super rigorous about the distinction between dev dep and dep in fixtures. VE is usually a dev dep since it's not used at runtime, but you're right that `sku` should be a dev dep. I'll fix that up in all the fixtures.

PR #773, opened 2023-04-18
storybook: Re-export `Meta` and `StoryObj` types from `@storybook/react`

askoufis:
The story story feature in storybook is now [enabled by default in storybook v7](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#storystorev7-enabled-by-default). This feature is not compatible with the `storiesOf` API, so in order to enable sku consumers to move off the `storiesOf` API, we should re-export types that can be used to write [CSF stories](https://storybook.js.org/docs/react/api/csf).

### DIFF
--- a/packages/sku/@storybook/react/index.ts
+++ b/packages/sku/@storybook/react/index.ts
@@ -11,6 +11,8 @@ import {
   storiesOf,
   forceReRender,
   getStorybook,
+  type Meta,
+  type StoryObj,
 } from '@storybook/react';
 
 export {
@@ -21,4 +23,6 @@ export {
   storiesOf,
   forceReRender,
   getStorybook,
+  type Meta,
+  type StoryObj,
 };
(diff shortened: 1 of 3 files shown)


packages/sku/@storybook/react/index.ts
```diff
@@ -11,6 +11,8 @@ import {
   storiesOf,
   forceReRender,
   getStorybook,
+  type Meta,
+  type StoryObj,
 } from '@storybook/react';
 
 export {
```
mrm007, line 18: Can we `export * from` and `export type * from`? 

> We can't re-export directly because eslint-plugin-import doesn't understand

I think now it does.

mrm007 (approved):

packages/sku/@storybook/react/index.ts
askoufis: Chatted about this. I'm a bit hesitant to just blank re-export if we don't need to, so will leave as-is for now.

PR #774, opened 2023-04-18
fix: Use `fs.copyFile` instead of `fs.copySync`, make some file utils async

askoufis:
`fs.copySync` doesn't exist 🤦 , so replaced it with `fs.copyFile`. Also made the `copyPublicFiles` and `ensureTargetDirectory` utilities async as they mostly already had `await`s on them at their call sites.


mrm007 (approved):
Good catch 🕵️

PR #776, opened 2023-04-18
Fix copying of `public` folder to `target` folder during post-build

askoufis:
Turns out not having typescript sucks. Who would've thought?

I don't think [the `dereference` flag](https://github.com/jprichardson/node-fs-extra/blob/master/docs/copy-sync.md#copysyncsrc-dest-options) is that important when copying files from `public` to `target`, but LMK if I'm wrong. Also, it was available on `copySync` from `fs-extra`, but it's only available on `fs/promises` via the `cp` and `cpSync` functions, both of which are experimental, hence why I removed the flag rather than migrating to one of those functions.

I opted to create a custom copy directory contents function that recursively copies the contents of one directory to another. Not having to depend on `fs-extra` for this single function is a fair tradeoff to having some code that will likely stick around until the aforementioned APIs become stable.

I'm very surprised this wasn't caught during tests, as we run `build` in a few fixture tests. I've been testing these in `cm-ui`, so I'll run a snapshot first before merging this fix to ensure it _actually_ works.

EDIT: Turns out we weren't actually testing the [`public` sku config option](https://seek-oss.github.io/sku/#/./docs/configuration?id=public) in any tests, so I tweaked the config of the `ssr-hello-world` fixture to trigger asset copying behaviour that triggers the bug, so now we have a test for it.


mrm007 (approved):

mrm007:
Keep fighting the good fight ✊

jahredhope (approved):

PR #787, opened 2023-05-03
Start dev server middleware alongside Storybook

askoufis:
Doing some QOL storybook uplift before the storybook 7 release.

Storybook will inject middleware found in `.storybook/middleware.js` into its own middleware stack. This is an [intentionally undocumented feature](https://github.com/storybookjs/storybook/issues/15300#issuecomment-866469521), but given that it has been stable for [at least 5 years](https://github.com/storybookjs/storybook/blame/e22ac0d0e8f5cec6cd8e10ef34aa66f758335fbf/code/lib/core-server/src/utils/middleware.ts#L17), I don't think it will disappear any time soon, so I'm happy to depend on it.

Adding this feature neatly solves the issue where the new `seekJobs` braid theme needs a webfont. Shared web assets exposes a middleware for serving the webfont locally, and with this feature the middleware will also be available when running storybook.

This feature has also been requested by teams that would like to run other middleware alongside storybook.

I also did a small docs pass to add some subheadings to the storybook docs, as well as document this new feature. Will do another docs pass for storybook 7.

Finally, I moved the `node-fetch` dev dep from sku to test utils, where it belongs.

### DIFF
--- a/.changeset/mighty-ghosts-enjoy.md
+++ b/.changeset/mighty-ghosts-enjoy.md
@@ -1,0 +1,7 @@
+---
+'sku': minor
+---
+
+Start sku dev server middleware when running storybook
+
+When running `sku storybook`, if you have configured `devServerMiddleware` in your sku config, that middleware will now be passed through to storybook and injected into its own middleware stack.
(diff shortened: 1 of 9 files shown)


michaeltaranto (approved):
👏

PR #790, opened 2023-05-05
Update fixtures to react 18

askoufis:
While sku supports apps that use react 18, and the init template uses react 18 APIs, up until now we weren't using react 18 internally. The main drivers for doing this are:
- Using braid v32 in fixtures (without the webpack hack) so we can ensure sku doesn't break braid and vice-versa
- Just generally having up-to-date dependencies

Migration was basically a standard react 18 migration. The only hitch was a having to add a `TextEncoder`/`TextDecoder` shim in a few fixtures because [jsdom doesn't include it](https://github.com/jsdom/jsdom/issues/2524) but we need it for a few tests that call `renderToString`. We had to do the same thing [in Braid](https://github.com/seek-oss/braid-design-system/blob/master/testConfig/setupTests.ts#L1-L7).

As usual, most of the changes in this PR are snapshot diffs.


fixtures/assertion-removal/package.json
```diff
@@ -2,12 +2,13 @@
...
   "dependencies": {
-    "react": "^17.0.1",
-    "react-dom": "^17.0.1"
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
   },
   "devDependencies": {
-    "@types/react": "^17.0.0",
-    "@types/react-dom": "^17.0.0",
+    "@types/node": "^16.0.0",
```
askoufis, line 9: This was added so that imports from `assert` are typed.

mrm007 (approved):

PR #791, opened 2023-05-05
Update nvmrc to node 16

askoufis:
I _think_ renovate might use `.nvmrc` when doing its updates, and pnpm 8 dropped support for node 14 so renovate PRs are bailing halfway through installation so they end up with no lockfiles. [Example](https://github.com/seek-oss/sku/pull/743#issuecomment-1532486092).


mrm007:
Can we get rid of it altogether? Or make it sync with `package.json` (in a postinstall step)?

askoufis:
> Can we get rid of it altogether? Or make it sync with `package.json` (in a postinstall step)?



> Can we get rid of it altogether? Or make it sync with `package.json` (in a postinstall step)?

Good idea. https://github.com/seek-oss/sku/pull/791/commits/43404f27120cdc9fe8454993472c96152bd779ec

If it doesn't fix the renovate issue I'll just remove it.

package.json
```diff
@@ -15,7 +15,8 @@
     "format-check": "prettier --list-different '**/*.{js,ts,tsx,md,less,css}'",
     "deploy-docs": "pnpm run --filter @sku-private/docs deploy",
     "release": "pnpm deploy-docs && changeset publish",
-    "version": "changeset version && pnpm install --lockfile-only"
+    "version": "changeset version && pnpm install --lockfile-only",
+    "postinstall": "node ./postinstall.js"
```
mrm007, line 19: Let's move this before `prepare` since they're both [life cycle scripts](https://docs.npmjs.com/cli/v8/using-npm/scripts?v=true#life-cycle-operation-order).

mrm007 (approved):

PR #629, opened 2021-08-16
Release sku 11

mattcompiles:

PR #792, opened 2023-05-10
Support storybook preview file

askoufis:
With the deprecation (and future removal) of the `storiesOf` API, the best place for global configuration of stories is the `preview.js` file. See [the storybook docs](https://storybook.js.org/docs/react/configure/overview#configure-story-rendering) for more info about what this file can be used for.

This PR adds support for storybook's `preview.(js|ts|tsx)` file. When running `sku storybook` or `sku build-storybook`, if a preview file is detected in the `.storybook` folder relative to the sku config folder, a symlink will be created in sku's storybook config directory that points to the preview file.

I also removed the decorator that sku was injecting into every story. This was required in order to render text correctly, back when we were still using basekick. It was kept [when we moved to capsize](https://github.com/seek-oss/sku/pull/629/files#diff-1e4a7b2dd7785c27a7153b5e0b48bbcdb51c42c563cc3c118c59406c4a4a0828L9), but isn't actually needed. Not marking this as a breaking change as it won't affect any projects other than Braid, and even then it will only result in some screenshot diffs with a few border pixels missing.

I've also optimized the storybook test utilities so we can re-use the same iframe that puppeteer gives us between tests. We test the same story in each test anyway. Previously, each storybook test was opening the page, clicking a button, getting the story iframe, and then running a selector on the iframe. This resulted in each test taking about 1-2s (on my machine). Now that we're re-using the same iframe between tests, each test takes a few milliseconds.

EDIT: Looks like a similar test speedup on CI too 🏎️


askoufis:
So, because we're now making/cleaning up symlinks when starting/building storybook, when running multiple storybooks at once, which happens when running sku's tests, there's a possibility that a symlink could be created by one instance but cleaned up by another, [so storybook throws an error because it can't resolve the file](https://github.com/seek-oss/sku/actions/runs/4942724877/jobs/8836515701#step:11:371).

I think the best solution to this would be to move all storybook tests to a single fixture, probably `storybook-config`, so we only ever run a single instance of storybook at once (hopefully).

askoufis:
Blocked pending #793.

packages/sku/config/storybook/decorators.js
michaeltaranto, line 1: 👋

packages/sku/package.json
```diff
@@ -143,6 +143,7 @@
     "@vocab/react": "^1.0.1",
     "assert": "^2.0.0",
     "braid-design-system": "^31.0.0",
+    "memfs": "^3.5.1",
```
michaeltaranto, line 146: I might have missed it, but where is this being used?

askoufis: Oh woops, I was playing around with it but didn't actually end up using it. Will remove.

askoufis: https://github.com/seek-oss/sku/pull/792/commits/c2a2779ab8ed647acab21873236dca67281e9cf1

tests/storybook-config.test.js
```diff
@@ -59,110 +41,142 @@ describe('storybook-config', () => {
...
     });
 
+    it('should render decorators defined in the storybook preview file', async () => {
+      const { text, fontSize } = await getTextContentFromStorybookFrame(
+        storybookFrame,
+        '[data-automation-decorator]',
+      );
+
+      expect(text).toEqual('Braid Text decorator');
+      expect(fontSize).toEqual('16px');
```
michaeltaranto, line 51: one day.... maybe one day.... but should be fine for the foreseeable future.

michaeltaranto (approved):

.changeset/chilled-planes-report.md
```diff
@@ -0,0 +1,22 @@
...
+
+Re-export the `DecoratorFn` type from `@storybook/react`
+
+**EXAMPLE USAGE**:
+
+```tsx
+// preview.tsx
+import React from 'react';
+
+import { type DecoratorFn } from 'sku/@storybook/react';
```
mrm007, line 13: For such cases, I recommend the `import type` syntax.

```suggestion
import type { DecoratorFn } from 'sku/@storybook/react';
```

.changeset/large-needles-serve.md
```diff
@@ -0,0 +1,31 @@
...
+```js
+import 'braid-design-system/reset';
+
+import apac from 'braid-design-system/themes/apac';
+import { BraidProvider } from 'braid-design-system';
+
+import React from 'react';
+
+// This will wrap every story in a BraidProvider
+export const decorators = [
```
mrm007, line 20: Having two ways of declaring decorators in the same changelog entry might be confusing. I'd suggest merging the two changesets. The other one doesn't stand on its own.

docs/docs/storybook.md
```diff
@@ -19,6 +19,32 @@ export const Secondary = () => <Button variant="secondary">Secondary</Button>;
...
+```js
+import 'braid-design-system/reset';
+
+import apac from 'braid-design-system/themes/apac';
+import { BraidProvider } from 'braid-design-system';
+
+import React from 'react';
+
+// This will wrap every story in a BraidProvider
+export const decorators = [
```
mrm007, line 35: We should suggest using the type `DecoratorFn`.

packages/sku/lib/storybook.js
```diff
@@ -0,0 +1,70 @@
...
+    storybookConfigDirectory,
+    previewFileName,
+  );
+  const symlinkFiles = await glob(symlinkFileAsbolutePath, {
+    dot: true,
+    onlyFiles: false, // Required to find dangling symlinks
+  });
+
+  // Clean up any existing files in the storybook config folder so we have a clean slate
+  for (const symlinkFile of symlinkFiles) {
```
mrm007, line 38: So if >1 files are found, this would remove all symlinks but not create any symlinks in place. Is that intentional?

packages/sku/lib/storybook.js
```diff
@@ -0,0 +1,70 @@
...
+    log(`Cleaning up exising file ${symlinkFile}.`);
+  }
+
+  const previewFiles = await glob(previewFileAbsolutePath);
+
+  if (previewFiles.length > 1) {
+    console.log(
+      'Multiple storybook preview files found. Please only define a single preview file.',
+    );
+    console.log(previewFiles);
```
mrm007, line 49: Should we make these `console.error`s?

packages/sku/lib/storybook.js
```diff
@@ -0,0 +1,70 @@
...
+  path.dirname(paths.appSkuConfigPath),
+  '.storybook',
+  previewFileName,
+);
+
+/**
+ * This function first cleans up any existing preview.{js,ts,tsx} files in the provided storybook config directory.
+ * This includes any potentially dangling symlinks.
+ * Then it looks for a `.storybook/preview.{js,ts,tsx}` file relative to the sku config folder.
+ * If 1 file is found, a symlink is created to it from the provided storybook config directory.
```
mrm007, line 22: I find **from**s and **to**s confusing in the context of symlinks.
```suggestion
 * If 1 file is found, a symlink pointing to it is created in the provided storybook config directory.
```

packages/sku/lib/storybook.js
```diff
@@ -0,0 +1,70 @@
...
+const previewFileAbsolutePath = path.join(
+  path.dirname(paths.appSkuConfigPath),
+  '.storybook',
+  previewFileName,
+);
+
+/**
+ * This function first cleans up any existing preview.{js,ts,tsx} files in the provided storybook config directory.
+ * This includes any potentially dangling symlinks.
+ * Then it looks for a `.storybook/preview.{js,ts,tsx}` file relative to the sku config folder.
```
mrm007, line 21:
```suggestion
 * Then it looks for a `.storybook/preview.{js,ts,tsx}` file relative to the sku config.
```

packages/sku/lib/storybook.js
```diff
@@ -0,0 +1,70 @@
...
+ * If 1 file is found, a symlink is created to it from the provided storybook config directory.
+ * Does nothing if 0 or >1 files are found.
+ *
+ * @param {string} storybookConfigDirectory The path to the storybook config directory
```
mrm007, line 25: I find your lack of respect for the Storybook brand name disturbing.

mrm007 (approved):
Nicely done 🍒 🐎

packages/sku/lib/storybook.js
askoufis: sToRyBoOk

packages/sku/lib/storybook.js
askoufis: Yes. Since it's globbing for multiple file extensions, it could potentially return multiple files. However, if you change from `preview.js` to `preview.tsx` for example, the next time you run `sku storybook`, it would remove the old symlink and then make the new one, so I don't even think it's possible to end up with multiple symlinks unless you manually add one yourself.

mrm007: It can happen you have `preview.js` and `preview.tsx` because of a bad merge. Very edgy case.

.changeset/large-needles-serve.md
askoufis: https://github.com/seek-oss/sku/pull/792/commits/2c92ac8447f0f69efeb05da0f7fb087bf6f9975b

PR #793, opened 2023-05-11
Move all storybook tests to a single fixture

askoufis:
Due to a feature being added in #792, we need to ensure that only a single fixture runs storybook on CI, specifically when running tests. This PR moves all storybook tests to a single fixture, the `storybook-config` fixture, so that no other fixtures end up starting their own storybook.


fixtures/storybook-config/src/StorybookDecorator.tsx
```diff
@@ -2,9 +2,15 @@ import 'braid-design-system/reset';
 import { BraidProvider } from 'braid-design-system';
 import apac from 'braid-design-system/themes/apac';
 import React from 'react';
+// @ts-expect-error
```
mrm007, line 5: What's the error here? No types?

fixtures/storybook-config/src/TestComponent.tsx
```diff
@@ -1,15 +1,30 @@
-import { Card, Text } from 'braid-design-system';
+import { Box, Card, Text } from 'braid-design-system';
 import React from 'react';
+// @ts-expect-error
+import { Text as SSGText, HeartIcon } from 'seek-style-guide/react';
```
mrm007, line 4: Thoughts on `import * as ssg`? Then you could write `<ssg.HeartIcon />`, `<ssg.Text />` etc.

tests/storybook-config.test.js
```diff
@@ -12,6 +12,31 @@ const appDir = path.dirname(
...
+  const firstStoryButton = await page.waitForSelector(
+    '#storybook-explorer-menu button',
+  );
+
+  // Ensure default story is activated
+  firstStoryButton.click();
+
+  const iframeElement = await page.waitForSelector('#storybook-preview-iframe');
+
+  const storybookFrame = await iframeElement.contentFrame();
```
mrm007, line 28: Oh yeah, this brings back [memories](https://www.cypress.io/blog/2020/02/12/working-with-iframes-in-cypress/)

ptsd dog

tests/storybook-config.test.js
```diff
@@ -12,6 +12,31 @@ const appDir = path.dirname(
 );
 const storybookDistDir = path.resolve(appDir, 'dist-storybook');
 
+const getStorybookElement = async (url, elementSelector) => {
```
mrm007, line 15:
```suggestion
const getElementWithinStorybookFrame = async (storybookUrl, elementSelector) => {
```

to make it more clear what this function does

mrm007 (approved):

fixtures/storybook-config/src/StorybookDecorator.tsx
askoufis: Yep. Seek style guide only ever had flow types. Could be a `@ts-ignore` but it doesn't really matter in this case.

mrm007: In that case, might be worth leaving a comment

```suggestion
// @ts-expect-error no types
```

PR #794, opened 2023-05-12
Don't run postinstall script if a project isn't using sku

mrm007:
When installing dependencies for a brand new Next.js project with Braid and Vanilla Extract, sku will run its postinstall script even though it's not used in the project.

Braid has a peer dependency on sku and npm/pnpm 8 will auto install peer deps, so the postinstall script will be run.


askoufis:
The configuration and init fixtures are failing. I have a feeling that we could probably remove the `skuSkipConfigure` from the top-level `package.json`. I also wonder if we're checking the correct package JSON relative to where `sku` is running/being installed.

askoufis (approved):
CI is failing, looks like some

PR #802, opened 2023-05-18
Add warning banner when `.less` files are detected

askoufis:
While we're still going to support `.less` files for now, we should encourage consumers to move to VE styles, as it is the preferred styling solution. The docs page for styling has been updated with the same deprecation message.

image

### DIFF
--- a/packages/sku/lib/configure.js
+++ b/packages/sku/lib/configure.js
@@ -135,7 +135,7 @@ module.exports = async () => {
       )} will be removed in a future release.`,
       `Consumers are encouraged to migrate to ${chalk.bold(
         'Vanilla Extract',
-      )} at their earliest opportunity.`,
+      )} at the earliest opportunity.`,
       'https://seek-oss.github.io/sku/#/./docs/styling?id=vanilla-extract',
     ]);
   }
--- a/packages/sku/lib/configure.js
+++ b/packages/sku/lib/configure.js
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 const { writeFile, rm } = require('fs/promises');
 const path = require('path');
+const glob = require('fast-glob');
+const fs = require('fs');
 
 const ensureGitignore = require('ensure-gitignore');
 const { cwd, getPathFromCwd } = require('./cwd');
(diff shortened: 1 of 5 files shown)


packages/sku/lib/configure.js
```diff
@@ -109,4 +113,24 @@ module.exports = async () => {
     comment: 'managed by sku',
     patterns: gitIgnorePatterns.map(convertToForwardSlashPaths),
   });
+
+  // Check for `.less` files
+  const lessFileGlobResults = await Promise.all(
+    paths.src
+      .filter((srcPath) => fs.statSync(srcPath).isDirectory())
```
askoufis, line 120: For whatever reason, async `fs.stat` throws an exception on certain files, like `sku.config.ts`. I don't know why.

packages/sku/lib/configure.js
```diff
@@ -109,4 +113,24 @@ module.exports = async () => {
...
+  const srcHasLessFiles = lessFileGlobResults.some((fileArray) => {
+    return fileArray.length > 0;
+  });
+  if (srcHasLessFiles) {
+    printBanner('warning', 'Less styles detected', [
+      `Vanilla Extract is the preferred styling solution supported by sku. Support for ${chalk.bold(
+        '.less',
+      )} styles may be removed in the future. Please migrate all ${chalk.bold(
+        '.less',
+      )} styles to Vanilla Extract styles.`,
```
michaeltaranto, line 132: Thoughts on being a little firmer, but also leading with the deprecation call out?

> Support for LESS has been deprecated.
>
> Vanilla Extract is the preferred styling solution supported by sku, and support for LESS will be removed in a future release.
>
>  Consumers are encouraged to migrate to Vanilla Extract at their earliest opportunity.

jahredhope (approved):

packages/sku/lib/configure.js
askoufis: https://github.com/seek-oss/sku/pull/802/commits/5e5dc36ede97d4397c321f20c435a89fa6a7bec1

michaeltaranto (approved):

PR #807, opened 2023-05-19
Remove stray `console.log`

askoufis:
Leftover from #802


packages/sku/lib/configure.js
```diff
@@ -124,10 +124,7 @@ module.exports = async ({ isPostInit } = {}) => {
     const lessFileGlobResults = await Promise.all(
       paths.src
         .filter((srcPath) => fs.statSync(srcPath).isDirectory())
```
mrm007, line 126: This comment would be useful as a code comment https://github.com/seek-oss/sku/pull/802#discussion_r1198393128

mrm007 (approved):

jahredhope (approved):

PR #809, opened 2023-05-19
Remove treat support

askoufis:


packages/sku/config/eslint/importOrderConfig.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
         ],
         pathGroups: [
           {
-            pattern: '*.+(treat|less|css)',
+            pattern: '*.+(less|css)',
```
mrm007, line 32:
```suggestion
            pattern: '*.@(less|css)',
```

https://github.com/micromatch/picomatch#extglobs

packages/sku/config/webpack/plugins/sku-webpack-plugin/index.js
```diff
@@ -65,7 +63,7 @@ class SkuWebpackPlugin {
         include: this.include,
         oneOf: [
           {
-            compiler: /(vanilla-extract|treat-webpack-loader)/,
+            compiler: /(vanilla-extract)/,
```
mrm007, line 66:
```suggestion
            compiler: /\bvanilla-extract\b/,
```

packages/sku/lib/validatePeerDeps.js
```diff
@@ -17,7 +17,7 @@ const asyncMap = (list, fn) => {
   return Promise.all(list.map((item) => fn(item)));
 };
 
-const singletonPackages = ['react-treat'];
+const singletonPackages = ['@vanilla-extract/css'];
```
mrm007, line 20: 👍

packages/sku/scripts/postinstall.js
```diff
@@ -30,7 +30,7 @@ if (packageJsonExists) {
   const skipPostInstall = skuSkipPostInstall || skuSkipPostinstall;
   const hasSku =
     Boolean(devDependencies?.sku) ||
-    // TODO: get rid of this part when we remove treat
+    // sku should always be a dev dependency, but some repos may still have it as a regular dependency
```
mrm007, line 33: Should we warn them? It would somewhat reduce the size of their Docker image 🌍🔥

mrm007 (approved):
Goodbye 🍬

packages/sku/scripts/postinstall.js
askoufis: That's a good idea. More warning banners 🎉

mrm007: jokes

packages/sku/config/webpack/plugins/sku-webpack-plugin/index.js
askoufis: Do we even need the brackets anymore, since there's only one variant?

mrm007: I replaced the brackets with `\b`s

askoufis: Oh, I though `\b` was brackets, it's word boundary 🤦

packages/sku/scripts/postinstall.js
askoufis: https://github.com/seek-oss/sku/pull/809/commits/64c4d37903975c73d8acdd8c5c58b44bcfc44a41

askoufis:
This is blocked by #743. Braid v31 still references `sku/react-treat`, so we need to get on v32 before we can remove treat.

PR #810, opened 2023-05-23
Upgrade to Storybook v7

askoufis:
This upgrade was _fairly_ straightforward. The main changes required to support v7 were in the storybook config itself. Other noteworthy changes are:

- Added a `storybookStoryStore` flag to sku config, which is then passed through to the storybook config. This is required in order to support the `storiesOf` API. Use of this API is required in Braid and Metropolis, and although the API is deprecated, [the storybook team have no intention to remove it until there is a suitable replacement for dynamically generating stories](https://github.com/storybookjs/storybook/issues/9828#issuecomment-1370291568) (which there currently isn't).
- We now customize storybook's babel config because it now has a pretty barebones one by default. This was necessary to keep support for `preview.tsx` files as they used to work in v6, but with the newer, trimmed down internal babel config they stopped working.
- Change how arguments are passed to `sku storybook`. I'm not sure why this suddenly stopped working, but it was necessary so that we can pass the `--ci` flag to `sku storybook` in order to stop your browser opening when running the storybook tests.


fixtures/storybook-config/src/TestComponent.stories.tsx
```diff
@@ -5,7 +5,7 @@ import type { Meta, StoryObj } from 'sku/@storybook/react';
 
 import { TestComponent } from './TestComponent';
 
-export default {
+const meta = {
   title: 'TestComponent',
   component: TestComponent,
   argTypes: {
```
mrm007, line 11: Verified locally and `argTypes` can be removed.

packages/sku/config/storybook/config.js
mrm007, line 1: Is this ESM or compiled with Babel/Webpack?

packages/sku/config/storybook/config.js
```diff
@@ -1,13 +1,41 @@
...
+  },
+  features: {
+    storyStoryV7: storybookStoryStore,
   },
+  babel: (config) => ({
+    ...config,
+    presets: [
+      ...config.presets,
+      [
+        '@babel/preset-env',
```
mrm007, line 31: Do we need to `require.resolve` these? Would it still work with strict/default pnpm (no `shamefully-hoist`)?

.changeset/slow-numbers-collect.md
```diff
@@ -0,0 +1,42 @@
...
+sku now supports Storybook v7. Please read [the Storybook migration guide] for a high-level overview of what has changed. For a more detailed list of changes, take a look at [the full migration notes].
+**NOTE**: Since sku installs and configures Storybook for you, a lot of the changes will not be relevant to users.
+
+**BREAKING CHANGE**
+As of Storybook v7, stories that use the `storiesOf` API will not work by default. The `storiesOf` API is deprecated and will be removed in Storybook v8, so it is highly encouraged to migrate your stories to the [Component Story Format (CSF)][csf].
+
+Migration can be done automatically via the migration tools provided by Storybook:
+
+```sh
+npx storybook@latest migrate storiesof-to-csf --glob="src/**/*.stories.tsx"
```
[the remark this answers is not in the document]
askoufis: `storybook@latest` is what's in their docs, but I agree we could just use it via `sku`.

fixtures/import-order/.gitignore
```diff
@@ -7,3 +7,5 @@ dist/
 report/
 tsconfig.json
 # end managed by sku
+
+src/
```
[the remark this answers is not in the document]
askoufis: I'd occasionally end up with `src` files from this fixture hanging around because I interrupted the test before the cleanup step. This just ensures that you can't accidentally commit these files, because I definitely did that a few times.

packages/sku/config/storybook/config.js
askoufis: [SB supports ESM for this config file now.](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#esm-format-in-mainjs)

packages/sku/config/storybook/config.js
askoufis: Hmm, not sure honestly, depends how babel resolves them. I suppose it couldn't hurt to `require.resolve` them, just in case.

.changeset/slow-numbers-collect.md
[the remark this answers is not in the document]
askoufis: Turns out this won't work because `sku storybook` passes the `dev` command, so `migrate` does nothing. Will change to `@7` though.

packages/sku/config/storybook/config.js
```diff
@@ -1,13 +1,40 @@
...
+  },
+  features: {
+    storyStoreV7: storybookStoryStore,
   },
+  babel: (config) => ({
+    ...config,
+    presets: [
+      ...config.presets,
+      [
+        require.resolve('@babel/preset-env'),
```
mrm007, line 30: `require` global doesn't exist in ESM.

askoufis: Ah yes. :(

mrm007 (approved):
Worth bumping Storybook to latest before releasing?

PR #813, opened 2023-05-25
Upgrade TypeScript to 5.0, bump babel deps

askoufis:
Always feels weird making a typescript update a minor, but this will be released in a major version anyway so it doesn't really matter.

NOTE: Snapshots are broken until the storybook 7 update is merged in. This is because SB6 doesn't support TS 5.0 because of a removed TypeScript API that's used by an old version of `@storybook/react-docgen-typescript-plugin`.


askoufis:
Should we do `~5.0.0` instead of `^5.0.0` (skuba does this) and get in the habit of bumping the typescript version more regularly? I always feel weird not doing them as major versions though.

kjots (approved):

PR #814, opened 2023-05-25
Upgrade to ESLint v8

askoufis:
Now that [`eslint-config-seek` completely supports ESLint v8](https://github.com/seek-oss/eslint-config-seek/releases/tag/v11.1.2), we can finally upgrade. Tested in metropolis, works great.

NOTE: This change can go out before the upcoming major.


.changeset/fifty-laws-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
```
mrm007, line 2: I would at least make it a `minor`.

My rule of thumb is:
- if a dep upgrade is a major, my bump is a major or minor (it depends™️)
- if a dep upgrade is a minor, my bump is a patch
- if a dep upgrade is a patch, my bump is a patch

packages/sku/lib/runESLint.js
```diff
@@ -1,5 +1,5 @@
 const { yellow, cyan, gray } = require('chalk');
-const EslintCLI = require('eslint').CLIEngine;
+const ESLint = require('eslint').ESLint;
```
mrm007, line 2: 🏅 

```suggestion
const { ESLint } = require('eslint');
```

mrm007 (approved):

.changeset/fifty-laws-hammer.md
askoufis: But that's so arbitrary. A downstream dep could have 0 effect on the API and it'd still be a minor (basically this PR).

mrm007: There are projects with `dangerouslySetEslintConfig` and custom ESLint plugins.

askoufis: Sure, but that's a liability they've taken on themselves. This doesn't really add any features to sku, but it provides access to eslint 8's new features I guess. I guess we're indirectly supporting new things, so maybe it does make sense to be a `minor`. Though I don't agree with picking the upgrade based on the dependency upgrade, that just feels too arbitrary if it has no effect on consumers.

PR #817, opened 2023-05-29
Check that paths exist before looking for `*.less` files in them

mrm007:
This also moves the logic to check for `.less` files into a separate module, so it's easier to remove when we drop support.

### DIFF
--- a/packages/sku/lib/validateLessFiles.js
+++ b/packages/sku/lib/validateLessFiles.js
@@ -1,0 +1,38 @@
+const fs = require('fs');
+const path = require('path');
+const glob = require('fast-glob');
+const chalk = require('chalk');
+
+const printBanner = require('./banner');
+const exists = require('./exists');
+const { paths } = require('../context');
+
+module.exports = async () => {
+  const srcPathsExist = await Promise.all(
+    paths.src.map(async (srcPath) => (await exists(srcPath)) && srcPath),
+  );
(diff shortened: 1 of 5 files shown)


packages/sku/lib/validateLessFiles.js
```diff
@@ -0,0 +1,38 @@
...
+const chalk = require('chalk');
+
+const printBanner = require('./banner');
+const exists = require('./exists');
+const { paths } = require('../context');
+
+module.exports = async () => {
+  const srcPathsExist = await Promise.all(
+    paths.src.map(async (srcPath) => (await exists(srcPath)) && srcPath),
+  );
```
mrm007, line 13: The old code doesn't have this bit.

packages/sku/lib/validateLessFiles.js
```diff
@@ -0,0 +1,38 @@
...
+const glob = require('fast-glob');
+const chalk = require('chalk');
+
+const printBanner = require('./banner');
+const exists = require('./exists');
+const { paths } = require('../context');
+
+module.exports = async () => {
+  const srcPathsExist = await Promise.all(
+    paths.src.map(async (srcPath) => (await exists(srcPath)) && srcPath),
```
askoufis, line 12: Would `filter` work here?
```suggestion
    paths.src.filter(async (srcPath) => await exists(srcPath)),
```

askoufis (approved):

PR #830, opened 2023-06-06
Bump `eslint-config-seek`

askoufis:
This will ensure that consumers get a version that includes [the fix for the custom eslint rule](https://github.com/seek-oss/eslint-config-seek/releases/tag/v11.1.3).


mrm007 (approved):
👏

PR #835, opened 2023-06-09
Dependency updates, bump jest `maxWorkers` to 3 

askoufis:
- Removed [`babel-plugin-dynamic-import-node`](https://www.npmjs.com/package/babel-plugin-dynamic-import-node). This plugin was only being used in jest, i.e. Node, but [dynamic import has been supported in Node since 13.2.0](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#browser_compatibility), so there's no need to transform it anymore.
- Updated `less-loader` version, primarily to prevent consumers getting locked to a version without [this fix that removed an unused, very old dependency](https://github.com/webpack-contrib/less-loader/pull/517) that ended up including various old versions of things, as well as `node-gyp` in `node_modules` 🙅 
- Updated `node-emoji` to [the latest major version](https://github.com/omnidan/node-emoji/releases/tag/v2.1.0). Was waiting for them to release [my change](https://github.com/omnidan/node-emoji/pull/138).
- Set [`maxWorkers`](https://jestjs.io/docs/cli#--maxworkersnumstring) to 3 on CI as [the macos github agents have 3 cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), so we might as well use them. Looks like we can get <5 minute test runs on CI pretty consistently now. This will go away if we eventually move back to ubuntu images, but it's nice to have for now.

### DIFF
--- a/.changeset/funny-geese-drive.md
+++ b/.changeset/funny-geese-drive.md
@@ -2,7 +2,7 @@
 'sku': minor
 ---
 
-Remove `babel-plugin-dynamic-import-node` dependency.
+Remove `babel-plugin-dynamic-import-node` dependency
 
 This plugin was used to transform dynamic imports into deferred requires within jest tests.
-However, dynamic imports are well supported in Node now so this plugin is no longer required.
+However, dynamic imports are well supported in Node, so this plugin is no longer required.
--- a/.changeset/funny-geese-drive.md
+++ b/.changeset/funny-geese-drive.md
@@ -1,0 +1,8 @@
+---
+'sku': minor
+---
+
+Remove `babel-plugin-dynamic-import-node` dependency.
+
+This plugin was used to transform dynamic imports into deferred requires within jest tests.
+However, dynamic imports are well supported in Node now so this plugin is no longer required.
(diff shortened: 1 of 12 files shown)


mrm007 (approved):
Not a huge issue, but I would mark these changes as `minor`. I would say the removal of code falls in the same category as adding code.

> Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. We would usually expect additional code for the latter instance, in which case it’s obviously a minor level increment.

I'm not too fussed about the semantics, though. Feel free to leave it as is if you disagree.

askoufis:
> Not a huge issue, but I would mark these changes as `minor`. I would say the removal of code falls in the same category as adding code.
> 
> > Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. We would usually expect additional code for the latter instance, in which case it’s obviously a minor level increment.
> 
> I'm not too fussed about the semantics, though. Feel free to leave it as is if you disagree.

Split out the changeset into two separate ones: one for dep updates, and the other for the dep removal. Also snuck in a `node-emoji` update too since they finally released a new version.

PR #838, opened 2023-06-19
Surface unhandled ESLint errors

mrm007:
Provide more context when ESLint reports an error. Because this isn't very helpful:

```
Linting
Running Vocab compile
Checking code with TypeScript compiler
Path: /some/path
Checking code with Prettier
Paths: **/*.{ts,cts,mts,tsx,js,cjs,mjs,jsx,md,less,css}
Checking code with ESLint
Paths: .
error Command failed with exit code 1.
```

It doesn't say what the error is.

### DIFF
--- a/packages/sku/lib/runESLint.js
+++ b/packages/sku/lib/runESLint.js
@@ -59,6 +59,8 @@ const runESLint = async ({ fix = false, paths }) => {
       if (e && e.message && e.message.includes('No files matching')) {
         console.warn(yellow(`Warning: ${e.message}`));
       } else {
+        console.warn(yellow('ESLint encountered an error:'));
+        console.log(e.message);
         return Promise.reject();
       }
     }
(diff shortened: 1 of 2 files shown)


packages/sku/lib/runESLint.js
```diff
@@ -59,6 +59,8 @@ const runESLint = async ({ fix = false, paths }) => {
       if (e && e.message && e.message.includes('No files matching')) {
         console.warn(yellow(`Warning: ${e.message}`));
       } else {
+        console.warn(yellow('ESLint encountered an error:'));
+        console.log(e.message);
```
askoufis, line 63: Do we want to `console.error` here instead of just `warn`/`log`?

mrm007: Happy for it to be either, but `console.log` isn't as noisy (red).

askoufis (approved):

PR #839, opened 2023-06-19
test: Don't load storybook on-demand

askoufis:
By default, Storybook 7 loads stories on-demand. This behaviour is a nice default, but it results in the storybook page loading very quickly, which can potentially result in a timeout before the actual story loads. By disabling [the storybook story store](https://storybook.js.org/docs/react/configure/overview#feature-flags), we can force storybook to load all stories upfront before making the storybook page available. This should hopefully make the storybook tests less flakey.


mrm007 (approved):
Nice find 👌

PR #841, opened 2023-06-22
Derive CI cache directories

askoufis:
The cache directories changed for both the jest and puppeteer caches when we moved to MacOS agents in #835. I've updated how we derive the cache directories such that we shouldn't have to make any changes if we eventually switch back to linux agents.

For ✨security reasons (probably)✨, you can't interpolate bash commands in any keys other than `run`, so we have to set output variables in another step in order to use them in the cache path.


.github/workflows/validate.yml
```diff
@@ -41,11 +41,14 @@ jobs:
...
 
+      - name: Set Puppeteer Cache Directory
+        run: echo "puppeteer_cache_dir=$HOME/.cache/puppeteer" >> "$GITHUB_ENV"
+
       - name: Cache Puppeteer
         id: puppeteer-cache
         uses: actions/cache@v3
         with:
-          path: /home/runner/.cache/puppeteer
+          path: ${{ env.puppeteer_cache_dir }}
```
mrm007, line 51: For this sort of thing, I prefer to have the output coupled to the step, like [here](https://github.com/seek-oss/crackle/blob/f6994aaa4d024adb8b515b4176d4307f8c93c3ae/.github/workflows/ci.yml#L122-L125). That way, there's no confusion as to where the value comes from.

askoufis: https://github.com/seek-oss/sku/pull/841/commits/d5716051f4195f665641057620947e75ab3f8238

mrm007 (approved):

PR #843, opened 2023-06-23
Make some more of the array types in `SkuConfig` readonly

benjervis:
This came up while we were doing the country-selection-ui statification work. In order to be able to pass in an array that has been set up with `as const`, which you often do so you can use the types elsewhere, the accepting type needs to be readonly.

Readonly fields do accept mutable versions of the value, so this doesn't create a problem for any existing consumers.


askoufis (approved):

packages/sku/sku-types.d.ts
```diff
@@ -39,16 +39,16 @@ interface SkuRouteObject {
   route: string;
   name?: string;
   entry?: string;
-  languages?: string[];
+  languages?: ReadonlyArray<string>;
```
mrm007, line 42: If you prefer the other syntax:

```suggestion
  languages?: readonly string[];
```

benjervis: Was aiming for consistency with the others here so I know where all my readonlys are

PR #845, opened 2023-06-27
Fix compile package detection

askoufis:
The `.git` folder may not be available in some situations where you want to run `sku`. Swapping to use `@manypkg/find-root` instead, as this gives us the name of the package manager as well as the root directory.

EDIT: Tested in the repo where the bug was originally found, confirmed that the fix works

### DIFF
--- a/packages/sku/context/defaultCompilePackages.js
+++ b/packages/sku/context/defaultCompilePackages.js
@@ -38,7 +38,6 @@ try {
       cwd,
     })
     .map((packagePath) => {
-      console.log({ packagePath });
       const packageJson = require(path.join(cwd, packagePath));
 
       return {
--- a/packages/sku/context/defaultCompilePackages.js
+++ b/packages/sku/context/defaultCompilePackages.js
@@ -1,37 +1,44 @@
 const { posix: path } = require('path');
 const chalk = require('chalk');
 const glob = require('fast-glob');
-const { execSync } = require('child_process');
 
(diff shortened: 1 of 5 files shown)


packages/sku/context/defaultCompilePackages.js
```diff
@@ -1,37 +1,44 @@
...
+    globs.push(pnpmVirtualStoreGlob);
+  }
 
   detectedCompilePackages = glob
-    .sync([pnpmVirtualStoreGlob, packageDependenciesGlob], {
+    .sync(globs, {
       cwd,
     })
     .map((packagePath) => {
+      console.log({ packagePath });
```
mrm007, line 41:
```suggestion
```

askoufis: lol

packages/sku/context/defaultCompilePackages.js
```diff
@@ -45,7 +51,6 @@ try {
   console.log(
     chalk.red`Warning: Failed to detect compile packages. Contact #sku-support.`,
   );
-  console.error(e);
```
mrm007, line 48: I think this was still useful.

mrm007 (approved):

packages/sku/context/defaultCompilePackages.js
askoufis: wooooooops

PR #848, opened 2023-07-03
Add `dangerouslySetTSConfig` sku config option

mrm007:
See changeset and docs for details and explanation.

Merge after #844 and #849

PR #849, opened 2023-07-03
Convert some fixtures to TypeScript

mrm007:
This PR prepares for #848.

The presence of a `sku.config.js` file implies the project is JavaScript-only and doesn't contain any TypeScript files.

https://github.com/seek-oss/sku/blob/a9a6de8c186f5366840772522e9e1bcdd03ee7b4/packages/sku/config/typescript/tsconfig.js#L8-L15

In practice, the assumption was not respected and this PR fixes that. Because of the changes in #848, the assumption must actually be correct for things to continue working as expected.


askoufis (approved):
LGTM. I'm guessing the reason for converting tests to TypeScript / adding `ts-jest` was to we can import TypeScript sku configs?

mrm007:
Yes, that's right

askoufis:
I would also argue that having a `sku.config.js` doesn't necessarily imply that the project is JS-only, but rather sku can't guarantee that there's at least one TypeScript file (without globbing your entire project), so it just adds a placeholder to make sure.

PR #853, opened 2023-07-06
Make Jest snapshots stable

mrm007:
Tired of doing the snapshot dance every time deps or the lockfile changes, which causes module ids to change, which causes the content hash to change.

We ignore the contents of those files anyway, so I don't see the point of having the hashes in snapshots. That part of the codebase is stable and won't change any time soon (👈 famous last words).


fixtures/custom-src-paths/sku.config.ts
```diff
@@ -1,9 +1,12 @@
 import type { SkuConfig } from 'sku';
+// @ts-ignore no types
```
mrm007, line 2: "Why not `ts-expect-error`?" you might ask. Because typescript-eslint and ts-jest disagree on the severity. We have [`strict: false`][false] in our monorepo and [`strict: true`][true] in fixtures.

[false]: https://github.com/seek-oss/sku/blob/ddd9ac1d571a0b02d1c8719cd7a655befd32430b/tsconfig.json#L7
[true]: https://github.com/seek-oss/sku/blob/9620c75b3d36bbcc48d6ce0634c8e9ca395a7b3a/packages/sku/config/typescript/tsconfig.js#L29

askoufis (approved):
I agree that it doesn't make sense to capture the content hash of a chunk without the content. Either you snapshot both or neither. I'm slightly worried that we're going to miss a change in a chunk that shouldn't have happened, but maybe there's other ways we can detect this.

PR #855, opened 2023-07-07
Bump eslint-config-seek

mrm007:
Also ran `pnpm dedupe`, hence the large lockfile update

### DIFF
--- a/.changeset/eslint-config.md
+++ b/.changeset/eslint-config.md
@@ -2,4 +2,9 @@
 'sku': patch
 ---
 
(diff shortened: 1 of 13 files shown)


askoufis:
Wasn't expecting such a large lockfile change. Does `pnpm dedupe` make it any better?

mrm007:
This is the result of `pnpm dedupe`.

askoufis:
Looks like there are still some test that have snapshots with content hashes in them 🤔

askoufis (approved):

PR #861, opened 2023-07-24
Allow importing external CSS from `node_modules`

mrm007:
CSS from third-party dependencies can be loaded using a side-effect import, e.g.

```tsx
import { SomeComponent } from "some-package";

import "some-package/dist/styles.css";

export const MyComponent = () => {
  return <SomeComponent>{/* ... */}</SomeComponent>;
};
```

I've also added a fixture to test styling functionality. Many fixtures do too much integration testing, and my intention is for this fixture to cover everything, removing styling from the other fixtures.


packages/sku/config/storybook/storybookWebpackConfig.js
```diff
@@ -26,7 +26,7 @@ module.exports = ({ config }, { isDevServer }) => {
           : [previousExclude]), // Ensure we don't clobber any existing exclusions
         ...paths.src,
         ...paths.compilePackages.map(resolvePackage),
-        /\.vanilla\.css$/,
+        /(\.vanilla)?\.css$/,
```
michaeltaranto, line 29: I read this as "exclude VE `.css` files from being handled by default storybook css loaders".

If we are excluding all css now, maybe this becomes `/\.css$/`, thoughts?

fixtures/styling/sku.config.ts
```diff
@@ -0,0 +1,12 @@
+import type { SkuConfig } from 'sku';
+// @ts-ignore no types
```
michaeltaranto, line 2:
```suggestion
// @ts-expect-error no types
```

packages/sku/config/webpack/utils/loaders.js
```diff
@@ -91,18 +91,18 @@ const makeCssLoaders = (options = {}) => {
...
 
-const makeVanillaCssLoaders = (options = {}) => {
-  const { isProductionBuild, MiniCssExtractPlugin, browserslist } = options;
+const makeExternalCssLoaders = (options = {}) => {
+  const { target, isProductionBuild, MiniCssExtractPlugin, browserslist } =
+    options;
 
   return [
-    MiniCssExtractPlugin.loader,
+    ...(!target || target === 'browser' ? [MiniCssExtractPlugin.loader] : []),
```
michaeltaranto, line 105: Curious what led to this change. Can you provide any context?

packages/sku/config/storybook/storybookWebpackConfig.js
mrm007: You're correct.

I left the `.vanilla` in there to be able to find references down the line. I suppose being more explicit would be better. Something like this?

```suggestion
        /\.vanilla\.css$/,
        /\.css$/,
```

packages/sku/config/webpack/utils/loaders.js
mrm007: This is the same behaviour as https://github.com/seek-oss/sku/blob/976fab0ecd97b02994a9d1fc6f60e1f1c51719e5/packages/sku/config/webpack/utils/loaders.js#L49

For VE files, we always use the loader (previous behaviour).
For non-VE files (plain CSS), we need the loader in the client build (`target` comes from [here](https://github.com/seek-oss/sku/blob/976fab0ecd97b02994a9d1fc6f60e1f1c51719e5/packages/sku/config/webpack/plugins/sku-webpack-plugin/index.js#L47)).

fixtures/styling/sku.config.ts
mrm007: https://github.com/seek-oss/sku/actions/runs/5640947831/job/15278287343?pr=861#step:13:1593

michaeltaranto (approved):

PR #863, opened 2023-07-25
Update styling docs

mrm007:
To go with #861

### DIFF
--- a/docs/docs/styling.md
+++ b/docs/docs/styling.md
@@ -1,6 +1,6 @@
 # Styling
 
-## vanilla-extract
+## Vanilla Extract
 
 The `@vanilla-extract/css` package is not available out of the box so it needs to be installed.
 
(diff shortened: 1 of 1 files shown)


docs/docs/styling.md
```diff
@@ -61,3 +64,17 @@ As of sku v12, [treat] support has been removed in favour of [Vanilla Extract].
...
+
+```tsx
+import { SomeComponent } from 'some-package';
+
+import 'some-package/dist/styles.css';
+
+export const MyComponent = () => {
+  return <SomeComponent>{/* ... */}</SomeComponent>;
+};
+```
```
michaeltaranto, line 80: Thoughts on adding a note about the scope?

````suggestion
```

Note, this should only be used when importing CSS dependencies from a third-party package in node_modules.
If you are writing custom CSS, please see the [Vanilla Extract](#vanilla-extract) section above.
````

mrm007: SGTM

michaeltaranto (approved):

PR #864, opened 2023-07-25
Allow docs to be deployed without a changeset

mrm007:
Right now a changeset + **Version Packages** PR is required to be able to make an update to the docs.
See [here](https://github.com/seek-oss/sku/actions/runs/5651226167/job/15308971217#step:6:155) how docs were not deployed until the **Version Packages** PR was [merged](https://github.com/seek-oss/sku/actions/runs/5651319335/job/15309233463#step:6:30).


mrm007:
@michaeltaranto Good point.

PR #876, opened 2023-08-24
Refactor package manager detection, add support for detecting `pnpm`

askoufis:
The primary change in this PR is refactoring how we detect package managers. This code is contained within the new `packageManager` lib, leveraging [`@manypkg/find-root`](https://www.npmjs.com/package/@manypkg/find-root).

During this refactor I found a few bugs too. They're detailed within the changesets, but I'll add comments explaining them a bit too.

There will be a followup PR after this one that will add `pnpm` support for `sku init`, likely via CLI flag. That change will be released alongside the changes in this PR.


packages/sku/context/defaultCompilePackages.js
```diff
@@ -14,9 +14,7 @@ try {
   const globs = ['node_modules/@seek/*/package.json'];
   const cwd = skuCwd();
 
-  const { rootDir, tool } = findRootSync(cwd);
-
-  if (tool === 'pnpm') {
```
askoufis, line 19: `tool` is an object, so this was always `false`.

packages/sku/lib/suggestScript.js
```diff
@@ -1,14 +1,16 @@
...
+ * @param {string} scriptContents
  */
 const findPackageScript = (scriptContents) => {
   let pkg;
   try {
-    pkg = requireFromCwd('package.json');
+    // CWD should be the root of the package
+    pkg = requireFromCwd('./package.json');
   } catch (err) {
     pkg = { scripts: {} };
```
askoufis, line 15: Resolving `package.json` from cwd (package root) would always fail, so the `catch` clause was executing every time, so scripts from `package.json` were never suggested.

packages/sku/lib/detectYarn.js
```diff
@@ -1,10 +0,0 @@
-const { execSync } = require('child_process');
-
-module.exports = () => {
-  try {
-    execSync('yarnpkg --version', { stdio: 'ignore' });
```
askoufis, line 5: `yarnpkg` is a rename that was done on debian machines I think. I think nowadays most people either just have `yarn`, or alias `yarn` to `yarnpkg`, so I'm calling the new logic package manager detection logic a "bugfix".

askoufis: Somehow, on my machine this command returns a version number, but in my terminal `yarnpkg` isn't a valid command.

askoufis:
Tests are failing because multiple storybooks are running simultaneously. There was only one fixture that used storybook, until [this PR](https://github.com/seek-oss/sku/pull/861).

One fixture has a `preview.tsx` file, which gets symlinked to sku's node modules for storybook to find. There is a cleanup step before the symlinking happens, so the fixtures clash and one cleans up the other one's symlinked file before storybook starts, causing errors.

This issue only happens if the two test suites happen to run at the same time, which seems deterministically based on something, though I don't know exactly what. It would be nice if jest had a way to explicitly disallow a test suite to run in parallel. We could make a separate folder to isolate the test, but that's not great.

I think the best option would be to stop doing all our storybook config shenanigans and just emit files to a `.storybook` folder (gitignoring main.js only), like we do with other config files, and point storybook there, rather than pointing it to sku's node modules.

This would be beneficial for a few reasons:
- Remove the symlinking shenanigans for `preview` config. Users can just write the file if they need it.
- Remove the dummy middleware file. Users can just `require`/copy/symlink their existing sku dev middleware to `.storybook` if they want it to run during `sku storybook`.
- There are other config files, namely `preview-head.html` and `preview-body.html` that sku doesn't currently support. [They allow injecting HTML into the story iframe](https://storybook.js.org/docs/react/configure/story-rendering). These could be useful for webfonts, among other things. Just using the `.storybook` folder would automatically implement support for these files too.
- This also fixes a bug where the `--config` flag was not honoured when running `sku storybook`, because sku calls the storybook CLI in a separate process and it cannot propagate that flag because the storybook CLI validates its arguments. This results in the storybook `main.js` config accessing plain `sku.config.js` (ignoring the flag), which it needs for a few config values.

This concludes my TED talk.

askoufis:
Depends on #878.

packages/sku/lib/suggestScript.js
mrm007: Good catch 🤭

packages/sku/lib/packageManager.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+const getPackageManager = () => {
+  try {
+    const { tool, rootDir } = findRootSync(cwd());
+
+    if (supportedPackageManagers.includes(tool.type)) {
+      return { packageManager: tool.type, rootDir };
+    }
+
+    throw new Error('Unsupported package manager found');
```
mrm007, line 41: A small nitpick: swapping the condition to throw when the condition is not met.

packages/sku/lib/packageManager.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+const detectPackageManagerCommand = (commandName) =>
+  which(commandName, { nothrow: true }) ? commandName : null;
+
+const detectPackageManager = () => {
+  return (
+    detectPackageManagerCommand('yarn') ||
+    detectPackageManagerCommand('pnpm') ||
+    detectPackageManagerCommand('npm') ||
+    'npm'
```
mrm007, line 23: Seems not worth calling `detectPackageManagerCommand('npm')`.

```suggestion
    'npm'
```

.changeset/gorgeous-nails-film.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+Fixes a bug where `pnpm` was not detected correctly when detecting compile packages
```
mrm007, line 5:
```suggestion
Fixes a bug where `pnpm` was not detected correctly when detecting compiled packages
```

perhaps?

mrm007 (commented):
A few comments related to `@antfu/ni`, but looks good otherwise 👍

mrm007 (commented):
A few comments related to `@antfu/ni`, but looks good otherwise 👍

packages/sku/lib/suggestScript.js
```diff
@@ -30,22 +42,19 @@ const findPackageScript = (scriptContents) => {
...
-    script += `npx sku ${scriptName}`;
-  }
+  const packageOrSkuScript = packageScript
+    ? packageScript
+    : `sku ${scriptName}`;
+
+  const sudoPrefix = options.sudo ? 'sudo ' : '';
 
-  return script;
+  return `${sudoPrefix}${packageManagerCommand} ${packageOrSkuScript}`;
```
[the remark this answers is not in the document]
askoufis: I didn't even know `@antfu/ni` was more than just a CLI. Definitely would prefer to use something else than building our own package manager command abstraction. 👍

packages/sku/package.json
```diff
@@ -127,6 +127,7 @@
     "webpack-dev-server": "4.11.1",
     "webpack-merge": "^5.8.0",
     "webpack-node-externals": "^3.0.0",
+    "which": "^3.0.1",
```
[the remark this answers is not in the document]
askoufis: Oh, didn't realize they had a release so recently!

packages/sku/lib/validatePeerDeps.js
```diff
@@ -55,11 +56,9 @@ module.exports = async () => {
             .join('\n'),
         );
 
-        if (detectYarn()) {
-          messages.push(
-            chalk`Try running "{blue.bold yarn why} {bold ${packageName}}" to diagnose the issue`,
-          );
-        }
+        messages.push(
+          chalk`Try running "{blue.bold ${packageManager} why} {bold ${packageName}}" to diagnose the issue`,
```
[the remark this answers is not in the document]
askoufis: Good point. `-r` works in single repos too, so might as well add the flag.

.changeset/gorgeous-nails-film.md
askoufis: I went for **compile** since we are the ones that compile the packages, rather then detecting already **compiled** packages.

mrm007: In that case, might be worth adding a link.
```suggestion
Fixes a bug where `pnpm` was not detected correctly when detecting [compile packages](https://seek-oss.github.io/sku/#/./docs/extra-features?id=compile-packages)
```

jahredhope (approved):

packages/sku/telemetry/provider.js
```diff
@@ -1,3 +1,6 @@
+const { packageManager, isYarn } = require('../lib/packageManager');
+const { getCommand } = require('@antfu/ni');
```
mrm007, line 2: I was thinking it would be ideal to absorb `@antfu/ni` into `lib/packageManager` so there's only one way to interface with the package manager. But we can iterate on that later.

mrm007 (approved):
Nice work 👍

packages/sku/telemetry/provider.js
askoufis: Good idea. I've got a follow up PR that will use `getCommand` a bit more, I can refactor it there.

PR #878, opened 2023-08-29
Support Storybook config directory

askoufis:
For reasons outlined in [my PR comment](https://github.com/seek-oss/sku/pull/876#issuecomment-1692641632), as well as an internal confluence document, leaning into the standard `.storybook` configuration folder rather than pointing into sku's node modules addresses some issues and reduces the complexity of sku's storybook integration.

This includes:
- Remove the symlinking shenanigans for preview config. Users can just write the file if they need it.
- Remove the dummy middleware file. Users can just require/copy/symlink their existing sku dev middleware to .storybook if they want it to run during sku storybook.
- Adds support for [other config files](https://storybook.js.org/docs/react/configure/story-rendering) that sku did not support

Sku still maintains control over the `.storybook/main.js` file as this contains critical configuration that enables storybook to work with sku's wepback setup, VE, TypeScript, etc.

This has been tested internally in both a regular app and an app within a monorepo.


packages/sku/scripts/build-storybook.js
```diff
@@ -1,23 +1,20 @@
...
-const configDir = path.resolve(__dirname, '../config/storybook/build');
 const { runVocabCompile } = require('../lib/runVocab');
-const { setUpStorybookPreviewFile } = require('../lib/storybook');
+const { setUpStorybookConfigDirectory } = require('../lib/storybook');
 
 (async () => {
   await runVocabCompile();
   await rimraf(storybookTarget);
-  await setUpStorybookPreviewFile(configDir);
+  await setUpStorybookConfigDirectory();
```
michaeltaranto, line 15: Given this is setting up a controlled folder with files on disk, was there a reason to do this separately to the `configure` script?

packages/sku/scripts/storybook.js
```diff
@@ -1,21 +1,18 @@
...
 const { storybookPort } = require('../context');
 const startStorybookPath = require.resolve('@storybook/cli/bin/index');
-const configDir = path.resolve(__dirname, '../config/storybook/start');
 const { watchVocabCompile } = require('../lib/runVocab');
-const { setUpStorybookPreviewFile } = require('../lib/storybook');
+const { setUpStorybookConfigDirectory } = require('../lib/storybook');
 
 // Unshift args to allow pushing --ci as an arg during storybook-config tests
 argv.unshift('--quiet');
-argv.unshift('--config-dir', configDir);
```
michaeltaranto, line 12: With the removal of this and reliance on the `.storybook` directory, should we validate the availability of this? Not sure what the DX is like if this goes wrong, or in a monorepo or some situation with the folder isnt available to cwd.

packages/sku/config/storybook/index.js
michaeltaranto, line 1: Great comments through here 👏

packages/sku/scripts/build-storybook.js
askoufis: I chose to associate it with the `storybook` commands because not all teams actually use Storybook. So this avoids apps that don't use Storybook getting an extra folder + file, even if it is gitignored.

packages/sku/scripts/storybook.js
askoufis: Looking at [the source code](https://github.com/storybookjs/storybook/blob/a810931c9321c0d95d2eb48e23f62856f11b5592/code/lib/core-common/src/utils/get-storybook-info.ts#L86-L108), it looks like it's doing a relative path `fs.existsSync` to check for the config files, so it won't pick up `.storybook` in a parent folder for example. However, this isn't any different to how it worked before.

[The error on storybook's side](https://github.com/storybookjs/storybook/blob/a810931c9321c0d95d2eb48e23f62856f11b5592/code/lib/cli/src/add.ts#L66-L69) could be a bit more specific though.

michaeltaranto: Maybe we consider improving the DX down the track if this scenario eventuates then. Im happy with that.

michaeltaranto (approved):

PR #883, opened 2023-09-04
Remove `lodash` dependency

askoufis:
Sku's usage of `lodash` is fairly limited, and I noticed there were a few quick replacements that could be done, so I just gutted it.

Side note: I never knew about the `env` sku config parameter. It's undocumented, and [has been deprecated since `v7`](https://github.com/seek-oss/sku/blob/master/docs/migration-guides/v7.0.0.md#env). I've slated it for removal in the next major in my notes.

### DIFF
--- a/packages/sku/config/webpack/webpack.config.js
+++ b/packages/sku/config/webpack/webpack.config.js
@@ -1,7 +1,6 @@
 const webpack = require('webpack');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const nodeExternals = require('webpack-node-externals');
-const lodash = require('lodash');
 const path = require('path');
 const LoadablePlugin = require('@loadable/webpack-plugin');
 const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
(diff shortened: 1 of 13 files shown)


packages/sku/config/webpack/webpack.config.js
```diff
@@ -58,28 +58,11 @@ const makeWebpackConfig = ({
...
-        process.exit(1);
-      }
-
-      return JSON.stringify(valueForEnv);
-    })
-    .set('SKU_ENV', JSON.stringify(args.env))
-    .set('PORT', JSON.stringify(port))
-    .mapKeys((value, key) => `process.env.${key}`)
-    .value();
+  const envars = stringifyEnvarValues({
```
jahredhope, line 61: Is the variable rename intentional?

I assumed this was: `Environment Variables => env var => envVar`

jahredhope (commented):
One question

packages/sku/config/webpack/webpack.config.js
askoufis: It was intentional. `envVar` works I guess, but [`envar` is much more common](https://trends.google.com/trends/explore?date=today%205-y&q=envar,envvar&hl=en-AU).

jahredhope (approved):

PR #885, opened 2023-09-07
Update webpack-related deps, clean up some deps

askoufis:
Not really sure why we had `webpack-dev-server` pinned. Also no idea how `require-from-string` got included in [this PR from a while ago](https://github.com/seek-oss/sku/pull/710/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R146). It's currently unused, and AFAICT it never was used.

Also removed the `assert` dev dep and changed all imports to `node:assert/strict` in tests. There is one more [usage of node `assert` in a fixture](https://github.com/seek-oss/sku/blob/master/fixtures/assertion-removal/src/App.tsx#L1), but that can't be changed to `node:assert` because webpack doesn't know how to load the `node:` protocol. I think adding the [`NodeTargetPlugin`](https://webpack.js.org/plugins/internal-plugins/#nodetargetplugin) may be the solution, but don't really want to investigate that right now.

- Removed the `webpack-cli` dev dep from `sku`, as it's not used. It's only used in the `tests` private package.
- Removed the `traverse` dep from sku as it was unused


mrm007:
> because webpack doesn't know how to load the `node:` protocol

It's the [Babel plugin](https://github.com/unassert-js/babel-plugin-unassert) that needs to be configured to recognise `node:assert/strict`. You can verify it [works OOTB with `node:assert`](https://github.com/unassert-js/babel-plugin-unassert/blob/master/README.md#supported-patterns).
Webpack shouldn't even see the import, because Babel runs before Webpack.

tests/assertion-removal.test.ts
```diff
@@ -1,4 +1,4 @@
-import assert from 'assert';
+import assert from 'node:assert/strict';
```
mrm007, line 1: The change from `assert` to `node:assert/strict` is harmless for integration tests because they're not being processed by sku.

askoufis:
> > because webpack doesn't know how to load the `node:` protocol
> 
> It's the [Babel plugin](https://github.com/unassert-js/babel-plugin-unassert) that needs to be configured to recognise `node:assert/strict`. You can verify it [works OOTB with `node:assert`](https://github.com/unassert-js/babel-plugin-unassert/blob/master/README.md#supported-patterns). Webpack shouldn't even see the import, because Babel runs before Webpack.

Babel unasssert is only run during production builds/dev mode. So it doesn't work with regular `sku start`, which is the whole point.

packages/sku/package.json
```diff
@@ -140,7 +139,6 @@
     "@types/which": "^3.0.0",
     "@vanilla-extract/css": "^1.0.0",
     "@vocab/react": "^1.0.1",
-    "assert": "^2.0.0",
```
mrm007, line 143: We should bump `babel-plugin-unassert` to latest, potentially [adding](https://github.com/unassert-js/babel-plugin-unassert/blob/master/README.md#customization) `node:assert/strict` to the list of [default patterns](https://github.com/unassert-js/babel-plugin-unassert/blob/master/index.js#L15-L24).

askoufis: Good idea.

askoufis: We wouldn't want sku to polyfill it, but configuring `node:assert/strict` in the babel plugin would allow consumers to use it a dev time.

mrm007: It needs to be polyfilled to run in the browser (`sku start`).

askoufis: Oh right. I guess the dep should be added to the fixture then.

mrm007: On master, `sku start` fails for that fixture because it can't find the `assert` package. It used to work with Yarn because of hoisting.
However, we don't test `sku start` 😅

askoufis:
Also realised that `webpack-cli` wasn't used in `sku` as a dev dep, so I removed it.

askoufis:
Found another unused dep, `traverse`.

PR #888, opened 2023-09-14
Enable strict mode in tsconfig

askoufis:
Setting `strict: true` isn't super critical for sku itself as it's all still JS, but it does give us more accurate JSDoc types, which _are_ valuable (IMO).

Setting `strict: true` required a bit of finessing to get everything to work. This included:
- Adding missing type annotations of course
- Replacing `ts-jest` with `@swc/jest` and adding a `tsc` lint script. `ts-jest` was complaining about errors the TS language server was not complaining about. I couldn't figure out what the cause of this was, but also more generally I don't think we're getting much from type checking tests at test-time, as opposed to just running `tsc` over the few typescript files we have. That's why I opted to just replace `ts-jest` as it was simpler than adding `@ts-expect-error`s to satisfy _only_ `ts-jest`.
- Not strictly necessary, but I converted `process.js` to TypeScript. This file exports the `runSkuScriptInDir` utility which is used in just about every test, and having accurate types made it easier to type the TS test files with `strict: true`.


.gitignore
```diff
@@ -2,10 +2,10 @@ node_modules/
 *.log
 package-lock.json
 .idea
-entry/render/startConfig.json
```
askoufis, line 5: This file does not exist, nor is it emitted by anything AFAICT.

test-utils/jestSetup.ts
```diff
@@ -1,11 +1,3 @@
 import 'jest-puppeteer';
 
-import appSnapshotSerializer from './appSnapshotSerializer';
-import cssSnapshotSerializer from './cssSnapshotSerializer';
-import htmlSnapshotSerializer from './htmlSnapshotSerializer';
-
-expect.addSnapshotSerializer(appSnapshotSerializer);
-expect.addSnapshotSerializer(cssSnapshotSerializer);
-expect.addSnapshotSerializer(htmlSnapshotSerializer);
-
```
askoufis, line 10: These are already configured in jest config.
https://github.com/seek-oss/sku/blob/e011b59e9191309ce4bd07d86f545f9b2bcd78ef/jest.config.js#L8-L12

tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "allowJs": true,
     "jsx": "react-jsx",
     "noEmit": true,
-    "strict": false
-  }
+    "incremental": true
```
askoufis, line 8: `strict` is enabled by default in the tsconfig we extend.
`incremental` seemed like a no-brainer to enable. Basically a type checking cache. We should probably do this for the sku tsconfig as well.

package.json
```diff
@@ -5,12 +5,14 @@
   "scripts": {
     "postinstall": "node ./postinstall.js",
     "prepare": "husky install",
-    "lint": "pnpm format-check && eslint --cache .",
+    "lint": "concurrently -c auto --group 'pnpm:lint:*'",
+    "lint:eslint": "eslint --cache .",
+    "lint:prettier": "prettier --cache --list-different '**/*.{js,ts,tsx,md,less,css}'",
+    "lint:tsc": "tsc",
```
askoufis, line 11: `concurrently` should suffice for lint scripts until sku has the need for a _real_ task runner.

package.json
```diff
@@ -40,10 +42,12 @@
     "@changesets/cli": "^2.13.0",
     "@changesets/get-github-info": "^0.4.5",
     "@sku-private/test-utils": "workspace:*",
+    "@swc/jest": "^0.2.29",
```
mrm007, line 45: We need `@swc/core` as well

package.json
mrm007, line 11: 👍

packages/sku/lib/runESLint.js
```diff
@@ -3,6 +3,7 @@ const { yellow, cyan, gray } = require('chalk');
 const { ESLint } = require('eslint');
 const eslintConfig = require('../config/eslint/eslintConfig');
 const { lintExtensions } = require('./lint');
+const assert = require('node:assert');
```
mrm007, line 6: What do you think about adding an ESLint rule to require the `node:` protocol?

askoufis: That's a good idea. Had to manually update a bunch of them. Do we go with [the `eslint-plugin-unicorn` version](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/6d15a02d48de7ecfc38d0683a8487b2f937d83a0/docs/rules/prefer-node-protocol.md) or [a standalone one](https://www.npmjs.com/package/eslint-plugin-node-import)?

mrm007: Let's go with our rainbow friend 👍

askoufis: Will do it in a separate PR for ease-of-review.

tests/assertion-removal.test.ts
```diff
@@ -34,15 +36,15 @@ describe('assertion-removal', () => {
...
+      const sourceHtml = await response?.text();
       expect(sourceHtml).toContain(
         'It rendered without throwing an assertion error',
       );
     });
   });
 
   describe('build-ssr', () => {
-    let server, closeAssetServer;
+    let server: ChildProcess, closeAssetServer: () => void;
```
mrm007, line 47: Surprised there isn't a rule against multiple variable declarations. I'm not really fussed about it, but personally I prefer one var per line.

askoufis: I'm surprised too. I thought we had one 🤔. I can't actually find the rule for this, only an old reference to [this](https://jscs-dev.github.io/rule/disallowMultipleVarDecl).

mrm007 (approved):

tests/assertion-removal.test.ts
askoufis: Oh, is it just `one-var` with `never`?

mrm007: I believe the rule and config is [`one-var: ["error", "never"]`](https://eslint.org/docs/latest/rules/one-var#never).

PR #892, opened 2023-09-22
Update eslint-config-seek to version 12.0

felixhabib:


pnpm-lock.yaml
```diff
@@ -1,9 +1,5 @@
 lockfileVersion: '6.0'
 
-settings:
-  autoInstallPeers: true
-  excludeLinksFromLockfile: false
-
```
askoufis, line 6: You _might_ be on an older version of pnpm. Don't think this should've been removed.

askoufis:
You might have a hard time upgrading the eslint config only in the root, and not the actual `sku` package. CI seems to agree.

pnpm-lock.yaml
askoufis: Oh, did it just get moved to the bottom of the file? That's weird.

felixhabib: haha not sure, we think it's in the original place now tho

.changeset/gentle-buses-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+"sku": patch
+---
+
+Update eslint-config-seek to version 12.
```
askoufis, line 5: IMO it would be good to have a message here mentioning that the majority of changes should be autofixable.

benjervis (approved):

PR #894, opened 2023-10-11
Better support for React `displayName`

mrm007:
Found a few edge cases where this plugin would not work. I went through a bunch of [forks of the original][forks] and they weren't much better either.

The replacement isn't perfect, but it's better than the one we have.

[forks]: https://github.com/opbeat/babel-plugin-add-react-displayname/network


benjervis (approved):

.changeset/fast-beds-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+Better support for React `displayName`
```
mrm007, line 5:
```suggestion
Better support for setting `displayName` on React components when [`displayNamesProd`](https://seek-oss.github.io/sku/#/./docs/configuration?id=displaynamesprod) is true.
```

PR #902, opened 2023-11-20
Remove `node-emoji` dependency

askoufis:
[Issues](https://github.com/seek-oss/sku/pull/900) aside, I don't think there's enough value in having this dependency given that we only use a small subset of its API and we can just use the emoji characters directly.

### DIFF
--- a/packages/sku/config/webpack/plugins/sku-webpack-plugin/validateOptions.js
+++ b/packages/sku/config/webpack/plugins/sku-webpack-plugin/validateOptions.js
@@ -6,11 +6,9 @@ const didYouMean = require('didyoumean2').default;
 const validator = new Validator();
 
 const exitWithErrors = async (errors) => {
-  const { emojify } = await import('node-emoji');
-
   console.log(bold(underline(red('SkuWebpackPlugin: Invalid options'))));
   errors.forEach((error) => {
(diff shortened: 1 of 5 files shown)


packages/sku/config/webpack/plugins/sku-webpack-plugin/validateOptions.js
```diff
@@ -6,11 +6,9 @@ const didYouMean = require('didyoumean2').default;
 const validator = new Validator();
 
 const exitWithErrors = async (errors) => {
-  const { emojify } = await import('node-emoji');
```
mrm007, line 9: The perf gains are going to be 🤯

mrm007 (approved):
Have you tested these changes in the terminal? Some emoji have a double width, so they need another space to preserve alignment e.g. [this](https://github.com/seek-oss/crackle/blob/7c71eb25ddd2e2b9a0e4634ac227e727a7b170c5/packages/core/entries/render/build.tsx#L76) vs [this](https://github.com/seek-oss/crackle/blob/7c71eb25ddd2e2b9a0e4634ac227e727a7b170c5/packages/core/entries/render/build.tsx#L106).

askoufis:
LGTM
image

PR #906, opened 2023-11-24
Run `.cjs` and `.mjs` files through preset env

askoufis:
This will affect packages like Braid and our internal component libraries as they're compiled with crackle, but also any 3rd party dependencies that are shipping `.cjs`/`.mjs` source files.

For now we're depending on the extension lists provided by `eslint-config-seek`, but this may change in the future.

### DIFF
--- a/packages/sku/config/webpack/utils/index.js
+++ b/packages/sku/config/webpack/utils/index.js
@@ -1,12 +1,18 @@
 const loaders = require('./loaders');
 const { resolvePackage } = require('./resolvePackage');
-const { js, ts } = require('eslint-config-seek/extensions');
+
+// TODO: Figure out a better way to share this config. This is currently copy-pasted from
+// `eslint-config-seek`
+const extensions = {
+  js: ['js', 'cjs', 'mjs', 'jsx'],
+  ts: ['ts', 'cts', 'mts', 'tsx'],
+};
 
 module.exports = {
   ...loaders,
   resolvePackage,
-  TYPESCRIPT: new RegExp(`\.(${ts.join('|')})$`),
-  JAVASCRIPT: new RegExp(`\.(${js.join('|')})$`),
+  TYPESCRIPT: new RegExp(`\.(${extensions.ts.join('|')})$`),
+  JAVASCRIPT: new RegExp(`\.(${extensions.js.join('|')})$`),
   LESS: /\.less$/,
   IMAGE: [/\.bmp$/, /\.gif$/, /\.jpe?g$/, /\.png$/],
   SVG: /\.svg$/,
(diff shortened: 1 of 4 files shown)


askoufis (commented):
Bit confused about the snapshot changes. I guess a `buffer` polyfill was being bundled, but now isn't? And `assert` is now no-longer externalized because the `unassert` plugin is now correctly running on Braid?

packages/sku/config/webpack/utils/index.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
   ...loaders,
   resolvePackage,
   TYPESCRIPT: /\.(ts|tsx)$/,
-  JAVASCRIPT: /\.js$/,
+  JAVASCRIPT: /\.(js|cjs|mjs)$/,
```
mrm007, line 8: Please use the extensions defined in `eslint-config-seek`.

We should have a single source of truth to avoid such situations in the future. Case in point: this doesn't include `jsx`, which is supported by our tooling ([1], [2], [3], [4]).

[1]: https://github.com/seek-oss/skuba/blob/2c640b6aeffff78f76ec7232b11fffca35bebc90/.eslintrc.js#L7
[2]: https://github.com/seek-oss/sku/blob/51186f39b0855c5c8b011810216a3d308b78a82b/packages/sku/lib/lint.js#L1-L4
[3]: https://github.com/seek-oss/sku/blob/721990812bb10d5c5879d19e15af29b652c5bbf7/packages/sku/lib/runPrettier.js#L7
[4]: https://github.com/seek-oss/eslint-config-skuba/blob/579a1fd97b25bf2023c5943ab7fca7078ca7c88f/index.js#L2

askoufis: I thought we were just going to get this out and then gold plate it later. I'm honestly unsure if we want to couple this config to the eslint config, not because I don't want to have a single source of truth, but more so that the list from the config is quite load bearing now, and I'm not sure if it's the best place for the source of truth the live.

askoufis: That being said, I don't see a reason not to include `jsx`. We'll use the extensions for now.

packages/sku/config/webpack/utils/index.js
```diff
@@ -1,11 +1,12 @@
 const loaders = require('./loaders');
 const { resolvePackage } = require('./resolvePackage');
+const { js, ts } = require('eslint-config-seek/extensions');
```
jahredhope, line 3: Depending on a linting plugin seems dangerous.

One might change the other package only thinking you are changing what code gets linted, but suddenly find you've changed production code.

Especially given the other extensions are explicit I'd like to see this kept explicit for now.

askoufis:
For a future PR: sku exposes a tiny package that contains the extensions, e.g. `@seek-sku/file-extensions`. `sku` and `eslint-config-seek` consume this package.

michaeltaranto (approved):

jahredhope (approved):

PR #913, opened 2023-12-15
Report `braid-design-system` version via telemetry

askoufis:
I initially tried getting the braid version from the app's `package.json` dependencies, but the version _could_ be a caret version (shows up as `_32.0.0` in datadog), for example in a library, so I instead opted to fetch it from the version in the app's `node_modules`.


packages/sku/telemetry/index.js
```diff
@@ -1,24 +1,32 @@
...
 try {
   const packageJson = require(getPathFromCwd('package.json'));
 
   if (packageJson.name) {
     projectName = packageJson.name;
   }
+
+  const braidPackageJson = require(require.resolve(
+    'braid-design-system/package.json',
+  ), [cwd()]);
```
mrm007, line 21:
```suggestion
  const braidPackageJson = requireFromCwd('braid-design-system/package.json');
```

👀 

https://github.com/seek-oss/sku/blob/721990812bb10d5c5879d19e15af29b652c5bbf7/packages/sku/lib/cwd.js#L25-L29

packages/sku/telemetry/index.js
```diff
@@ -1,24 +1,32 @@
 const os = require('node:os');
 
-const { getPathFromCwd } = require('../lib/cwd');
+const { getPathFromCwd, cwd } = require('../lib/cwd');
```
mrm007, line 3:
```suggestion
const { requireFromCwd } = require('../lib/cwd');
```

packages/sku/telemetry/index.js
```diff
@@ -1,24 +1,32 @@
...
 const isCI = require('../lib/isCI');
 const provider = require('./provider');
 const skuVersion = require('../package.json').version;
 
 const { languages } = require('../context');
 
 let projectName = 'unknown';
+let braidVersion = 'unknown';
 try {
   const packageJson = require(getPathFromCwd('package.json'));
```
mrm007, line 13:
```suggestion
  const packageJson = requireFromCwd('package.json');
```

packages/sku/telemetry/index.js
```diff
@@ -1,24 +1,32 @@
...
   if (packageJson.name) {
     projectName = packageJson.name;
   }
+
+  const braidPackageJson = require(require.resolve(
+    'braid-design-system/package.json',
+  ), [cwd()]);
+
+  braidVersion = braidPackageJson.version;
 } catch (e) {}
```
mrm007, line 24: Can we add a `require('debug')('sku:telemetry')` message here?

askoufis: For the `catch`? Sure. The telemetry package already spits out debug logs for the stuff it handles. https://github.com/SEEK-Jobs/sku-telemetry/pull/17

mrm007: I meant on the sku side, so we see the error if any of the `package.json`s can't be loaded.

packages/sku/telemetry/index.js
askoufis: Ooo, should've scouted this out, felt like I was re-implementing something that already existed.

mrm007 (approved):

PR #951, opened 2024-04-18
Support dangerouslySetPrettierConfig

AaronMoat:
I have what I think is a legitimate use case for customisation in our closed source - happy to share in Slack if you want the context. 

@ Maintainers - the CONTRIBUTING appears quite out of date so I'm unsure if there's any due diligence I should be doing in addition, e.g. is there a practise of snapshotting/running locally on real repos?

### DIFF
--- a/.changeset/curvy-cherries-joke.md
+++ b/.changeset/curvy-cherries-joke.md
@@ -1,0 +1,5 @@
+---
+'sku': minor
+---
+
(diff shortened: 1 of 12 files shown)


AaronMoat:
Spoke with Adam and currently prefer not to upstream this unless there are other legitimate / non-aesthetic reasons (of 
which what I hit is one and perhaps the only so far)

PR #952, opened 2024-04-19
Replace `fast-glob` with `fdir` and `picomatch`, optimize peer dep validation crawler

askoufis:
## Replace `fast-glob` with `fdir` and `picomatch`

[`fdir`](https://www.npmjs.com/package/fdir) is a 0 dependency directory crawler with optional glob support via a peer dependency. For our use cases, we don't need most of the features of `fast-glob` (or dependencies), but we do need globbing, which `fdir` supports via [`picomatch`](https://www.npmjs.com/package/picomatch), a 0 dependency glob matcher.

## Replacing globs with simple filters

Due to API differences and some changes in how we crawl in certain cases, I've replaced some globs with simple filters.

## Optimize peer dep validation crawler

`validatePeerDeps` was crawling all of `node_modules` (and `.pnpm` if present) many times. We now crawl these places just once, and process the results appropriately. This results in a _much_ faster output of the warning messages.

Running `sku build` in a moderately sized app where I modified `node_modules` to induce a peer dependency warning, the warning was output within a few seconds. On sku `11.5.0` this took >1 minute.

## Disable peer dependency validation for pnpm

Our current method of peer dep validation doesn't work for pnpm. This is because it relies on crawling through nested `node_modules` directories, which pnpm only uses to store binaries. 

If we really want this functionality for pnpm, we'll need to implement it some other way.

> [!NOTE]
> Not marking this change as breaking as it wasn't doing anything in the first place.


packages/sku/package.json
```diff
@@ -104,6 +104,7 @@
     "node-html-parser": "^6.1.1",
     "open": "^7.3.1",
     "path-to-regexp": "^6.2.0",
+    "picomatch": "^3.0.1",
```
askoufis, line 107: `picomatch` v4 is the latest, but `fdir`'s peer dependency hasn't been updated to support it. We're not really missing anything by staying on v3 for now.

michaeltaranto (approved):

PR #955, opened 2024-04-23
Update contribution guide, small readme format change

askoufis:
A long overdue update. Happy to iterate on this re: open source/outside contributors, but it should be good enough for now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,8 +15,6 @@ We'd be happy to help out!
 1. Clone the repo
 1. Install the appropriate versions of `pnpm` and `node` as specified in `package.json` and `.nvmrc`. This repo is configured to use [`volta`]
    for managing toolchain dependencies, but feel free to use `nvm`, `corepack`, or whatever suits you best.
-1. Setup your hosts file for tests by running `sudo pnpm setup-test-hosts`
-1. Run `pnpm run test` to ensure everything is working as expected
 
 [`volta`]: https://volta.sh/
 
@@ -36,6 +34,19 @@ git checkout -b add-my-cool-new-feature
 ```
 
 ## Testing Your Changes
+
+Before running tests, some entries need to be added to your `/etc/hosts` file.
+
+| Hostname            | IP Address  |
+| ------------------- | ----------- |
+| `dev.seek.com.au`   | `127.0.0.1` |
+| `dev.jobstreet.com` | `127.0.0.1` |
+
+This can either be done manually, or via the following script:
+
+```sh
+sudo pnpm setup-test-hosts
+```
 
 To run the test suite locally:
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -2,71 +2,79 @@
...
 ```
 
+## Testing Your Changes
+
 To run the test suite locally:
 
-```bash
-$ npm test
+```sh
+pnpm run test
```
adenj, line 42: Might be worth noting they'll need to run the setup-test-hosts script
```sh
sudo pnpm setup-test-hosts
```

packages/sku/README.md
```diff
@@ -6,15 +6,18 @@
...
 
-Quickly get up and running with a zero-config development environment, or optionally add minimal config when needed. Designed for usage with [braid-design-system](https://github.com/seek-oss/braid-design-system), although this isn't a requirement.
+Quickly get up and running with a zero-config development environment, or optionally add minimal config when needed.
+Designed for usage with [braid-design-system](https://github.com/seek-oss/braid-design-system), although this isn't a requirement.
 
 This tool is heavily inspired by other work, most notably:
 
 - [facebookincubator/create-react-app](https://github.com/facebookincubator/create-react-app)
 - [insin/nwb](https://github.com/insin/nwb)
 - [NYTimes/kyt](https://github.com/NYTimes/kyt)
```
adenj, line 16: I get that once upon a time sku drew inspiration from these projects, I'm wondering if we still need it here? Especially considering kyt looks to be the only one thats actively maintained.

packages/sku/README.md
```diff
@@ -6,15 +6,18 @@
...
 
-**WARNING: While this software is open source, its primary purpose is to improve consistency, cross-team collaboration and code quality at SEEK. As a result, it’s likely that we will introduce more breaking API changes to this project than you’ll find in its alternatives.**
+> [!WARNING]
+> While this software is open source, its primary purpose is to improve consistency, cross-team collaboration and code quality at SEEK.
+> As a result, it’s likely that we will introduce more breaking API changes to this project than you’ll find in its alternatives.
```
adenj, line 20: If we do remove the "inspirations" above, may be worth suggesting Vite as an alternative.

CONTRIBUTING.md
askoufis: Good catch, forgot about that step.

packages/sku/README.md
askoufis: Feel like this needs rewording altogether. The message should probably be more geared towards clarifying that this was developed for SEEK use, and that consumers should probably use something else, rather than highlighting breaking changes, which aren't really too relevant (vite break more often that sku does anyway).

adenj (approved):

PR #957, opened 2024-05-06
Ensure all gitignored sku-generated files are also ignored by prettier and eslint

askoufis:
We were creating the `lintIgnorePatterns` array before modifying it later on. The effect of the change should be self-explanatory based on the changed files.

> [!NOTE]
> It turns out that we were never actually linting/formatting these files anyway. This was because we are explicit about which file extensions we lint/format, but these files don't have any file extensions, so they were effectively being ignored.

I also converted a lot of the `configure` tests to use snapshots rather than asserting on each array item. These were effectively manual snapshots, which were a bit tedious to update when the time came to do so.


tests/configure.test.ts
```diff
@@ -135,34 +134,35 @@ describe('configure', () => {
...
-        expect(ignoreContents).toContain('*.less.d.ts');
-        expect(ignoreContents).toContain(`.storybook/main.js`);
-        expect(ignoreContents).toContain(`${skuConfig.target}/`);
-        expect(ignoreContents).toContain(`${skuConfig.storybookTarget}/`);
-        expect(ignoreContents).toContain(`${bundleReportFolder}/`);
-        expect(ignoreContents).toContain(`${coverageFolder}/`);
-      }),
+
+        expect(ignoreContents).toMatchSnapshot();
+      },
```
michaeltaranto, line 165: Thoughts on bringing this inline like the above snapshot?

michaeltaranto (approved):

tests/configure.test.ts
askoufis: Ordinarily I would, but you can't have inline snapshots inside an `it.each` unfortunately.

PR #961, opened 2024-05-07
Remove `rimraf` dependency in favour of Node.js's `rm` method

askoufis:
Node's `rm` supports the `recursive` and `force` flags, so we don't need `rimraf` anymore.

Depends on #960 for CI to pass.

EDIT: Hoisting `opencollective-postinstall`, because who knows why [this error](https://github.com/seek-oss/sku/actions/runs/8995872941/job/24711492129#step:5:25) is happening.


jahredhope (approved):

jahredhope:
FYI Link step failed.

askoufis:
> FYI Lint step failed.

> Depends on https://github.com/seek-oss/sku/pull/960 for CI to pass.

PR #962, opened 2024-05-07
Ensure `sku-init` test does a thorough cleanup, run it separately to other tests

askoufis:
This fixes the same issue that #960 fixes, but now I have a better understanding of the problem, but still not an _exact_ root cause. #960 will still be merged, but just as an update to pnpm v9.

> [!NOTE]
> There are some hoisting changes in this PR, in addition to the fix. These were done as an attempt to fix the problem, but didn't actually do anything. I've left them in because they're still worth doing IMO.

## The Problem

The `sku-init` test runs the sku CLI's `init` command. As part of `sku init`, the newly-created app has its dependencies installed. In the past, this has barely modified the lockfile, only adding an entry for the `sku-init` workspace package, and nothing else.

For some reason, potentially related to peer dependencies, a second `@vanilla-extract/css` package is now being installed during the `sku-init` test. This interferes with style loading in one of the `braid-design-system` tests, probably because the `sku-init` test tends to run before the `braid-design-system` suite in CI.

The result is `master` (and other branches) tests failing with [this error](https://github.com/seek-oss/sku/actions/runs/8964836135/job/24617261550#step:13:848).

## The Fix

The `sku-init` test file no longer runs alongside all the other tests in the `tests` directory. Instead, it is run _after_ all the other tests, so as not to affect them. Additionally, the test now runs a `pnpm install` after completion in order to undo any of the changes to the lockfile/`node_modules` it makes. This doesn't matter on CI, but it allows all tests to be run locally without interfering with one another.


jest.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   transform: {
     '^.+\\.(t|j)sx?$': '@swc/jest',
   },
+  maxWorkers: process.env.CI ? 2 : undefined,
```
askoufis, line 6: Setting this here as we can no longer pass the flag to `pnpm test` anymore because that script now runs multiple other scripts.

kjots (approved):

PR #964, opened 2024-05-09
Undo pnpm v9

askoufis:
[CI is broken in `master`](https://github.com/seek-oss/sku/actions/runs/9012051956/job/24760613037#step:5:27) because #960 was accidentally merged even though the fix it proposed didn't work. This PR is a revert of #960, except for the pnpm setup action update.

EDIT: Seems fixed for now, will keep PR open though.


askoufis:
The error seems to have fixed itself as of today. Going to keep this PR open for now, as I'm skeptical that the fix is permanent.

PR #966, opened 2024-05-14
Add support for removing functions named `invariant` and assertions from `tiny-invariant`

askoufis:
## Support `invariant` function and `tiny-invariant` library

[`tiny-invariant`](https://www.npmjs.com/package/tiny-invariant) is a small, 0 dependency assertion library. The standard way of importing its assertion function is to name it `invariant`, which is why I added it to the `variables` config for `babel-plugin-unassert`.

> [!NOTE]
> Technically this changes removes support for removing assertions from [`power-assert`](https://www.npmjs.com/package/power-assert), because we're now overriding [the default config](https://github.com/unassert-js/babel-plugin-unassert?tab=readme-ov-file#options). There is no internal usage of `power-assert` that I'm aware of, so not marking this change as breaking.

## Why should we support `tiny-invariant`?

For dev-time assertions in the browser, usually we'd reach for the [`assert`](https://www.npmjs.com/package/assert) library. This is what Braid uses. However, it has [**an unnecessarily large dependency graph**](https://npmgraph.js.org/?q=assert#zoom=w) for an assertion library.

Hopefully we can eventually replace `assert` with `tiny-invariant` within Braid. Part of that involves supporting an alternative assertion library in sku. Not sure when the change in Braid will happen, as it depends on consumers using a version of sku that supports `tiny-invariant` (i.e. whatever version this change is released in).

## Tradeoffs between `assert` and `tiny-invariant`

`assert` supports checking non-primitive values, has more useful error messages, and offers various assertion functions. `tiny-invariant` just asserts on a user-provided condition, throwing an error if it fails. That's it. This is likely suitable for most use cases, and will probably work just fine in Braid and most front-end apps/libraries.

## Meme

<details>
<summary>Show</summary>

drug eyes polyfills meme

</details>

### DIFF
--- a/docs/docs/extra-features.md
+++ b/docs/docs/extra-features.md
@@ -109,8 +109,8 @@ export const Rating = ({ rating }) => <div>...</div>;
 
 ### Supported Assertion Function Names
 
-- `assert`
 - `invariant`
+- `assert`
 
 ### Supported Assertion Libraries
 
@@ -119,6 +119,7 @@ export const Rating = ({ rating }) => <div>...</div>;
 - `node:assert` ([Node.js built-in])
 
 Any combination of function name and library name is supported.
+[`tiny-invariant`] is recommended over [`assert`][browser port] due to its simplicity and size.
 
 [`tiny-invariant`]: https://www.npmjs.com/package/tiny-invariant
 [Node.js built-in]: https://nodejs.org/api/assert.html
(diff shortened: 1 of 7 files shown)


docs/docs/extra-features.md
```diff
@@ -105,6 +105,25 @@ import React from 'react';
 export const Rating = ({ rating }) => <div>...</div>;
 ```
 
+[`babel-plugin-unassert`]: https://github.com/unassert-js/babel-plugin-unassert
+
+### Supported Assertion Function Names
+
+- `assert`
+- `invariant`
```
michaeltaranto, line 113: Suggest flipping the order here if we feel that `invariant` is the preferred option, possibly adding a one liner as to why

michaeltaranto (approved):

PR #977, opened 2024-05-28
Add docs for sku's API/entrypoints

askoufis:
- Added docs for all sku entrypoints. Currently, consumers can import from any subpath within `sku`, so I've only documented the ones we officially support. In the next major version, these will be explicitly defined in `package.json#exports`, preventing imports from arbitrary subpaths.
- Added the v12 migration guide link to the navbar
- Added a docsify plugin that adds a copy code button to code blocks

Copy code button demo:
output

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -14,7 +14,7 @@ import type { SkuConfig, Render } from 'sku';
 
 ## `sku/@loadable/component`
 
-A re-export of the `@loadble/component` package, which `sku` provides as a dependency.
+A re-export of the `@loadable/component` package, which `sku` provides as a dependency.
 
 Example:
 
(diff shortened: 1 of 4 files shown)


docs/docs/api.md
```diff
@@ -0,0 +1,74 @@
...
+
+Example:
+
+```ts
+import type { SkuConfig, Render } from 'sku';
+```
+
+## `sku/@loadable/component`
+
+A re-export of the `@loadble/component` package, which `sku` provides as a dependency.
```
michaeltaranto, line 17:
```suggestion
A re-export of the `@loadable/component` package, which `sku` provides as a dependency.
```

michaeltaranto (approved):
One minor typo. Other than that 🚀

PR #979, opened 2024-05-30
sku v13: Set `sourceMapsProd` default value to `true`

askoufis:
We should make this the default to align with our recommended behaviour of publishing production sourcemaps.


.changeset/many-deers-check.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': major
+---
+
+`sourceMapsProd` is now `true` by default
```
jahredhope, line 5: I'd make it REALLY easy for someone going through a breaking change to see how they can resolve breaks:

```suggestion
`sourceMapsProd` is now `true` by default

To return to the previous behaviour set `sourceMapsProd: false` in your sku.config.js
```

jahredhope (approved):

askoufis:
Updated the docs and changeset to mention potential heap memory exhaustion that can result from prod source maps, as well as a way to mitigate the issue. This info was added after it was raised internally that production source maps can be quite expensive. 

Instead of just documenting the issue, one way to reduce friction for users could be to set `NODE_OPTIONS=--max-old-space-size=4096` in sku's bin file so consumers don't have to set it themselves. In cases where consumers need to increase the limit, we could also provide an environment variable to allow overriding our set value. However, this feels a bit redundant when a consumer can just set the option themselves from the get go.

Another option could be to run the webpack build in a separate process, so sku can directly control its `--max-old-space-size`. This isn't really feasible though because we use a programmatic webpack config, so running it in a separate process would be tricky (we can't just run the CLI and point it to a file). It also has the same issue where we need to allow configuring the memory amount.

Overall I'm happy with just documenting this for now, and if it results in a significant number of internal support issues, we can do something about it then.

docs/docs/extra-features.md
```diff
@@ -9,8 +9,8 @@ If you want to use a currently unsupported format feel free to submit a PR or co
 
 ## Source maps
 
-Source maps are enabled by default when running the `sku start` command.
-However, if you want to generate source maps when running `sku build`, you can do so by enabling [`sourceMapsProd`](./docs/configuration#sourcemapsprod).
+Source maps are enabled by default when running both `sku start` and `sku build`.
```
michaeltaranto, line 12: Is there a reason they are enabled for `start`? Would assume this is dev and unminified, not requiring the source map?

askoufis: We still need sourcemaps in dev because otherwise logs/errors end up pointing directly to dev-server bundles which contain webpack stuff transpiled code, rather than source code, e.g. you see the error in `main.js` instead of the exact line in your `App.tsx`.

michaeltaranto (approved):

PR #980, opened 2024-05-30
sku v13: Remove old configuration values and CLI arguments

askoufis:
Removing unused and deprecated config values and CLI args. See changesets for details.

## Usage

- [0 usage of `--tenant`](https://github.com/search?q=org%3ASEEK-Jobs+NOT+is%3Aarchived+path%3Apackage.json+%2Fsku.*--tenant+%2F&type=code)
- [0 usage of `--env`](https://github.com/search?q=org%3ASEEK-Jobs+NOT+is%3Aarchived+path%3Apackage.json+%2Fsku.*--env+%2F&type=code)
- [0 usage of `--build`](https://github.com/search?q=org%3ASEEK-Jobs+NOT+is%3Aarchived+path%3Apackage.json+%2Fsku.*--build+%2F&type=code)
- [A few usages of the `env` config](https://github.com/search?q=org%3ASEEK-Jobs+NOT+is%3Aarchived+path%3Asku.config.*+%22env%3A%22&type=code). Not too concerned as the migration path should be straightforward.


packages/sku/lib/parseArgs.js
```diff
@@ -19,24 +17,10 @@ module.exports = (processArgv) => {
...
-      ],
-      default: {
-        env: 'production',
-        tenant: '',
-      },
+      string: ['config', 'environment', 'packageManager', 'site', 'stats'],
       alias: {
-        e: 'env',
+        e: 'environment',
         t: 'tenant',
```
michaeltaranto, line 23: This should go too right?

packages/sku/config/webpack/webpack.config.js
```diff
@@ -43,10 +40,8 @@ const {
   externalizeNodeModules,
 } = config;
 
-// port is only required for dev builds
 const makeWebpackConfig = ({
   isIntegration = false,
-  port = 0,
```
michaeltaranto, line 49: Can you clarify the removal of `port`? Not immediately obvious from the diff

askoufis: `port` isn't specified anywhere `makeWebpackConfig` is called, and the only usage of it was in the now-deleted `stringifyEnvarValues`.

I don't think anyone is depending on `process.env.PORT` inside their app code. A high-level look at [a github search for `process.env.PORT`](https://github.com/search?q=org%3ASEEK-Jobs+NOT+is%3Aarchived+process.env.PORT&type=code&p=1) aligns with that assumption.

packages/sku/lib/parseArgs.js
askoufis: https://github.com/seek-oss/sku/pull/980/commits/1ced58d3a44487409ddf70eb8674c397f1ec65ed

michaeltaranto (approved):

PR #981, opened 2024-05-30
Add `Server` type for server entrypoint, unit test CSP functionality

askoufis:
There's a few hand-made server entrypoint types in the wild. `sku` should provide a canonical one. Also added some unit tests for CSP functionality.


packages/sku/entry/csp.js
```diff
@@ -31,18 +45,22 @@ export default function createCSPHandler({
...
-    } else if (
-      !scriptTypeIgnoreList.includes(scriptNode.getAttribute('type'))
-    ) {
-      addScriptContents(scriptNode.firstChild.rawText);
+      return;
+    }
+
+    const scriptType = scriptNode.getAttribute('type');
+    if (scriptType == null || !scriptTypeIgnoreList.includes(scriptType)) {
+      addScriptContents(scriptNode.firstChild?.rawText);
```
askoufis, line 59: Type-checking this file resulted in this small refactor.

packages/sku/entry/csp.js
```diff
@@ -53,18 +71,18 @@ export default function createCSPHandler({
...
-    ) {
+    if (process.env.NODE_ENV !== 'production' && !valid(script)) {
       console.error(`Invalid script passed to 'registerScript'\n${script}`);
     }
 
-    parse(script, { script: true })
+    parse(script, { blockTextElements: { script: true } })
       .querySelectorAll('script')
       .forEach(processScriptNode);
   };
```
askoufis, line 81: The `script` property is now under `blockTextElements`. It actually defaults to `true`, so we don't _really_ need it, but probably best to be explicit. Unit tests were added to ensure this change didn't break anything.

packages/sku/entry/csp.js
```diff
@@ -108,7 +125,16 @@ export default function createCSPHandler({
 
     root.querySelectorAll('script').forEach(processScriptNode);
 
-    root.querySelector('head').insertAdjacentHTML('afterbegin', createCSPTag());
+    const headElement = root.querySelector('head');
+    if (!headElement) {
```
askoufis, line 129: Type-checking this file meant we needed to handle this error in some way.

.changeset/shaggy-grapes-push.md
michaeltaranto, line 1: Is this a breaking change? The potential "break" is only if consumers import and assign the new type, in which case they can refactor to satisfy the correct type.

packages/sku/entry/csp.js
michaeltaranto: Should we be consistent and add it to the `valid` call [on line 74]?

[on line 74]: https://github.com/seek-oss/sku/pull/981/files#diff-890cc153983cf273d29b3e92fe5218d82a8661da2cbde0582d32b4f8a445ce99R74

packages/sku/entry/csp.js
```diff
@@ -90,13 +108,12 @@ export default function createCSPHandler({
...
+  /**
+   * @param {string} html
+   * @returns {string}
+   */
   const handleHtml = (html) => {
-    const root = parse(html, {
-      script: true,
-      style: true,
-      pre: true,
-      comment: true,
```
michaeltaranto, line 98: Are these all defaults? Should we define and share a single config for all `parse` and `valid` calls?

packages/sku/sku-types.d.ts
michaeltaranto, line 1: ❤️ 👏

.changeset/shaggy-grapes-push.md
askoufis: Good point. Updated in https://github.com/seek-oss/sku/pull/981/commits/4d0db85fa04c3f54b32ede96248ad633e006e39e.

packages/sku/entry/csp.js
askoufis: This `parse` call is the only one that cares about more than `script` tags. The others can probably just drop the explicit config actually.

In fact, I'm realising now that [the default actually doesn't preserve HTML comment structure](https://github.com/taoqf/node-html-parser/blob/a3e41a9c9dd8d229a03ace04fff054f9baf7ed88/src/nodes/html.ts#L1001).

askoufis: https://github.com/seek-oss/sku/pull/981/commits/ea6525c34ece493695aa167ccc810c60cdf75d97

packages/sku/entry/csp.js
askoufis: Have removed `script` as it's the `true` by default. https://github.com/seek-oss/sku/pull/981/commits/ea6525c34ece493695aa167ccc810c60cdf75d97

kjots (approved):

PR #983, opened 2024-05-30
sku v13: Provide types for iterable DOM APIs, remove type-only imports during transpilation

askoufis:
As iterable API usage becomes more prevalent, it will be useful to have types for these APIs available out of the box, so consumers don't need to use `dangerouslySetTsConfig`.

[`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) (and its babel equivalents) is pretty standard TypeScript config thesedays. It enables much simpler and more predictable behaviour for type imports.


packages/sku/config/babel/babelConfig.js
```diff
@@ -79,6 +79,10 @@ module.exports = ({
           {
             isTSX: true,
             allExtensions: true,
+            // babel equivalent of tsconfig `verbatimModuleSyntax: true`
+            onlyRemoveTypeImports: true,
+            // babel equivalent of tsconfig `allowImportingTsExtension: true`
+            rewriteImportExtensions: true,
```
askoufis, line 85: We already enable `allowImportingTsExtensions`, so we should enable it in our babel config too, though it doesn't really affect output code since it's all bundled anyway.

.changeset/odd-mugs-enjoy.md
```diff
@@ -0,0 +1,22 @@
...
+
+```ts
+// Erased entirely
+import type { A } from "a";
+
+// Rewritten to `import { b } from "bcd";`
+import { b, type c, type d } from "bcd";
+
+// Rewritten to `import {} from "xyz";`
+import { type xyz } from "xyz";
```
michaeltaranto, line 17: This one surprised me, shouldnt it be erased entirely as well? Or is that incase there are side effect imports in `xyz`?

askoufis: Yep, it just strips the type import, leaving the side-effect. In practice though, our lint rules would convert this to `import type { xyz } from "xyz"`, so it actually _would_ be stripped entirely.

michaeltaranto (approved):

askoufis:
Turns out `rewriteImportExtensions: true` was the wrong move. It makes babel transform `.ts` imports into `.js`, but this isn't actually what we want because the imported file itself isn't renamed to `.js`, so the import fails.

PR #984, opened 2024-05-30
sku v13: Explicitly define entrypoints via `package.json#exports`

askoufis:
This enforces the newly documented [API entrypoints](https://seek-oss.github.io/sku/#/./docs/api).


michaeltaranto (approved):

packages/sku/scripts/test.js
```diff
@@ -7,17 +6,15 @@ const { argv } = require('../config/args');
 const { runVocabCompile } = require('../lib/runVocab');
 
 const log = debug('sku:jest');
-// https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#configuring-your-testing-environment
-process.env.IS_REACT_ACT_ENVIRONMENT = true;
```
askoufis, line 11: According to the linked docs, this never worked? It needs to be set on `globalThis`, not `process.env`, so I'm unsure what this was accomplishing, if anything.

PR #986, opened 2024-05-30
sku v13: Update `eslint` and `eslint-config-seek`, bump minimum node version

askoufis:
See changesets.


michaeltaranto (approved):

askoufis:
Babel is dropping support for node <18.20 in [their next major](https://github.com/babel/babel/releases/tag/v8.0.0-alpha.9), so maybe that should be our minimum version instead? Or at least, whatever the latest 18.x LTS version is at time of release.

askoufis:
Actually, going to pull out the TS 5.5 update so it's not gated behind a major version.

PR #990, opened 2024-06-04
Enable `babel-loader` cache, document webpack and `babel-loader` cache

askoufis:
From the changeset:

> Sku's webpack configuration now configues `babel-loader` to emit a cache to `node_modules/.cache/babel-loader`. The primary use case for this cache is speeding up production builds. It can also speed up local development in situations where the webpack cache is invalidated.

The emitted cache files take up a relatively small amount of space (10s of MBs, and the compress quite well), making it much more practical to cache on CI, relative to the webpack cache.

In a moderately-sized, internal monorepo, this cache dropped the `sku build` time in CI from an average of **4m 11s** (across 5 builds) to an average of **3m 3s** (across 5 builds). The cost of each cache file is roughly an extra 6s (saving and restoring the cache) and 8.3mb of storage.

Caching `node_modules/.cache/babel-loader` across all of sku's fixtures decreased test time by ~20s in sku's CI.

I've added docs for both this cache and the existing webpack cache in the `Extra features` docs. The docs are relatively high-level, and don't go into details on actually setting up a cache in CI.


.changeset/fast-ads-tease.md
```diff
@@ -0,0 +1,7 @@
+---
+'sku': patch
+---
+
+Enable `babel-loader` cache
+
+Sku's webpack configuration now configues `babel-loader` to emit a cache to `node_modules/.cache/babel-loader`. The primary use case for this cache is speeding up production builds. It can also speed up local development in situations where the webpack cache is invalidated.
```
michaeltaranto, line 7:
```suggestion
Sku's webpack configuration now configures `babel-loader` to emit a cache to `node_modules/.cache/babel-loader`. The primary use case for this cache is speeding up production builds. It can also speed up local development in situations where the webpack cache is invalidated.
```

docs/docs/extra-features.md
```diff
@@ -53,6 +53,34 @@ export default {
...
+
+[webpack filesystem cache]: https://webpack.js.org/configuration/cache/#cachetype
+[`persistentCache` configuration]: ./docs/configuration#persistentcache
+
+### [`babel-loader` cache]
+
+This cache stores the result of module transpilation performed by `babel-loader`.
+It is emitted during both local development and production builds.
+Its purpose is to speed up transpilation of TypeScript/JavaScript code.
+This can benefit both local develpoment (when the webpack cache is invalidated) and production builds.
```
michaeltaranto, line 77:
```suggestion
This can benefit both local development (when the webpack cache is invalidated) and production builds.
```

.github/workflows/validate.yml
```diff
@@ -52,6 +52,14 @@ jobs:
           path: ${{ steps.set-puppeteer-cache.outputs.PUPPETEER_CACHE_DIR }}
           key: puppeteer-${{ runner.os }}-${{ hashFiles('./pnpm-lock.yaml') }}
 
+      - name: Cache babel-loader
+        id: babel-loader-cache
+        uses: actions/cache@v3
+        with:
+          path: 'fixtures/*/node_modules/.cache/babel-loader'
+          key: babel-loader-${{ runner.os }}-${{ hashFiles('./pnpm-lock.yaml') }}
+          restore-keys: babel-loader-${{ runner.os }}-
```
michaeltaranto, line 61: Just checking, is the trailing "-" correct?

.github/workflows/validate.yml
```diff
@@ -52,6 +52,14 @@ jobs:
           path: ${{ steps.set-puppeteer-cache.outputs.PUPPETEER_CACHE_DIR }}
           key: puppeteer-${{ runner.os }}-${{ hashFiles('./pnpm-lock.yaml') }}
 
+      - name: Cache babel-loader
+        id: babel-loader-cache
+        uses: actions/cache@v3
```
michaeltaranto, line 57: Worth bumping to v4? Or maybe doing that across the board separately?

michaeltaranto (approved):

.github/workflows/validate.yml
askoufis: All [these examples](https://github.com/actions/cache/blob/0c45773b623bea8c8e75f6c82b208c3cf94ea4f9/examples.md#c---nuget) have a trailing hyphen, so I'd say yes.

It's probably not required, I think the restore key is just meant to be a non-unique prefix of the `key`, so it would probably work without the `-` too.

.github/workflows/validate.yml
askoufis: Was going to do it separately 👍

PR #996, opened 2024-06-11
sku v13: Remove storybook dependencies, CLI and automatic configuration

askoufis:
## Changes

Sku no longer provides the following Storybook-related features:
- Installation of Storybook dependencies
- Storybook CLI commands (`sku storybook` and `sku build-storybook`)
- Automatic configuration of `.storybook/main.ts`
- Re-exporting of storybook APIs under `sku/@storybook/react`

Users are now expected to install and configure Storybook themselves. `sku` provides some minimal configuration via the `sku/config/storybook` entrypoint.

## Reasoning

There's an internal document with a more thorough writeup, but a non-exhaustive summary would be:
- Decoupling the sku version from the storybook version allows consumers to upgrade when they're ready
- Storybook is a large dependency, and roughly 75% of sku consumers don't use it, so why should they install it?
- Sku's CLI wrappers weren't providing much value, and the automatic configuration was restrictive, and we can still provide useful config without being prescriptive
- We don't really have any opinions on how you should _use_ storybook, so why should we restrict how you configure it?

## Reviewer notes

A lot of changed files are just `*ignore` file updates.


tadhglewis:
@askoufis considering not many people use SB this makes sense. In my use case, it makes it easier than relying on Sku to expose the right functionality e.g. with the preview.tsx changes

Is it worth beta releasing this and I can test it on our panels repo? Or are you confident in the migration guides for things like the dev server middleware working nicely with sku?

tadhglewis (approved):

askoufis:
> @askoufis I reckon this makes sense and honestly makes it easier for consumers - there's not many SB users, exposing storybook within sku is annoying for updates (e.g. exposing preview.tsx config you had to change for our panels repo etc)

Have added a reasoning section to the PR just to call this out. Didn't want it to seem like a sudden decision, there was a decent amount of though put into this.

As per your original comment, would be happy to cut a snapshot release if you want to test it out.

askoufis:
@tadhglewis 

```json
"sku": "0.0.0-remove-storybook-deps-config-cli-20240617001049",
```

tadhglewis (commented):
If you're using MSW with Storybook, you now need to add `staticDirs: [{ from: '../public', to: '' }],` to the config as sku isn't serving the public dir

.changeset/curly-crews-wait.md
```diff
@@ -0,0 +1,38 @@
...
+// .storybook/main.ts
+import { babel, webpackFinal } from 'sku/config/storybook';
+import type { StorybookConfig } from '@storybook/react-webpack5';
+
+export default {
+  stories: ['../src/**/*.stories.tsx'],
+  framework: '@storybook/react-webpack5',
+  core: {
+    builder: {
+      name: '@storybook/builder-webpack5',
```
tadhglewis, line 25: Just small fyi it looks like when running `storybook upgrade` it performs a framework migration which moves the builder under the `framework` object... It seemed to work with either one... Is the migration a bad migration or should we be using that?

Screenshot 2024-06-26 at 10 24 25 AM

askoufis: This contradicts [the config structure in their docs](https://storybook.js.org/docs/builders/webpack#configure) 😡. I don't know who to believe.

Not sure why it's detecting a new framework migration though. Maybe it would be better if we just advised installing storybook 7 deps when doing the migration, then moving to storybook 8 separately, potentially via the CLI. Thoughts?

.changeset/curly-crews-wait.md
```diff
@@ -0,0 +1,38 @@
...
+Sku no longer configures your `.storybook/main.ts` file for you. If you are using Storybook, you will need to configure it yourself. See [`sku`'s Storybook documentation][storybook docs] for more information.
+
+**MIGRATION GUIDE**:
+
+Minimal Storybook configuration:
+
+```ts
+// .storybook/main.ts
+import { babel, webpackFinal } from 'sku/config/storybook';
+import type { StorybookConfig } from '@storybook/react-webpack5';
```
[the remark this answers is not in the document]
askoufis: It's mentioned in [a separate changeset](https://github.com/seek-oss/sku/pull/996/files/4c489207c3992f48b2fbf31a27574391c6682d30), though maybe it's worth combining everything into one big changeset 🤔.

.changeset/chilled-tips-clean.md
```diff
@@ -0,0 +1,29 @@
...
+```sh
+pnpm install -D storybook
+```
+
+```diff
+// package.json
+{
+  "scripts": {
+-    "storybook": "sku storybook",
++    "storybook": "storybook dev",
```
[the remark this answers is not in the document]
askoufis: Good catch. There _was_ a config for the port, but no longer will be. Will update with the default port.

.changeset/curly-crews-wait.md
askoufis: I do prefer this syntax though:

```
  framework: {
    name: '@storybook/react-webpack5',
    options: {
      builder: {
        fsCache: true,
      },
    },
  },
```

- No reference to `@storybook/builder-webpack5`
- No `core` property, so less nesting

Despite the differences between it and the docs, I'd rather go with the simpler config.

.changeset/dull-colts-fetch.md
```diff
@@ -0,0 +1,108 @@
...
+
+# (Optional) This addon is required for Storybook >=8.0.0
+pnpm install -D @storybook/addon-webpack5-compiler-babel
+```
+
+#### Configure Storybook
+
+Sku no longer configures your `.storybook/main.ts` file for you. You will now need to configure it yourself. See [`sku`'s Storybook documentation][storybook docs] for more information.
+
+`sku` now provides the `babel` and `webpackFinal` configuration objects under the `sku/config/storybook` entrypoint. These objects are best utilized within in order to  Here's an example of a minimal, `sku`-compatible Storybook configuration:
```
adenj, line 42:
```suggestion
`sku` now provides the `babel` and `webpackFinal` configuration objects under the `sku/config/storybook` entrypoint. These objects are best utilized within in order to  

Here's an example of a minimal, `sku`-compatible Storybook configuration:
```

felixhabib: "These objects are best utilized within in order to" think this sentence got cut off too? Also utilised?

.changeset/dull-colts-fetch.md
```diff
@@ -0,0 +1,137 @@
...
+
+**BREAKING CHANGE**
+
+All Storybook-related features have been removed from `sku`. Sku no longer installs Storybook dependencies, configures Storybook, provides Storybook CLI commands, or re-exports Storybook APIs. If you are using Storybook, you will need to install Storybook dependencies yourself. See [`sku`'s Storybook documentation][storybook docs] for more information.
+
+[storybook docs]: https://seek-oss.github.io/sku/#/./docs/storybook
+
+**MIGRATION GUIDE**:
+
+#### Update your `sku.config.js`
```
askoufis, line 15:
```suggestion
#### Update your `sku` config
```

.changeset/dull-colts-fetch.md
```diff
@@ -0,0 +1,137 @@
...
+- `storybookStoryStore`
+- `storybookTarget`
+
+Please remove them from your sku configuration file.
+
+Save the value of `storybookTarget` to use later in the migration.
+
+#### Update your ignore files
+
+Sku no longer manages certain storybook related ignorefile entries.
```
askoufis, line 30: We did some googling and it seems like `ignore files` is used more often than `ignorefiles`.

```suggestion
Sku no longer manages certain Storybook-related ignore file entries.
```

felixhabib (approved):

PR #997, opened 2024-06-13
Convert all `.less` styles in fixtures to Vanilla Extract styles, update tests, remove redundant fixture and tests

askoufis:
This is pre-work to removing support for `.less` styles in sku v13. To prevent a huge PR, I've split up feature removal (will be in a follow up PR) from the style conversion/test updates (this PR).

The changes should be fairly straightforward. Basically just converted all less styles in fixtures to VE styles, and updated tests accordingly. The `react-css-modules` fixture was doing a subset of the tests that the `typescript-css-modules` fixture already does, so I removed the entire fixture.

There's potentially some more redundant tests/fixtures we could clean up, and we should ideally rename `typescript-css-modules` to something like `typescript-vanilla-extract`, but I didn't want to bloat the PR too much, so I'll probably get to that stuff post-v13 release.


tadhglewis (approved):
LGTM from my skim through - lots of converting simple root styling to VE

kjots (approved):

PR #1003, opened 2024-06-26
Update TypeScript to 5.5

askoufis:
Splitting out this update from https://github.com/seek-oss/sku/pull/986 so it doesn't end up gated behind a major version.

I think the lockfile changes are coming from Braid's peer dependency on `sku`. It results in the latest version of sku being installed in `node_modules`, so we temporarily end up with two version of `typescript` installed, the new one and the old one. Once the next release is out this should go away.

### DIFF
--- a/.changeset/chilled-rocks-tap.md
+++ b/.changeset/chilled-rocks-tap.md
@@ -1,0 +1,10 @@
+---
+'sku': minor
+---
+
+Update TypeScript from 5.3 to 5.5
+
+Both the 5.4 and 5.5 releases includes breaking changes. See the [TypeScript 5.4 announcement] and [TypeScript 5.5 announcement] for more information.
+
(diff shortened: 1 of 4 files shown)


michaeltaranto (approved):

.changeset/chilled-rocks-tap.md
```diff
@@ -0,0 +1,10 @@
+---
+'sku': minor
+---
+
+Update TypeScript from 5.3 to 5.5
+
+Both the 5.4 and 5.5 releases includes breaking changes. See the [TypeScript 5.4 announcement] and [TypeScript 5.5 announcement] for more information.
```
askoufis, line 7:
```suggestion
Both the 5.4 and 5.5 releases include breaking changes. See the [TypeScript 5.4 announcement] and [TypeScript 5.5 announcement] for more information.
```

PR #1013, opened 2024-07-09
sku v13: remove `persistentCache` config option

felixhabib:
See changeset for details

### DIFF
--- a/.changeset/stupid-seals-think.md
+++ b/.changeset/stupid-seals-think.md
@@ -1,0 +1,24 @@
+---
+'sku': major
+---
+
+Remove support for `persistentCache` configuration option
+
+**BREAKING CHANGE:**
+
+Disabling `persistentCache` was previously necessary when using `treat`, which is no longer supported in `sku`. `persistentCache` is now always enabled has been removed from the configuration options.
+
+**MIGRATION GUIDE:**
+
+Remove the `persistentCache` option from your sku config.
+
+```diff
+// sku.config.ts
+
+import { type SkuConfig } from 'sku';
+
+export default {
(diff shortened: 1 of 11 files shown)


.changeset/stupid-seals-think.md
```diff
@@ -0,0 +1,24 @@
+---
+'sku': major
+---
+
+Remove support for `persistentCache` configuration option
+
+**BREAKING CHANGE:**
+
+Disabling `persistentCache` was previously necessary when using `treat`, which is no longer supported in `sku`. `persistentCache` is now always enabled has been removed from the configuration options.
```
michaeltaranto, line 9: Maybe make `treat` a link to the site for those who dont know what it is.

```suggestion
Disabling `persistentCache` was previously necessary when using `treat`, which is no longer supported in `sku`. `persistentCache` is now always enabled and has been removed from the configuration options.
```

.changeset/stupid-seals-think.md
```diff
@@ -0,0 +1,26 @@
+---
+'sku': major
+---
+
+Remove support for `persistentCache` configuration option
+
+**BREAKING CHANGE:**
+
+Disabling `persistentCache` was previously necessary when using [treat], which is no longer supported in `sku`. `persistentCache` is now always enabled has been removed from the configuration options.
```
askoufis, line 9:
```suggestion
Disabling `persistentCache` was previously necessary when using [treat], which is no longer supported in `sku`. `persistentCache` is now always enabled and has been removed from the configuration options.
```

askoufis (approved):

PR #1015, opened 2024-07-10
Update tsconfig comment and changeset

askoufis:
It turns out that `verbatimModuleSyntax: true` enables `tsc` to emit errors when type-only imports are missing. This is only really an issue if a sku app is depending on a `compilePackage` that ships raw `.ts` files that do not adhere to our ESLint rules that enforce correct type import syntax. Only 1 such case (an internal package) has been identified internally, and will be fixed very soon.

This PR updates the changeset and comment relevant to the `verbatimModuleSyntax: true` change from #983 to be a bit clearer about what the option does and to mention that TypeScript errors _may_ occur. However, I'm confident that most consumers won't actually run into these errors at all, unless they depend on the package mentioned above.

### DIFF
--- a/.changeset/odd-mugs-enjoy.md
+++ b/.changeset/odd-mugs-enjoy.md
@@ -17,6 +17,6 @@ import { b, type c, type d } from "bcd";
 import { type xyz } from "xyz";
 ```
 
-This change is not expected to have an affect on bundled application code or library code. However, it may surface some TypeScript errors in `compilePackage` dependencies that do not adhere to the correct type-only import syntax. These errors should be fixed in the dependency's codebase.
+This change is not expected to have an effect on bundled application code or library code. However, it may surface some TypeScript errors in `compilePackage` dependencies that do not adhere to the correct type-only import syntax. These errors should be fixed in the dependency's codebase.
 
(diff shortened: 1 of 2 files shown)


.changeset/odd-mugs-enjoy.md
```diff
@@ -17,6 +17,6 @@ import { b, type c, type d } from "bcd";
 import { type xyz } from "xyz";
 ```
 
-Since bundled code should not contain `import` statements, this change should not affect the output of most app code. Additionally, since our ESLint config ensures that `import { type xyz }` is transformed into `import type { xyz }`, this change is unlikely to affect library code either.
+This change is not expected to have an affect on bundled application code or library code. However, it may surface some TypeScript errors in `compilePackage` dependencies that do not adhere to the correct type-only import syntax. These errors should be fixed in the dependency's codebase.
```
felixhabib, line 20:
```suggestion
This change is not expected to have an effect on bundled application code or library code. However, it may surface some TypeScript errors in `compilePackage` dependencies that do not adhere to the correct type-only import syntax. These errors should be fixed in the dependency's codebase.
```

felixhabib (approved):

PR #1018, opened 2024-07-11
Update `@swc/core` and `@vocab/phrase`, loosen `esbuild` dependency version

askoufis:
- `@swc/core` has fixed some minification bugs recently that an internal consumer ran into. Bumping the version to the latest release so other consumers get the update when moving to sku v13.
- `@vocab/phrase` had a bug which as now been fixed, updating to [the version with the fix](https://github.com/seek-oss/vocab/releases/tag/%40vocab%2Fphrase%402.0.1).
- There are multiple `esbuild` versions ending up in `node_modules`. We can afford to be much looser with our `esbuild` dep version because we don't depend on any particular functionality other than traspilation via `esbulid-register`. This change lets us resolve to higher 0.x versions, instead of being stuck to 0.19.


.changeset/brave-balloons-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+Loosen `esbuild` dependency version to mitigate multiple versions of it ending up in `node_modules`
```
felixhabib, line 5:
```suggestion
Loosen `esbuild` dependency version to help prevent multiple versions of it ending up in `node_modules`
```

Nitpick, whatever you prefer

felixhabib (approved):

askoufis:
Snuck in a vocab dep update too.

PR #1021, opened 2024-07-16
Exclude pnpm-lock.yaml from linting 

AaronMoat:
`prettier` causes mayhem with `pnpm-lock.yaml` and this reared its head a bit more by the recent changes not to use an extensions list, I think.

### DIFF
--- a/fixtures/translations/.prettierignore
+++ b/fixtures/translations/.prettierignore
@@ -6,7 +6,7 @@ translations.ts
 .eslintrc
 .prettierrc
 coverage/
-dist/
+dist-ssr/
 report/
 tsconfig.json
 # end managed by sku
--- a/fixtures/translations/.prettierignore
+++ b/fixtures/translations/.prettierignore
@@ -6,7 +6,8 @@ translations.ts
 .eslintrc
 .prettierrc
 coverage/
-dist-ssr/
+dist/
 report/
 tsconfig.json
+pnpm-lock.yaml
 # end managed by sku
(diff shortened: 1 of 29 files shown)


fixtures/translations/.prettierignore
```diff
@@ -6,7 +6,7 @@ translations.ts
 .eslintrc
 .prettierrc
 coverage/
-dist/
+dist-ssr/
 report/
 tsconfig.json
```
AaronMoat, line 11:
```suggestion
dist/
report/
tsconfig.json
pnpm-lock.yaml
```

packages/sku/lib/configure.js
```diff
@@ -84,7 +84,9 @@ module.exports = async () => {
   await ensureGitignore({
     filepath: getPathFromCwd('.prettierignore'),
     comment: 'managed by sku',
-    patterns: lintIgnorePatterns.map(convertToForwardSlashPaths),
+    patterns: lintIgnorePatterns
+      .concat('pnpm-lock.yaml')
+      .map(convertToForwardSlashPaths),
```
michaeltaranto, line 89: Feel like `pnpm-lock.yaml` should be considered earlier in here so we dont even lint it (rather than focusing on prettier).

Maybe shifting it up to the `lintIgnorePatterns` array?

michaeltaranto (commented):
In terms of package managers, I believe we still have enough `yarn` adoption that would justify including `yarn.lock` in this change.

Would you mind extending this PR to include it?

AaronMoat:
Oh interesting, I think the reason pnpm gets caught up is the .yaml file extension; I think tooling like eslint and prettier don't know about .lock files and ignore them natively. Do you think it's worth including just in case?

michaeltaranto:
Ah ok, sorry I missed that nuance. Maybe we stick with just pnpm then.

michaeltaranto (approved):

PR #1024, opened 2024-07-22
Better jest preset

askoufis:
I'm considering this a fast-follow to the jest preset work done in sku 13, and since it's a barely-used feature, I'm not marking it as breaking, even though it is. I've updated the v13 changelog and docs accordingly.

Before the v13 release, I tried adding a `jest-preset` entrypoint to sku's `package.json`, but it didn't work, so I went with a relative path exported as a constant instead. It turns out that, if jest actually surfaced the relevant errors from [`resolve.exports`](https://www.npmjs.com/package/resolve.exports) (the package used by jest to resolve package exports), the fix would've been trivial and a `jest-preset` entrypoint would've worked just fine.

Jest uses [`resolve.exports`](https://www.npmjs.com/package/resolve.exports) to resolve package.json exports. It turns out `resolve.exports` doesn't like this entrypoint, causing it to fail, so preset module resolution fails when trying to resolve a `sku` preset.
```json
 ".": {
      "types": "./sku-types.d.ts"
    },
```

Simplifying it to this allows exports to be resolved, and then `sku` works as a jest preset:
```json
 ".": "./sku-types.d.ts"
```

This change doesn't affect the `sku` entrypoint at all, it still just exports types. The primary benefit of this change is that `jest` can now resolve `sku`'s jest preset even if it's hoisted to higher `node_modules` directory, which is a nice to have in certain monorepo situations. This was not possible using a relative path.

This has been tested in two different internal projects.

### DIFF
--- a/.changeset/pink-carpets-return.md
+++ b/.changeset/pink-carpets-return.md
@@ -1,0 +1,27 @@
+---
+'sku': patch
+---
+
+Replace `sku/config/jest` entrypoint with `sku/jest-preset`
+
+**BREAKING CHANGE**:
+
+This breaking change will only affect a very small number of projects that have a custom Jest configuration/wish to debug tests in their IDE, and it is intended to be a quick fix/improvement for a feature released in `sku@13.0.0`. Therefore, it's not being released as a major version.
+
+The `sku/config/jest` entrypoint has been removed in favor of a new `sku/jest-preset` entrypoint. The `sku/jest-preset` module is a better way to expose a [jest preset], rather than a relative path, as it works even if `sku` is hoisted to a parent `node_modules` directory.
+
+**MIGRATION GUIDE**:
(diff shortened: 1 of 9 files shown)


.changeset/pink-carpets-return.md
```diff
@@ -0,0 +1,27 @@
+---
+'sku': patch
+---
+
+Replace `sku/config/jest` entrypoint with `sku/jest-preset`
+
+**BREAKING CHANGE**:
```
michaeltaranto, line 7: Maybe scope this message to the API which makes sense why its not considered a formal break:

```suggestion
**BREAKING CHANGE FOR `dangerouslySetJestConfig`**:
```

Thoughts?

michaeltaranto (approved):

PR #1028, opened 2024-08-01
Use Relative path for sku Config

jahredhope:
Fixes a situation when an incorrect path would be used for ignore files.

Previously you might get a list with whatever the users current directory was, minus the app name, with 'dist' on the end:

**.gitignore**
```
.eslintcache
.eslintrc
.prettierrc
/users/current/dir/dist/
coverage/
report/
tsconfig.json
```

Now this value will just be `dist/`, as it would be if running `sku configure`.

More information in changeset.

---

Changed `sku-init` test to also check generated files so we can potentially catch this in the future, and make changes to init steps with more confidence.

I needed to special-case package.json as we don't want to update the snapshot when there are new dependency versions.

Whilst jest does support [Property Matchers](https://jestjs.io/docs/snapshot-testing#property-matchers), I chose to implement this by just replacing the version as it's easier to still test WHICH dependencies were installed (regardless of the version), and IMO it's a cleaner snapshot.

### DIFF
--- a/fixtures/sku-init/sku-init.test.js
+++ b/fixtures/sku-init/sku-init.test.js
@@ -9,13 +9,37 @@ const projectName = 'new-project';
 const projectDirectory = path.join(fixtureDirectory, projectName);
 
 describe('sku init', () => {
-  beforeAll(async () => {
-    await fs.rm(projectDirectory, { recursive: true, force: true });
-  });
+  let child;
+  let stdout;
+  let stderr;
+
+  beforeAll(
+    async () => {
+      await fs.rm(projectDirectory, { recursive: true, force: true });
+
+      const projectName = 'new-project';
+      await fs.rm(path.join(fixtureDirectory, projectName), {
+        recursive: true,
+        force: true,
+      });
+
+      ({ child, stdout, stderr } = await runSkuScriptInDir(
+        'init',
+        fixtureDirectory,
+        [projectName],
+      ));
+
+      console.log('sku init stdout');
+      console.log(stdout);
+      console.log('sku init stderr');
+      console.error(stderr);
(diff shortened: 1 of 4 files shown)


askoufis (commented):
Can we add a test for this in https://github.com/seek-oss/sku/blob/48ae90b810ff1b3fe9cdb18eb90d331d64539403/fixtures/sku-init/sku-init.test.js?

jahredhope:
> Can we add a snapshot of `.gitignore` in https://github.com/seek-oss/sku/blob/48ae90b810ff1b3fe9cdb18eb90d331d64539403/fixtures/sku-init/sku-init.test.js?

Yes good idea. But I'd prefer to go further and also snapshot package.json. Will update PR soon.

fixtures/sku-init/sku-init.test.js
```diff
@@ -9,62 +9,84 @@ const projectName = 'new-project';
...
-      );
+  it('should create package.json', async () => {
+    const contents = await fs.readFile(
+      path.join(fixtureDirectory, projectName, 'package.json'),
+      'utf-8',
+    );
 
-      expect(skuConfig).toMatchInlineSnapshot(`
-"import type { SkuConfig } from 'sku';
+    expect(replaceDependencyVersions(JSON.parse(contents))).toMatchSnapshot();
```
jahredhope, line 60: Whilst jest does support [Property Matchers](https://jestjs.io/docs/snapshot-testing#property-matchers), I chose to implement this by just replacing the version as it's easier to still test WHICH dependencies were installed (regardless of the version), and IMO it's a cleaner snapshot.

askoufis (approved):

PR #1036, opened 2024-08-19
`start` script: reuse tab if possible in more Chromium browsers

felixhabib:
`sku` will reuse an existing tab if available in more Chromium browsers.

If the user's default browser is not Chromium based, there will be no change. If it is Chromium based, and the user has multiple supported Chromium browsers installed, the tab will open (or be reused) in:

* The user's default browser, if it is open, or if no browsers are open.
* Another supported browser, if it is open and the default browser is not open.

Using [default-browser](https://www.npmjs.com/package/default-browser) package instead of [x-default-browser](https://www.npmjs.com/package/x-default-browser) as it has support for Microsoft Edge and is more widely used.

NB:

This PR is heavily based on the changes from [this PR](https://github.com/facebook/create-react-app/pull/8367), referenced [here](https://github.com/seek-oss/sku/compare/open-edge-one-tab?expand=1#diff-b07bf4a632a44c85e64c21303e58f117e276133822fc8d1a0769e5fc2542d1daL2).
It is an update to https://github.com/seek-oss/sku/pull/426.

### DIFF
--- a/.changeset/new-eggs-admire.md
+++ b/.changeset/new-eggs-admire.md
@@ -22,5 +22,3 @@ A tab will be reused if:
 * The user has an existing tab open in a supported Chromium browser with the exact same URL.
 
 If any of the above is not true, a new tab will be created in the user's default browser.
-
-The supported Chromium browsers are:
--- a/.changeset/new-eggs-admire.md
+++ b/.changeset/new-eggs-admire.md
@@ -1,0 +1,26 @@
+---
+'sku': minor
+---
+
+Widen support for reusing existing browser tab to more Chromium browsers.
+
+`start` and `start-ssr` scripts would previously only reuse an existing tab in Google Chrome.
+This change adds support for the following Chromium browsers:
+
+- Google Chrome,
+- Google Chrome Canary,
+- Microsoft Edge,
+- Brave Browser,
+- Vivaldi,
+- Chromium,
+- Arc.
+
+A tab will be reused if:
+
+* The OS is macOS,
+* The user's default browser is a supported Chromium browser,
+* The user has an existing tab open in a supported Chromium browser with the exact same URL.
+
+If any of the above is not true, a new tab will be created in the user's default browser.
+
+The supported Chromium browsers are:
--- a/packages/sku/lib/openBrowser/index.js
+++ b/packages/sku/lib/openBrowser/index.js
@@ -2,32 +2,35 @@
 // https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/openBrowser.js
 
 const execSync = require('node:child_process').execSync;
-const defaultBrowser = require('x-default-browser');
 const open = require('open');
 
 const isCI = require('../isCI');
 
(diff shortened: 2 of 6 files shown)


packages/sku/lib/openBrowser/index.js
```diff
@@ -2,32 +2,61 @@
...
 
 const isCI = require('../isCI');
 
-module.exports = (url) => {
+const OSX_CHROME = 'google chrome';
+
+module.exports = async (url) => {
   if (process.env.OPEN_TAB !== 'false' && !isCI) {
-    defaultBrowser((err, res) => {
-      const useChrome = err ? false : res.isChrome;
```
askoufis, line 13: I'm experiencing incorrect behaviour when I have both Firefox (my default) and Chrome open (no existing tab). I'd expect Firefox to get the new tab, but instead Chrome gets it.

I think the logic needs to be something like this:
1. Check if default browser is one of the supported chromium browsers.
2. If it is, do the new logic you've implemented
3. If it's not, check if the default browser is currently open with:
```
execSync(`ps cax | grep "${defaultBrowser}"`);
```
4. If that fails, it'll hit a `catch`, at which point you'll want to try doing the Chromium stuff.
5. If _that_ fails, it should fall through to `open`

This might require some abstracting of chromium browser check and some loop shenanigans.

felixhabib: Ahh yes, I have not been testing with non-Chromium. Will fix this

packages/sku/lib/openBrowser/index.js
```diff
@@ -2,32 +2,63 @@
 // https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/openBrowser.js
```
askoufis, line 2: Worth updating this comment to point to the actual commit rather than just `master`.

askoufis (commented):
Tested on my machine, works great! Just 1 comment, also the PR needs a changeset.

.changeset/new-eggs-admire.md
```diff
@@ -0,0 +1,23 @@
+---
+'sku': minor
+---
+
+Update `start` and `start-ssr` scripts to reuse an existing browser tab if available.
```
askoufis, line 5: Technically this feature already exists, but we're just extending it to support more browsers.

.changeset/new-eggs-admire.md
```diff
@@ -0,0 +1,26 @@
...
+
+A tab will be reused if:
+
+* The OS is macOS,
+* The user's default browser is a supported Chromium browser,
+* The user has an existing tab open in a supported Chromium browser with the exact same URL.
+
+If any of the above is not true, a new tab will be created in the user's default browser.
+
+The supported Chromium browsers are:
```
askoufis, line 26:
```suggestion
```

askoufis (approved):

.changeset/new-eggs-admire.md
felixhabib: 😅

PR #1039, opened 2024-08-26
Use HTTPS URL for `zendesk/babel-plugin-react-displayname` dependency

askoufis:
It has been observed internally that PNPM will sometimes replace the artifact URL for a github dependency with an SSH URL.

The only relevant PNPM issue I could find was https://github.com/pnpm/pnpm/issues/4527. I'm not sure if this fix on `sku`'s end will actually prevent it from happening again because the resolution of the dependency is in PNPM's hands, but it's worth a shot.

### DIFF
--- a/.changeset/famous-spiders-thank.md
+++ b/.changeset/famous-spiders-thank.md
@@ -1,0 +1,5 @@
+---
+'sku': patch
+---
+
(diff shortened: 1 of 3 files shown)


michaeltaranto (approved):
🤞

PR #1042, opened 2024-08-28
Add `@ts-check` to some more files

askoufis:
Added `@ts-check` annotations to a few files where it either required no other changes or minimal type tweaks in other files.

### DIFF
--- a/packages/sku/lib/configure.js
+++ b/packages/sku/lib/configure.js
@@ -1,5 +1,3 @@
-#!/usr/bin/env node
-
 const { writeFile, rm } = require('node:fs/promises');
 const path = require('node:path');
 
--- a/packages/sku/lib/packageManager.js
+++ b/packages/sku/lib/packageManager.js
@@ -16,6 +16,24 @@ const skuArgs = require('../config/args');
 
 /** @typedef {'yarn' | 'pnpm' | 'npm'} SupportedPackageManager */
 
+const supportedPackageManagers = ['yarn', 'pnpm', 'npm'];
+
+/**
+ * @param {string} packageManager
+ * @returns {SupportedPackageManager}
+ */
+const validatePackageManager = (packageManager) => {
+  if (!supportedPackageManagers.includes(packageManager)) {
+    throw new Error(
+      `Unsupported package manager: ${packageManager}. Supported package managers are: ${supportedPackageManagers.join(
+        ', ',
+      )}`,
+    );
+  }
+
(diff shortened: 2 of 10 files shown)


packages/sku/lib/configure.js
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env node
```
askoufis, line 1: `scripts/configure.js` already has a shebang, this is a library module and doesn't need it.

packages/sku/lib/packageManager.js
```diff
@@ -16,6 +16,24 @@ const skuArgs = require('../config/args');
 
 /** @typedef {'yarn' | 'pnpm' | 'npm'} SupportedPackageManager */
 
+const supportedPackageManagers = ['yarn', 'pnpm', 'npm'];
+
+/**
+ * @param {string} packageManager
+ * @returns {SupportedPackageManager}
+ */
+const validatePackageManager = (packageManager) => {
```
askoufis, line 25: This is technically a fix because we didn't have any validation before, but IMO not worth mentioning in the patch notes.

packages/sku/lib/runPrettier.js
```diff
@@ -36,22 +37,19 @@ const runPrettier = async ({ write, listDifferent, paths }) => {
   const pathsToCheck = paths && paths.length > 0 ? paths : ['.'];
 
   prettierArgs.push(...pathsToCheck);
-  /*
-   * Show Prettier output with stdio: inherit
-   * The child process will use the parent process's stdin/stdout/stderr
-   * See https://nodejs.org/api/child_process.html#child_process_options_stdio
-   */
-  const processOptions = {
```
askoufis, line 44: Chose to just inline this object because it removes the need to the JSDoc equivalent of `as const`.

packages/sku/package.json
```diff
@@ -136,14 +137,14 @@
...
-    "wrap-ansi": "^7.0.0",
-    "default-browser": "^5.2.1"
+    "wrap-ansi": "^7.0.0"
   },
   "devDependencies": {
     "@types/cross-spawn": "^6.0.3",
     "@types/debug": "^4.1.12",
     "@types/eslint": "^8.56.10",
     "@types/minimist": "^1.2.5",
+    "@types/node": "^18.19.31",
```
askoufis, line 147: This is already a monorepo root dependency, but I figured it's worth being thorough about which types the `sku` CLI uses.

felixhabib (approved):

PR #1051, opened 2024-09-16
Update `didyoumean2` dep to fix `npx sku init`

askoufis:
Running `npx sku init my-app` fails on `node@18.20.1/npm@10.5.0` with no error message. Adding `--loglevel=verbose` surfaced this log line:

```
npm info run didyoumean2@6.0.1 postinstall { code: 127, signal: null }
```

This lead me to believe that the cause of the failure was probably the issue that Karl raised a while ago https://github.com/foray1010/didyoumean2/issues/1531. Figured the fix would eventually come through via renovate, but that hasn't happened for some reason, so I'm updating it manually.

### DIFF
--- a/.changeset/pretty-meals-tie.md
+++ b/.changeset/pretty-meals-tie.md
@@ -1,0 +1,5 @@
+---
+'sku': patch
+---
+
+Update `didyoumean2` dependency to fix issue when running `npx sku init`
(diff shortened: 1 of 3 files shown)


jahredhope (approved):

felixhabib (approved):
nice one!

PR #1055, opened 2024-09-16
Skip running tests on version packages PRs

askoufis:
Small optimization that means releases will go out faster.


askoufis:
Actually nah, they're already fast enough, and I've seen the `skip concurrent runs` action sometimes skip when it shouldn't have, so this is a the last barrier.

PR #1061, opened 2024-09-30
`start` script: Add fallback browser names to supported browser list

felixhabib:
Fix bug from https://github.com/seek-oss/sku/pull/1036

Looks like certain instances of Chrome use "Google Chrome" and others "Chrome" as their respective names. "Chrome" was not in the supported browser list and was not being picked up correctly.
Adds both names for fallback purposes to ensure the feature works in more versions. Also giving "Edge" the same treatment.

Have tested locally and appears to be working

### DIFF
--- a/.changeset/nine-penguins-attend.md
+++ b/.changeset/nine-penguins-attend.md
@@ -1,0 +1,5 @@
+---
+'sku': patch
+---
+
+Adds "Chrome" and "Edge" as fallback browser names for the reusing existing tabs, improving compatibility with different Chromium browser versions which may use abbreviated browser names.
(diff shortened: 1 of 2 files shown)


.changeset/nine-penguins-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+Adds "Chrome" and "Edge" as fallback browser names for the reusing existing tabs, improving compatibility with different Chromium browser versions which may use abbreviated browser names.
```
askoufis, line 5:
```suggestion
Adds "Chrome" and "Edge" as fallback browser names for reusing existing tabs, improving compatibility with different Chromium browser versions which may use abbreviated browser names.
```

askoufis (approved):

PR #1063, opened 2024-10-03
Replace some dev deps with native alternatives

askoufis:
- Replace `node-fetch` with native `fetch`
- Replace `strip-ansi` with native [`stripVTControlCharacters`](https://nodejs.org/api/util.html#utilstripvtcontrolcharactersstr)


jahredhope (approved):

adenj (approved):
👏

PR #1069, opened 2024-10-24
fix error when determining default browser in `start` script

felixhabib:
Improve error handling and messaging in `start` script.

This fixes a macOS error when the user's terminal application's Finder automation permissions were disabled, which caused an error when trying to detect the user's default browser.
This change provides instructions to resolve the system permission issue and prevents script failure.

### DIFF
--- a/.changeset/three-bugs-sip.md
+++ b/.changeset/three-bugs-sip.md
@@ -4,5 +4,5 @@
 
 Improve error handling and messaging when opening the user's browser during the `start`, `start-ssr` and `serve` scripts
 
-This fixes a macOS error when the user's terminal application's Finder automation permissions were disabled, which caused an error when trying to detect the user's default browser.
-This change provides instructions to resolve the system permission issue and prevents script failure.
+This fixes a macOS error where `sku` would crash when failing to detect the user's default browser. 
+Instead, the user will be shown a warning message with instructions to enable the system permission required to fix the issue.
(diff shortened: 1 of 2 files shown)


.changeset/three-bugs-sip.md
```diff
@@ -0,0 +1,8 @@
+---
+'sku': patch
+---
+
+Improve error handling and messaging in `start` script.
```
askoufis, line 5:
```suggestion
Improve error handling and messaging when opening the user's browser during the `start`, `start-ssr` and `serve` scripts
```

.changeset/three-bugs-sip.md
```diff
@@ -0,0 +1,8 @@
...
+
+This fixes a macOS error when the user's terminal application's Finder automation permissions were disabled, which caused an error when trying to detect the user's default browser.
+This change provides instructions to resolve the system permission issue and prevents script failure.
```
askoufis, line 8: IMO the first sentence is a bit heavy on the details and could probably just be dropped. I think rewording the second sentence better communicates the issue and the change.
```suggestion
Failure to detect the user's default browser should no longer cause `sku` to crash when it tries to open a browser window.
Instead the user will be shown a warning and provided with instructions to enable the system permission required to fix the warning.
```

felixhabib: Good point, I agree. Riffed on it slightly
```suggestion
This fixes a macOS error where `sku` would crash when failing to detect the user's default browser. 
Instead, the user will be shown a warning message with instructions to enable the system permission required to fix the issue.
```

askoufis (approved):

PR #1075, opened 2024-10-31
Improve logging for `sku translations` CLI commands

askoufis:
Although the `sku translations` script itself is in an async IIFE, the calls to vocab's APIs were not awaited, so logs weren't exactly matching up with what was being executed, at least timing-wise. This adds `await`s where necessary to improve this.

Additionally, log messages have been improved a bit.

On a side note, during this change I realized that `sku translations validate` does not fail if it finds missing translations, whereas `vocab validate` does. Given that the command is undocumented, I reckon we could just update this behaviour and then document it without a breaking change.


packages/sku/scripts/translations.js
```diff
@@ -32,52 +32,70 @@ const ensureBranch = () => {
...
   try {
     if (translationSubCommand === 'compile') {
-      console.log('Watching for changes to translations');
-      compile({ watch }, vocabConfig);
+      log('Compiling translations...');
+
+      if (watch) {
+        log('Watching for changes to translations');
+        // Don't await so it can run in the background
+        compile({ watch }, vocabConfig);
```
kjots, line 63: Am I correct that not using `await` here will prevent the `catch` block below from catching any exceptions thrown here and thus ensure the process does not exit on error? 

If that is the case it might be better to include this detail in the comment as "run in the background" is a bit vague.

kjots (approved):

packages/sku/scripts/translations.js
askoufis: https://github.com/seek-oss/sku/pull/1075/commits/a7392fee9c54f1865f7d7ea25588f54b99b8ce6a

PR #1076, opened 2024-11-07
Set `esbuild-register` target to `node18.20`, align targets with supported node version

askoufis:
Sku [targets node 18.20 at a mininum](https://github.com/seek-oss/sku/blob/c210b789c1502c9af89490e4b382c7de69e6a7dd/packages/sku/package.json#L22), so `esbuild-register` should target that version too.

Have updated the repo's `postinstall` script to generate `targets.json` with a `browserslistNodeTarget` and `esbuildNodeTarget`.


package.json
```diff
@@ -48,6 +48,7 @@
     "@tsconfig/node-lts": "^18.12.3",
     "@types/jest": "^29.0.0",
     "@types/node": "^18.19.31",
+    "@types/prettier": "^2.7.3",
```
askoufis, line 51: `prettier` is used in `postinstall.js`.

PR #1078, opened 2024-11-08
Update `browserslist-config-seek` to version 3.1.0, and pin version

felixhabib:
Making this change manually to pin the dependency to the specific version. This means we can tie browser policy updates to sku versions in the future.

### DIFF
--- a/.changeset/nice-brooms-whisper.md
+++ b/.changeset/nice-brooms-whisper.md
@@ -2,4 +2,4 @@
 'sku': minor
 ---
 
-Update `browserslist-config-seek` to version 3.1.0.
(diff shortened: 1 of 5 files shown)


.changeset/nice-brooms-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': minor
+---
+
+Update `browserslist-config-seek` to version 3.1.0.
```
askoufis, line 5:
```suggestion
Update and pin `browserslist-config-seek` to version `3.1.0`
```

askoufis (approved):
LGTM, but some snapshots need updating.

PR #1081, opened 2024-11-10
Improve update snapshot flow

felixhabib:
Adding `test:update` script to run both sku contract updates and updates from within the `braid-design-system` fixture (the only relevant fixture that needs updates).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ sudo pnpm setup-test-hosts
 To run the test suite locally:
 
 ```sh
-pnpm run test:update
+pnpm run test
 ```
 
 > [!NOTE]
@@ -62,7 +62,7 @@ pnpm run test:update
 If snapshots are out of date, you can update them with:
 
 ```sh
-pnpm run test -u
+pnpm run test:update
 ```
 
 The test suite needs to pass for your changes to be accepted, so it's worth running this locally during development and right before committing.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ sudo pnpm setup-test-hosts
 To run the test suite locally:
 
 ```sh
-pnpm run test
+pnpm run test:update
 ```
 
 > [!NOTE]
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -65,25 +65,7 @@ If snapshots are out of date, you can update them with:
 pnpm run test -u
```
askoufis, line 65: Shouldn't this line be `pnpm run test:update` instead?

CONTRIBUTING.md
```diff
@@ -65,25 +65,7 @@ If snapshots are out of date, you can update them with:
...
-pnpm exec sku test -u
-```
-
-You can also run any `sku` CLI command against any of the app fixtures.
-This can be a faster way to iterate on a feature than rather than running the test suite every time you make a change.
-
-1. `cd` into the fixture you want to test. E.g. `cd fixtures/styling`.
-1. Run your sku command. E.g. `pnpm run sku build`.
-
-Once you've made the desired changes and you're ready to commit, stage your local changes.
```
askoufis, line 86: This stuff is separate to tests and still relevant.

package.json
```diff
@@ -11,6 +11,7 @@
     "check": "pnpm install --frozen-lockfile && echo 'Ignore paths from lint-staged'",
     "test": "SKU_TELEMETRY=false SKU_DISABLE_CACHE=true OPEN_TAB=false jest --verbose",
     "test:sku-init": "SKU_TELEMETRY=false SKU_DISABLE_CACHE=true OPEN_TAB=false jest --verbose --config=jest.config.sku-init.js",
+    "test:update": "pnpm run test:sku-init && pnpm --filter @sku-fixtures/braid-design-system exec sku test -u",
```
askoufis, line 14: I would leave `sku-init` out of this one as it's best to just run that on CI. Also this isn't updating the `test` script snapshots`.
```suggestion
    "test:update": "pnpm run test -u && pnpm --filter @sku-fixtures/braid-design-system exec sku test -u",
```

CONTRIBUTING.md
```diff
@@ -51,7 +51,7 @@ sudo pnpm setup-test-hosts
 To run the test suite locally:
 
 ```sh
-pnpm run test
+pnpm run test:update
```
askoufis, line 54: This should be unchanged.

askoufis (approved):

PR #1092, opened 2024-11-22
Update fixtures Braid Version, Theme, and sku-init Template Enhancements

adenj:
- **Bump Braid version:** Update fixtures to use 33.1.0.
- **Switch theme:** Replace `apac` theme with `seekJobs`.
- **Update sku-init template:** Add prompts for installing `sku-telemetry` and `shared-web-assets`.
- **Refresh snapshots:** Update to reflect changes.


packages/sku/template/src/App/NextSteps.tsx
```diff
@@ -74,9 +72,6 @@ export function NextSteps({ environment }: NextStepsProps) {
                     </TextLink>
                     , <TextLink href="https://jestjs.io/">Jest</TextLink>,{' '}
                     <TextLink href="https://eslint.org/">ESLint</TextLink>,{' '}
```
askoufis, line 74: I think there's a double comma here now that it gets combined with the `, and more.` line below.

askoufis (approved):
LGTM, just 1 more comment on the template.

packages/sku/template/src/App/NextSteps.tsx
adenj: Nice catch

PR #1094, opened 2024-11-25
Update ESLint to v9, migrate to flat config, automatically migrate eslint ignore entries during `sku configure`

askoufis:
> [!CAUTION]
> This is a breaking change and should not be merged until other breaking changes have been reviewed and it has been agreed that a new major version should be released. Or maybe we'll do a feature branch, I dunno yet.

## Reviewer notes

Lockfile changes, snapshot updates and generated file updates (anything inside `/fixtures`) are the bulk of the changes.

### Flat config

[ESLint v9 is now configured via an `eslint.config.js` file](https://eslint.org/docs/latest/use/migrate-to-9.0.0#-new-default-config-format-eslintconfigjs), so naturally sku now emits this file rather than the old `.eslintrc`. Additionally, sku's custom ESLint config (import ordering and ignores) has been migrated to the new flat config format.

`eslint-config-seek` migrated to flag config [in v14 a few months](https://github.com/seek-oss/eslint-config-seek/releases/tag/v14.0.0), so sku can now depend on `eslint-config-seek@^14`.

The way configs are extended has also changed. Rather than specifying `extends: ['seek']` and letting ESLint resolve the config, config is now imported/exported explicitly. This means we can just re-export the sku eslint config which includes any user ignores and/or dangerously set changes. As such, the sku eslint config is now exported from the `sku/config/eslint` entrypoint.

### Ignores
`.eslintignore` is no longer used as of ESLint v9. Ignores are now specified as a regular config object within a flat config file ([docs](https://eslint.org/docs/latest/use/configure/ignore)).

It was decided that we didn't want to force consumers to use `dangerouslySetEslintConfig` just to customize eslint ignores, so a new `eslintIgnore` configuration option has been added to the sku config. It accepts an array of glob patterns to ignore.

### Migration

To facilitate a smooth migration, `sku configure` (run before most `sku` commands and on `postinstall`) will attempt to migrate any custom (not already ignored by sku) ignores from `.eslintignore` to the new `eslintIgnore` sku config option.

## Other changes

Some other changes in this PR:
- Generated files within fixtures have been updated. For most fixtures this involves updating `.prettierignore` and `.gitignore`, deleting `.eslintignore` and adding `eslint.config.js`. A few fixtures may have their sku configs updated too.
- `link-workspace-packages="true"` has been set in `.npmrc`. This results in the workspace version of `sku` being installed in the `sku-init` test, instead of the latest published version. Without this, the `sku-init` test was failing because an eslint config importing from `sku/config/eslint` would be generated, but the latest published version of sku doesn't have that entrypoint configured yet.
- Fixes for errors and warnings within the sku monorepo flagged by the new eslint version and config


.changeset/pink-ghosts-battle.md
```diff
@@ -0,0 +1,58 @@
...
+
+Additionally, the `.eslintignore` file is no longer used by ESLint. Instead, [ignore patterns] are now configured directly in your ESLint config file. Since `sku` controls your ESLint config, you can configure any custom ignore patterns in your sku config via the new `eslintIgnore` option.
+
+For example:
+
+```ts
+import type { SkuConfig } from 'sku';
+
+export default {
+  eslintIgnore: ['a-large-file.js', '**/generated/]
```
DanDroryAu, line 41: **suggestion:** missing an apostrophe here.

```suggestion
  eslintIgnore: ['a-large-file.js', '**/generated/']
```

packages/sku/config/eslint/index.js
```diff
@@ -0,0 +1,27 @@
...
+// @ts-expect-error `eslint-config-seek` has no types yet
+const eslintConfigSeek = require('eslint-config-seek');
+const { importOrderConfig } = require('./importOrder.js');
+const { createEslintIgnoresConfig } = require('./ignores.js');
+const {
+  eslintDecorator,
+  eslintIgnore,
+  languages,
+  paths: { relativeTarget },
+} = require('../../context/index.js');
```
DanDroryAu, line 11: **question:** Is the barrel file not picked up here?

Also I don't believe the `.js` filetypes are needed here. Correct me if I'm wrong.

packages/sku/lib/configure.js
```diff
@@ -1,32 +1,31 @@
...
-const managedConfigBanner = require('./managedConfigBanner.js');
+const { getPathFromCwd, writeFileToCWD } = require('./cwd');
 
 const coverageFolder = 'coverage';
 
 const convertToForwardSlashPaths = (pathStr) => pathStr.replace(/\\/g, '/');
 const addSep = (p) => `${p}${path.sep}`;
-const prependBanner = (str) => `${managedConfigBanner}\n${str}`;
-
-const writeFileToCWD = async (fileName, content, { banner = true } = {}) => {
```
DanDroryAu, line 23: **praise:** nice refactor of this function.

packages/sku/config/eslint/importOrder.js
```diff
@@ -7,14 +7,14 @@ const internalRegex = `^(${paths.src
 
 const rootResolutionConfig = {
   settings: {
-    'import/internal-regex': internalRegex,
+    'import-x/internal-regex': internalRegex,
   },
 };
 
-module.exports = {
+const importOrderConfig = {
```
DanDroryAu, line 14: **note:** This will be great for the new CLI work.

packages/sku/config/eslint/ignores.js
```diff
@@ -0,0 +1,40 @@
...
+ * @param {string | undefined} options.target - The configured target directory in sku config
+ */
+const createEslintIgnoresConfig = ({ hasLanguagesConfig, target }) => {
+  // ESLint migrates ignore entries differently depending on whether a path contains a subdirectory.
+  // We want the sku target to match the migrated entry so it ends up excluded in the migrated
+  // output.
+  const targetIgnore =
+    target && target.includes('/') ? `${target}/` : `**/${target}/`;
+
+  const ignores =
```
DanDroryAu, line 17: **thought:** Worth mentioning the ignore list in the docs?

packages/sku/config/eslint/index.js
askoufis: My IDE auto-completed the path, and I guess it chose to be explicit. The `.js` isn't needed, but if we start writing ESM then it will be necessary, though I guess that can also be avoided if we use a bundler.

The extension on the import actually result in faster module resolution because node doesn't have to figure the extension. But overall it's not a huge deal in this codebase so it's not something I pay much attention to.

packages/sku/config/eslint/ignores.js
askoufis: Yeah I think so.

askoufis: Added in https://github.com/seek-oss/sku/pull/1094/commits/9c1fd38cca36e8e8d06d45917bdd7405e0db02d6. Also realized I didn't add an entry for `eslintIgnore` to the configuration page, so did that too.

DanDroryAu (approved):

PR #1096, opened 2024-11-27
Change cli interface to commander.js

DanDroryAu:
> [!CAUTION]
> This is a breaking change and should not be merged until other breaking changes have been reviewed and it has been agreed that a new major version should be released.

## CLI change

Uplifting the CLI for `sku`. The current CLI uses [`minimist`](https://www.npmjs.com/package/minimist) for parsing. It's implementation was very loose in the definition of commands, arguments, and options. 

The new CLI uses [`commander.js`](https://github.com/tj/commander.js). This allows for more tightly scoped commands, arguments, and options. Further more the built in `-h, --help` flag displays the usage to the consumer. When changing a command you can now more easily see what options and arguments a specific command uses.

Specific options are now more tightly scoped and only accept certain values.

global options:
- the `-e, --environment` option now only accepts a valid `node.js` environment.
- the `-c, --config` option accepts the path to the `sku.config.js` file.
- the `-p, --packageManager` option accepts a string for the packagemanger you're using.
- the `-d, --debug` option sets debugging to `true`.

scoped options:
- the `build`, `build-ssr`, `start`, and `start-ssr` commands have the following options
  - `-s, --stats` the webpack preset to use. This option now only accepts a valid webpack stats preset.
- the `serve` command now has the following options
  - `--port` the port you want to serve on. This option accepts a port as a number. 
  - `--site` the url you want to serve on. This option accepts a site as a string. 
- the `test`, and `translation compile` commands have the following options
  - `-w, --watch` makes the commands watch for changes.
- the `translation push` command now has the following option
  - `--delete-unused-keys` removes unused keys from translation files when pushing to Phrase. 

## Other Changes
- Add `plop` for code generation.
- Add `add-command` plop generator. This creates a new command boilerplate. It allows for sub-command generation with a depth of one.


packages/sku/lib/program/commands/init/init.action.js
```diff
@@ -228,7 +203,8 @@ const packageNameRegex =
   );
   console.log();
 
-  const logLevel = args.argv.indexOf('--verbose') >= 0 ? 'verbose' : 'regular';
+  // TODO: add loglevel to sku init as option.
```
jahredhope, line 206: Is this a TODO for this PR?

If it's just a suggestion for a later change I'd suggest leaving the comment out and tracking the suggested change/feature separately.

packages/sku/lib/program/commands/translations/commands/compile/compile.action.js
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    await compile({ watch }, vocabConfigFromSkuConfig);
+
+    if (!watch) {
+      log('Successfully compiled translations');
+    }
+  } catch (e) {
+    if (e) {
+      console.error(`Error running ${chalk.bold(`translations compile`)}:`, e);
```
jahredhope, line 28: Is it safe to swallow the error here? Shouldn't it be re-thrown or in some way returned so the error code is set?

packages/sku/lib/program/commands/pre-commit/pre-commit.action.js
```diff
@@ -0,0 +1,13 @@
+const preCommit = require('../../../preCommit');
+const { configureProject } = require('../../../utils/config-validators');
+
+const preCommitAction = async () => {
+  await configureProject();
+  try {
+    await preCommit();
+  } catch (e) {
+    process.exit(1);
```
jahredhope, line 9: Is it possible for a thrown error to not otherwise have been logged out? It feels like this could end up hiding errors, where the command fails with no reason why.

packages/sku/lib/program/commands/init/init.action.js
DanDroryAu: Tying this in to whatever you said in your general comment. How about I re-work the `-d, --debug` flag to actually accept a level and we use that to determine all the other levels of logging in the application?

I don't think I can get everything to the correct level of logging throughout the application in one go. But doing this sets the CLI api so that we can upgrade things over time that make more sense.

packages/sku/lib/program/options/config/config.option.js
```diff
@@ -0,0 +1,7 @@
+const { Option } = require('commander');
+const configOption = new Option(
+  '-c, --config [config]',
+  'Path to your sku.config.js file',
```
askoufis, line 4: Sku config can be written in either a `.js` or `.ts` file, so I try to refer to it agnostic to the file type, i.e. just "sku config".
```suggestion
  'Path to your sku config file',
```

packages/sku/lib/program/commands/init/init.action.js
jahredhope: How do you feel about just removing the comment?
Or change the comment to describe why it is how it is rather than saying how it should change in the future.

packages/sku/plop/helpers/camel-case-helper.mjs
```diff
@@ -0,0 +1,9 @@
+import camelCase from 'lodash/camelCase.js';
```
askoufis, line 1: I've developed an allergy towards lodash lately. It's not inherently bad, but I'd rather not add it as a dependency (even as a dev dep) if we're just using a single utility. Since this is ESM, we could use [`camelcase`](https://www.npmjs.com/package/camelcase) instead.

.changeset/olive-cups-kiss.md
```diff
@@ -0,0 +1,34 @@
...
+
+**BREAKING CHANGE**
+
+SKU now uses [commander.js](https://github.com/tj/commander.js) for its CLI. The CLI options and commands are now more narrowly defined and scoped. Some combinations of options that were previously unintended may no longer work.
+
+The new `-h, --help` flag now shows the help output generated by commander.js and shows the available commands and options.
+
+The following option flags are now available in their respective scope
+
+#### global options:
```
askoufis, line 15: These are heading, so should ideally be title case, or at least capitalized.
```suggestion
#### Global options
```

.changeset/olive-cups-kiss.md
```diff
@@ -0,0 +1,34 @@
...
+
+- the `-e, --environment` option now only accepts a valid node.js environment.
+- the `-c, --config option` accepts the path to the sku.config.js file.
+- the `-p, --packageManager` option accepts a string for the packagemanger you're using.
+- the `-d, --debug` option sets debugging to true.
+- __new:__ the `-h, --help` flag now shows you the full CLI usage of any command or subcommand
+- __new:__ the `-v, --version` flag now shows you the version of the CLI
+
+#### scoped options:
```
askoufis, line 24:
```suggestion
#### Scoped options
```

packages/sku/lib/program/commands/init/init.action.js
DanDroryAu: I feel good removing the comment. I'm just spitballing here on how we want to do all the error handling. I left it in there for me to have a second look at this. What I'm going to do is update the description of the option in the command interface and remove this todo.

packages/sku/lib/program/commands/init/init.action.js
```diff
@@ -97,27 +69,27 @@ const packageNameRegex =
 
   if (!isValidPackageName) {
     console.error(dedent`
-    Could not create a project called ${chalk.red(
-      `"${appName}"`,
-    )} because of npm naming restrictions. \
-    Please see https://docs.npmjs.com/cli/configuring-npm/package-json for package name rules.
-  `);
+  Could not create a project called ${chalk.red(
```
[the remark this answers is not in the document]
askoufis: I think this happened because this code is now inside a function rather than an IIFE, but not 100% certain.

packages/sku/lib/program/commands/init/init.command.js
```diff
@@ -0,0 +1,14 @@
...
+
+const init = new Command('init');
+
+init
+  .description('Initialize a new sku project')
+  .argument('[packageName]', 'Package name')
+  .option('--verbose', 'Verbose log output')
+  .action((packageName, verbose) => {
+    const initAction = require('./init.action');
+    initAction(packageName, { verbose });
```
askoufis, line 14: Do async actions need to be awaited, or does `commander` handle this for you/not care?

packages/sku/lib/program/commands/setup-hosts/setup-hosts.command.js
```diff
@@ -0,0 +1,14 @@
+const { Command } = require('commander');
+
+const setupHosts = new Command('setup-hosts');
+
+setupHosts
+  .description(
+    'Update your hosts file to point any configured hosts to your local machine.',
+  )
```
askoufis, line 8: Should we mention here that you'll need admin (`sudo`) permissions?

packages/sku/lib/program/commands/translations/commands/push/push.command.js
```diff
@@ -0,0 +1,14 @@
...
+const deleteUnusedKeysOption = require('./delete-unused-keys.option');
+
+const push = new Command('push');
+
+push
+  .description('Push translations to Phrase')
+  .addOption(deleteUnusedKeysOption)
+  .action(() => {
+    const pushAction = require('./push.action');
+    pushAction();
```
askoufis, line 11: Looks like `deleteUnusedKeysOption` needs to be passed through `pushAction`.

packages/sku/lib/program/commands/translations/commands/compile/compile.action.js
askoufis: The old `translations` command would `process.exit(1)` after catching any sub-command error, so maybe each sub-command needs its own one now.

packages/sku/plop/generators/add-command/index.mjs
askoufis, line 1: Worth thinking about how we can exclude all the files in the `plop` directory from being published. Can do it in another PR, but potentially these files could live in a separate (private) monorepo package, or we could explicitly include files we want published via [`package.json#files`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files).

Up until now we just included everything because all files were needed, but these files certainly don't need to be published.

DanDroryAu:
> This is a really exciting change.
> 
> I have some questions about error handling. I wouldn't be surprised if it's all the re-create previous behaviours, but given this is a breaking change already I'd think it worth standardising how errors are caught/displayed at some point.

Good point. I too think that the error handling can be improved quite a bit. However this work is out of scope for this pr. Since we've already got a few breaking changes lined up I think we should create a feature branch for this work and separately merge a pr in that aims to solve the error handling work.

.changeset/olive-cups-kiss.md
```diff
@@ -0,0 +1,34 @@
...
+
+- the `build`, `build-ssr`, `start`, and `start-ssr commands` have the following options
+  - `-s, --stats` the webpack preset to use. This option now only accepts a valid webpack stats preset.
+- the `serve` command now has the following options
+  - `--port` the port you want to serve on. This option accepts a port as a number.
+  - `--site` the url you want to serve on. This option accepts a site as a string.
+- the `test`, and `translation compile` commands have the following options
+  - `-w, --watch` makes the commands watch for changes.
```
askoufis, line 32: I think `--watch` should really only be mentioned for `translations compile`. As far as `sku test` is concerned it's just forwarding arguments to `jest`.

I think the only reason it's explicitly forwarded to `jest` in the `test` script is because `minimist` didn't have a nice way to get all arguments in a list when you wanted to forward them, so I ended up just explicitly passing through `--watch`.

packages/sku/lib/utils/config-validators.js
askoufis, line 1: I think the name of this file is misleading. Nothing in this file validates config, Maybe it should just be called `configure.js`?

packages/sku/lib/program/options/stats/stats.option.js
```diff
@@ -0,0 +1,16 @@
+const { Option } = require('commander');
+const statsOption = new Option(
+  '-s, --stats <preset>',
+  'Webpack stats presets',
```
askoufis, line 4: I think this could be a bit more descriptive:
```suggestion
  'Webpack stats preset to pass through to webpack',
```

packages/sku/lib/program/options/site/site.option.js
```diff
@@ -0,0 +1,4 @@
+const { Option } = require('commander');
+const siteOption = new Option('-si, --site [site]', 'Site to serve on');
```
askoufis, line 2: Honestly not a fan of `-si`, and I think this sort of breaks short-hand flag composition, e.g. if `-s` and `-i` were both boolean flags you could do `-si` to set both.

Even though those aren't boolean flags in sku, I'd rather not confuse anyone that expects `-si` to set two flags. Since `-s` is already taken by `--stats`, I'd just remove the shorthand for this option.

packages/sku/lib/program/commands/translations/commands/compile/compile.action.js
askoufis: Actually, is it possible to hoist the `catch` to the top-level `translations` command, and then call `process.exit(1)` there.

packages/sku/lib/program/options/environment/environment.option.js
```diff
@@ -0,0 +1,7 @@
+const { Option } = require('commander');
+const environmentOption = new Option(
+  '-e, --environment [environment]',
+  'Node environment to run in',
```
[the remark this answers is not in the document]
DanDroryAu: Alrighty, I'll edit this.

.changeset/olive-cups-kiss.md
DanDroryAu: Good point. It had some weird way of setting it and forwarding it before. I think that was probably a limitation of `minimist` and the way it was setup.

packages/sku/lib/utils/config-validators.js
DanDroryAu: Yeah I wasn't too sure on this file. Configure.js may be confusing since there are a couple of files already named that. It may just be a command helper.

packages/sku/lib/program/options/site/site.option.js
DanDroryAu: Good call. This should just be a single flag without a shortcut. I think I mentioned it without a shortcut in the PR description.

packages/sku/lib/program/commands/setup-hosts/setup-hosts.command.js
DanDroryAu: Yeah we could mention that here. I'll see if I can add it in concisely.

packages/sku/lib/program/commands/init/init.command.js
DanDroryAu: We don't care if we handle errors internally. If we want to add some custom error handling in commander we may have to switch to using async functions.

packages/sku/lib/program/commands/translations/commands/compile/compile.action.js
DanDroryAu: If the functions are async I can override the error behaviour of the command. let me check what I can do there. I don't think the translation action handler is aware of whats happening in the subcommands.

packages/sku/package.json
```diff
@@ -87,6 +88,7 @@
     "browserslist": "^4.16.1",
     "browserslist-config-seek": "3.2.0",
     "chalk": "^4.1.0",
+    "commander": "^12.1.0",
```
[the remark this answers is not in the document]
askoufis: Oh, I forgot that we used it for the server entry too. Maybe we could move to node 20 as part of the breaking change then just use https://nodejs.org/api/util.html#utilparseargsconfig for the server entry since it's so simple.

packages/sku/lib/program/commands/init/init.command.js
DanDroryAu: > I'm more concerned with whether commander will wait for the action to finish

What should it do once the action is finished? From my understanding is that it does nothing when the action finishes. If the action throws it does something but in this case we're not awaiting the action so any error thrown inside will not be captured by commander.

We don't need to await the `parseAsync` I think. We only want to await it if we're doing something else after the program has finished.

I.E.

```js
// sync.
program.parse();
// program finished here. Your actions are done.
console.log('do something after');

// async.
program.parseAsync();
// program not finished here. Console log firing before your program has finished running.
console.log('do something after');
```

I'll have to test it to see if this is the case. This is how I understand it.

packages/sku/config/targets.json
```diff
@@ -1,4 +1,4 @@
 {
-  "browserslistNodeTarget": "node 18.20",
-  "esbuildNodeTarget": "node18.20"
+  "browserslistNodeTarget": "node 20.18.1",
+  "esbuildNodeTarget": "node20.18.1"
```
askoufis, line 3: These were intentionally only locked to minor versions I think because browserslist doesn't track anything more granular than minor versions. It also seems like [browserslist doesn't recognize anything past node 20.15](https://browsersl.ist/#q=node+%3E+20) just yet, so not sure if this is going to work 🤔

DanDroryAu: I think this was mise auto updating this stuff.

askoufis: It's the monorepo's local postinstall script actually. https://github.com/seek-oss/sku/blob/6dab604a1be76df5d6346a000b3ac880fb829752/postinstall.js#L13

DanDroryAu: Shall we set it to 20.15 then? I just picked the latest 20.18.1 LTS

askoufis: Yeah let's do that for now. Maybe by the time we release 20.18 will work.

.changeset/olive-cups-kiss.md
```diff
@@ -37,4 +37,4 @@ The following option flags are now available in their respective scope
 
 #### Node version change
 
-Moved to node `v20.18.1`
+Moved to node `v20.15`
```
askoufis: This is a significant enough change that it should be in its own changeset and have a CTA for consumers to update their node version. See [here](https://github.com/seek-oss/sku/pull/986/files#diff-495e18c37d72bd31e3e6052dc0d76b80d6102de62f8e25e5eea3d984dd2ea29d) for prior art.

PR #1097, opened 2024-11-28
Drop support for React 17

askoufis:
All of sku's dependencies support react 18 (at least according to their peer dependency ranges), so this is just a peer dep change.


DanDroryAu (approved):
:shipit:

PR #1098, opened 2024-11-28
Don't try to migrate `.eslintignore` if it doesn't exist

askoufis:
This fixes an issue where ignore file migration (handled by an eslint dep) would throw a file not found error.


packages/sku/lib/configure.js
```diff
@@ -39,39 +39,44 @@ module.exports = async () => {
...
-    console.log("'.eslintignore' file detected. Attempting migration...");
+  const { shouldMigrate, eslintIgnoreExists } =
+    await shouldMigrateOldEslintConfig();
 
-    const customIgnores = migrateEslintignore({
-      hasLanguagesConfig: Boolean(languages && languages.length > 0),
-      target: paths.relativeTarget,
-    });
+  if (shouldMigrate) {
+    console.log("'.eslintignore' file detected. Attempting migration...");
```
DanDroryAu, line 46: **suggestion:** We may want to change this message.

Or move it into the second if block. We may not have an ignorefile here.

packages/sku/lib/configure.js
```diff
@@ -39,39 +39,44 @@ module.exports = async () => {
...
       }
-
-      console.log(
-        "Please note that this is a best-effort migration and you should manually review the 'eslintIgnore' property in your sku config.",
-      );
-    } else {
-      console.log("No custom ignores found in '.eslintignore' file");
     }
 
     console.log("Deleting '.eslintrc' and '.eslintignore' files...");
```
DanDroryAu, line 82: **suggestion:** This message can be misleading

Being more explicit in what we are actually removing can be more clear. If we say we're removing your `.eslintignore` file and you have none you may not believe the next thing sku tells you ;)

DanDroryAu (commented):
Couple changes regarding the logs.

packages/sku/lib/configure.js
DanDroryAu: There is one right below this one too that mentions both files.

askoufis: Maybe we can cover all cases by logging "Deleting `.eslintrc` and `.eslintignore` files if they exist"?

askoufis: We probably don't even need the success message, if it throws an error we'll likely catch it elsewhere.

DanDroryAu (approved):
:shipit:

PR #1099, opened 2024-11-28
Uplift cli docs

DanDroryAu:
### Changes

Uplifting the CLI docs to reflect the new interface.

Fixed a typo in the changeset and added more information regarding a flag change.

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -30,11 +30,11 @@ pnpm dlx sku init my-app
 
 This command supports the following options and arguments:
 
-| Type     | Name                           | Description                                                                                                   |
-| -------- | ------------------------------ | ------------------------------------------------------------------------------------------------------------- |
-| argument | `project-name` <br> _required_ | The name of the project to create <br> `pnpm dlx sku init my-app`                                             |
-| option   | `--package-manager, -p`        | The package manager to use when installing dependencies <br> `pnpm dlx sku init my-app --package-manager=npm` |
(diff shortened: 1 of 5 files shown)


docs/docs/cli.md
```diff
@@ -26,6 +28,14 @@ Unless you've installed `sku` globally (not recommended), you should use `npx` o
 pnpm dlx sku init my-app
 ```
 
+This command supports the following options and arguments:
+
+| Type     | Name                           | Description                                                                                                   |
+| -------- | ------------------------------ | ------------------------------------------------------------------------------------------------------------- |
+| argument | `project-name` <br> _required_ | The name of the project to create <br> `pnpm dlx sku init my-app`                                             |
+| option   | `--package-manager, -p`        | The package manager to use when installing dependencies <br> `pnpm dlx sku init my-app --package-manager=npm` |
```
askoufis, line 36: Worth mentioning that this defaults to the package manager being used to run `sku init`, so it's really more of an override.

docs/docs/cli.md
```diff
@@ -187,4 +228,13 @@ See the documentation on [phrase-specific features] for more information.
 sku translations pull
 ```
 
+### `translations validate`
+
+Validate translations defined in `.vocab` directories.
```
askoufis, line 233: We should probably define what this actually means. Probably worth documenting this more thoroughly in `vocab` and then just linking to it. Happy to add this as-is for now though.

askoufis (approved):
Just some tiny comments.

PR #1103, opened 2024-12-02
Functionalize global args

askoufis:
Replaces direct usage of `program` to get global args by either passing the args into the action handler from the command handler, or in the case of the sku config path, still setting it globally, but ✨elsewhere✨.


DanDroryAu (approved):
LGTM

PR #1106, opened 2024-12-02
Test command mjs

DanDroryAu:
Turning the test command file into MJS.

Also updated the name of the generator for plop.

### DIFF
--- a/packages/sku/lib/program/commands/test/test.command.mjs
+++ b/packages/sku/lib/program/commands/test/test.command.mjs
@@ -1,8 +1,8 @@
 import { Command } from 'commander';
 
-const test = new Command('test');
+const testCommand = new Command('test');
 
-test
+testCommand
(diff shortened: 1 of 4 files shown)


packages/sku/lib/program/commands/test/test.command.mjs
```diff
@@ -0,0 +1,13 @@
...
+
+testCommand
+  .description('Run unit tests.')
+  .allowUnknownOption(true)
+  .action(async ({ args }) => {
+    const { testAction } = await import('./test.action.mjs');
+    testAction({ args });
+  });
+
+export { testCommand };
```
askoufis, line 13: Oh I changed mine to exporting the const directly 😅, thought you meant we should standardize on that approach rather than this one.

We'll fix it in post.

askoufis (approved):

PR #1109, opened 2024-12-02
 Migrate `build` command and action to mjs

DanDroryAu:


packages/sku/lib/program/commands/build/build.action.mjs
```diff
@@ -1,27 +1,27 @@
 // First, ensure the build is running in production mode
 process.env.NODE_ENV = 'production';
```
askoufis, line 2: In #1107 I moved this line to below the imports. Not sure if it matters 🤷

askoufis (approved):

PR #1114, opened 2024-12-02
Migrate start and start-ssr commands to mjs.

DanDroryAu:


packages/sku/lib/program/commands/start/start.action.mjs
```diff
@@ -91,6 +91,8 @@ const startAction = async ({
 
   const appHosts = getAppHosts();
 
+  const devServerMiddleware = await import(paths.devServerMiddleware);
+
```
askoufis, line 95: I think the `if (useDevServerMiddleware)` should be moved to surround this import. Otherwise the dynamic import could end up importing `undefined`.

askoufis (approved):

PR #1115, opened 2024-12-02
Convert `serve` command and action to mjs

askoufis:


packages/sku/lib/program/commands/serve/serve.action.mjs
```diff
@@ -89,7 +89,7 @@ const serveAction = async ({
   const app = express();
 
   if (useDevServerMiddleware) {
-    const devServerMiddleware = require(paths.devServerMiddleware);
+    const devServerMiddleware = await import(paths.devServerMiddleware);
```
askoufis, line 92: Converted this to an async import. We _may_ need to handle CJS->ESM interop here (check if there's a `default` property` on the import result), but we can do that later.

DanDroryAu (approved):

PR #1121, opened 2024-12-03
Convert some files in `lib` to ESM

askoufis:

Note that file extensions in dependent files have intentionally not been updates. They will be addressed in the future.


packages/sku/lib/allocatePort.mjs
```diff
@@ -1,6 +1,6 @@
 // @ts-check
-const { yellow, bold } = require('chalk');
-const getPort = require('get-port');
+import { yellow, bold } from 'chalk';
+import getPort from 'get-port';
 const debug = require('debug')('sku:allocatePort');
```
DanDroryAu, line 4: **issue:** Still a require here. Its being immediately invoked.

```suggestion
import debug from 'debug';

debug('sku:allocatePort');
```

askoufis: Good catch

DanDroryAu (approved):

PR #1123, opened 2024-12-03
Migrate more files to ESM

askoufis:

Note that file extensions in dependent files have intentionally not been updated. They will be addressed in the future.

### DIFF
--- a/packages/sku/lib/hosts.mjs
+++ b/packages/sku/lib/hosts.mjs
@@ -1,20 +1,20 @@
-const { promisify } = require('node:util');
-const hostile = require('hostile');
-const { red, yellow, bold } = require('chalk');
+import { promisify } from 'node:util';
+import { set, get } from 'hostile';
+import { red, yellow, bold } from 'chalk';
 
-const { hosts, sites: contextSites } = require('../context');
-const { suggestScript } = require('./suggestScript');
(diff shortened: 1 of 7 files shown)


packages/sku/lib/hosts.mjs
```diff
@@ -63,10 +63,3 @@ const checkHosts = async () => {
     // swallow error as this just a warning check
   }
 };
-
-module.exports = {
-  setupHosts,
-  checkHosts,
-  getAppHosts,
-  getMissingHosts,
```
askoufis, line 71: `getMissingHosts` is not used outside this file, so it is no longer exported.

packages/sku/lib/install.mjs
```diff
@@ -1,13 +1,13 @@
 // @ts-check
-const { getAddCommand } = require('./packageManager');
+import { getAddCommand } from './packageManager.js';
 
-const spawn = require('cross-spawn');
+import { spawn } from 'cross-spawn';
 
 /**
- * @param {import("../lib/packageManager").GetAddCommandOptions} options
```
DanDroryAu, line 7: **praise:** A keen eye on a keen guy!

DanDroryAu (approved):

PR #1124, opened 2024-12-03
Turn all runner files in Lib into MJS

DanDroryAu:

Note that file extensions in dependent files have intentionally not been updated. They will be addressed in the future.


packages/sku/lib/runBin.mjs
```diff
@@ -13,7 +16,7 @@ const resolveBin = (packageName, binName) => {
       ? packageJson.bin
       : packageJson.bin[binName || packageName];
 
-  return require.resolve(path.join(packageName, binPath));
+  return import.meta.resolve(path.join(packageName, binPath));
```
askoufis, line 19: This returns a file URL, rather than a file path. I think passing it to [`fileUrlToPath`](https://nodejs.org/api/url.html#urlfileurltopathurl-options) should get it into the right form.

packages/sku/lib/runBin.mjs
```diff
@@ -1,6 +1,9 @@
 // @ts-check
-const path = require('node:path');
-const spawn = require('cross-spawn');
+import path from 'node:path';
+import spawn from 'cross-spawn';
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
```
askoufis, line 6: Was thinking that this might not work if the thing being required is ESM. I guess that's the entire reason why synchronous ESM require is such a big deal, otherwise everyone would just use this. We'll see later on whether this actually works, but this might be a situation we _do_ want to refactor as part of the ESM migration.

DanDroryAu: I think once we're in the "bug resolution" phase of the beast it would be good to do small refactors and turn everything into async imports if possible.

DanDroryAu: Currently, this is only requiring a `package.json` file.

askoufis: Oh yeah good point. Should hopefully be fine then? We'll find out I guess.

DanDroryAu: > We'll find out I guess.

what_could_go_wrong

packages/sku/lib/runBin.mjs
askoufis: @DanDroryAu what about this one?

DanDroryAu: Wait... I may have messed this up. I thought I pushed a fix for this. Let me check my local.

DanDroryAu: > Your branch is ahead of 'origin/v14-lib-runners-mjs' by 1 commit.

Classic.

askoufis (approved):

PR #1127, opened 2024-12-03
Migrate config/weback over to mjs

DanDroryAu:


packages/sku/config/webpack/plugins/metrics-plugin/index.mjs
```diff
@@ -1,7 +1,9 @@
-const { performance } = require('node:perf_hooks');
-const prettyMilliseconds = require('pretty-ms');
-const debug = require('debug')('sku:metrics');
-const track = require('../../../../telemetry');
+import { performance } from 'node:perf_hooks';
+import prettyMilliseconds from 'pretty-ms';
+import debug from 'debug';
+import { timing } from '../../../../telemetry';
+
+debug('sku:metrics');
```
askoufis, line 6: Should this be assigned to something?

askoufis (approved):

packages/sku/config/webpack/plugins/metrics-plugin/index.mjs
DanDroryAu: It should be a `const log = debug('sku:metrics');`. I'll update.

DanDroryAu: Also renaming the rest of the debug calls to log.

DanDroryAu: I've found some other instances of this mistake. I've added them to this pr.

PR #1129, opened 2024-12-03
Migrate final config files to mjs

DanDroryAu:
Moved the final config files to mjs.

Changed a final few requires to imports in `cwd.mjs`

Moved `configure.js` in `lib` over to `mjs`


packages/sku/config/storybook/index.mjs
askoufis, line 1: Need to confirm that this will actually work at some point. The comment I wrote at the top of the file a while ago is there because I initially tried to make this config ESM, but storybook didn't like it.

askoufis (approved):

packages/sku/config/storybook/index.mjs
DanDroryAu: Yeah might revert this.

PR #1131, opened 2024-12-03
Refer to full node version for Volta config

jahredhope:
The shorthand 20.15 is not something volta is happy with.

Displaying error `Volta error: Could not parse version "20.15"`

Screenshot 2024-12-03 at 5 03 49 PM

---

Refer to the full version


askoufis:
I think I never noticed this because I'm using mise now. IMO we should ditch Volta and use .node-version or something.

jahredhope:
I agree. But I wanted to fix it as long as it's there and move over deliberately.

askoufis:
If you run `pnpm install` the `postinstall` script should also update `.nvmrc`. That should be committed too.

askoufis:
Also would be better to merge this into `v14-mjs-refactor` IMO.

PR #1135, opened 2024-12-05
fix all test cases after mjs refactor

DanDroryAu:
Fixed all test cases. This does not mean everything is working 100%.


package.json
```diff
@@ -10,8 +10,8 @@
     "lint:prettier": "prettier --cache --list-different .",
     "lint:tsc": "tsc",
     "check": "pnpm install --frozen-lockfile && echo 'Ignore paths from lint-staged'",
-    "test": "SKU_TELEMETRY=false SKU_DISABLE_CACHE=true OPEN_TAB=false jest --verbose",
-    "test:sku-init": "SKU_TELEMETRY=false SKU_DISABLE_CACHE=true OPEN_TAB=false jest --verbose --config=jest.config.sku-init.js",
+    "test": "SKU_TELEMETRY=false SKU_DISABLE_CACHE=true NODE_OPTIONS=\"--experimental-vm-modules\" OPEN_TAB=false jest --verbose",
+    "test:sku-init": "SKU_TELEMETRY=false SKU_DISABLE_CACHE=true NODE_OPTIONS=\"--experimental-vm-modules\" OPEN_TAB=false jest --verbose --config=jest.config.sku-init.js",
```
askoufis, line 14:
```suggestion
    "test": "SKU_TELEMETRY=false SKU_DISABLE_CACHE=true NODE_OPTIONS='--experimental-vm-modules' OPEN_TAB=false jest --verbose",
    "test:sku-init": "SKU_TELEMETRY=false SKU_DISABLE_CACHE=true NODE_OPTIONS='--experimental-vm-modules' OPEN_TAB=false jest --verbose --config=jest.config.sku-init.js",
```

packages/sku/config/jest/jest-preset.js
```diff
@@ -8,6 +8,8 @@ const compilePackagesRegex = paths.compilePackages
   .map((pkg) => `.*${escapeRegex(pkg)}`)
   .join('|');
 
+console.log('paths:', paths);
```
askoufis, line 11:
```suggestion
```

tests/sku-with-https.test.js
```diff
@@ -6,26 +6,33 @@ import {
...
-);
+let appDir;
+
+try {
+  appDir = path.dirname(
+    require.resolve('@sku-fixtures/sku-with-https/sku.config.mjs'),
+  );
+} catch (error) {
+  console.error(error);
+}
```
askoufis, line 25: I'm guessing that this was debug code that can be cleaned up now?

packages/sku/context/index.js
```diff
@@ -1,23 +1,20 @@
 import { existsSync } from 'node:fs';
 import { join } from 'node:path';
 import chalk from 'chalk';
-import { register } from 'esbuild-register/dist/node';
```
askoufis, line 4: I think both `esbuild` and `esbuild-register` can now be removed as dependencies with this change.

fixtures/assertion-removal/sku.config.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-ignore TS1479
```
askoufis, line 1: We'll definitely need to address this somehow (in another PR).

DanDroryAu: Yeah, this was a bit of a struggle to get all the lint to pass. I don't think TSC is picking the correct tsconfig.json settings. It should exclude these files. It should probably also exclude the test files since these require the fixture SKU configs.

fixtures/assertion-removal/sku.config.ts
```diff
@@ -1,3 +1,4 @@
...
 import type { SkuConfig } from 'sku';
```
askoufis, line 2: Apparently this is a thing:
```suggestion
import type { SkuConfig } from 'sku' with { 'resolution-mode': 'import' };
```

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-3.html#stable-support-resolution-mode-in-import-types

Not ideal if consumers have to do it though, but maybe unavoidable?

DanDroryAu: I'll have a look at it.

askoufis (approved):

PR #1137, opened 2024-12-13
Moved the program folder and dependencies over to typescript

DanDroryAu:
The `program` folder in the `sku` package is now typescript.

Other updates:

Update the `tsc` check. The root `tsc` check now excludes the `sku` package. `sku` now checks itself via a filter lint run. 
Update the `eslint.config.js`. It now runs `@typescript-eslint/parse` for each separate project instead of applying the root `tsconfig.json` for the `sku` package folder.


package.json
```diff
@@ -9,7 +9,7 @@
     "lint": "pnpm run '/^lint:.*/'",
     "lint:eslint": "eslint --cache --cache-location 'node_modules/.cache/eslint/.eslintcache' .",
     "lint:prettier": "prettier --cache --list-different .",
-    "lint:tsc": "tsc",
+    "lint:tsc": "tsc && pnpm --filter 'sku' lint:tsc",
```
askoufis, line 12: We could run these in parallel if instead there was a separate `lint:tsc:sku` command. The `lint` command above runs all `lint:*` scripts in parallel.

packages/sku/src/lib/program/commands/start/start.action.ts
```diff
@@ -122,60 +123,57 @@ export const startAction = async ({
...
-          try {
-            chosenLanguage = getLanguageFromRoute(req, matchingRoute);
-          } catch (e) {
-            return res.status(500).send(
-              exceptionFormatter(e, {
+      middlewares.push(((req, res, next) => {
+        const matchingSiteName = getSiteForHost(req.hostname, undefined);
+
+        const matchingRoute = routes.find(({ route }) =>
+          // Comments on this? SiteIndex does not exist on the route type. I am unaware if this is irrelevant now.
```
askoufis, line 130: It should exist here. I think the removal of [these types](https://github.com/seek-oss/sku/pull/1137/files#diff-8b865db76e7386b701375c086a35411fbd382810b5792995a6120cba8ff012cfL71-R89) made it disappear. I think the type changes during the mutation of `normalizedRoutes` inside the `forEach`, which isn't picked up by typescript. `normalizedRoutes` might need an explicit type still.

packages/sku/src/context/index.ts
```diff
@@ -147,47 +147,45 @@ if (devServerMiddleware && !useDevServerMiddleware) {
...
-  devServerMiddleware,
-  src: skuConfig.srcPaths.map(getPathFromCwd),
-  /** @type {string[]} */
-  compilePackages: [...defaultCompilePackages, ...skuConfig.compilePackages],
-  clientEntry: getPathFromCwd(skuConfig.clientEntry),
-  renderEntry: getPathFromCwd(skuConfig.renderEntry),
+  appSkuConfigPath: appSkuConfigPath!,
+  devServerMiddleware: devServerMiddleware!,
+  src: skuConfig.srcPaths!.map(getPathFromCwd),
+  compilePackages: [...defaultCompilePackages, ...skuConfig.compilePackages!],
```
askoufis, line 153: I know that we need all these `!`s right now, but removing in the future is high priority as they're just _the worst_.

askoufis (commented):
Just a general comment: in pretty much all cases I think we should be using `@ts-expect-error` over `@ts-ignore` (though sometimes it's unavoidable). That way it's easy to know when they can be removed: `tsc` tells you.

packages/sku/src/context/index.ts
DanDroryAu: Agreed. It was either this or do a `||` short circuit to ensure it had values.

askoufis (approved):

PR #1139, opened 2024-12-16
V14 ts lib folder

DanDroryAu:
Turned the `lib` folder into ts files. 

Add a moduleNameMapper removing `.js` extensions so that @swc/jest can pick up the transform.
Remove test files from the `tsc` build in `packages/sku`.
Add tsconfig.build.json for specified build outputs.


packages/sku/src/lib/language-utils.ts
```diff
@@ -37,22 +36,15 @@ export function getValidLanguagesForRoute(route) {
...
- * @param {string} pathname
- * @param {string} route
- */
-function getLanguageParamFromUrl(pathname, route) {
+function getLanguageParamFromUrl(pathname: string, route: any) {
   const match = routeMatcher(route)(pathname);
 
-  return match ? match.params[LANGUAGE_NAMED_PARAM] : null;
+  return match
+    ? (match.params as Record<string, any>)[LANGUAGE_NAMED_PARAM]
```
askoufis, line 43: Would prefer to avoid this `any` type creeping into the codebase. Passing a generic parameter to `match` seems to work well enough:
```ts
  return match<Record<string, string>>(normalisedRoute);
```

https://github.com/seek-oss/sku/blob/065bd6543dc3fa697f6052d79f83e03303420972/packages/sku/src/lib/routeMatcher.ts#L16

packages/sku/tsconfig.json
```diff
@@ -15,5 +15,11 @@
...
   },
   "include": ["./src/**/*"],
-  "exclude": ["node_modules", "**/@loadable"]
+  "exclude": [
+    "node_modules",
+    "**/@loadable",
+    "**/*.test.ts",
+    "**/*.test.tsx",
+    "**/*.test.js"
+  ]
```
askoufis, line 24: Again, would prefer to typecheck tests if possible. I think the typical way of doing this is to have a separate `tsconfig` for building. E.g.

```json
// tsconfig.build.json
{
 "extends": "./tsconfig.json",
 "compilerOptions": {
   "outDir": "./dist"
 },
 "exclude": ["**/*.test.ts"]
}
```

And then in the build script run `tsc --project tsconfig.build.json`.

packages/sku/src/lib/eslintMigration.ts
```diff
@@ -83,13 +73,15 @@ export const migrateEslintignore = ({
...
- * @param {string[]} options.eslintIgnore - Array of paths to ignore
  */
 export const addEslintIgnoreToSkuConfig = async ({
   skuConfigPath,
   eslintIgnore,
+}: {
+  // Path to the sku config file
+  skuConfigPath: string;
+  // Array of paths to ignore
+  eslintIgnore: string[];
```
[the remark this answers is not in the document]
DanDroryAu: Right... They show up in IntelliJ. I'll revert it back to `/** */` format.

packages/sku/src/lib/configure.ts
```diff
@@ -22,8 +22,13 @@ import { getPathFromCwd, writeFileToCWD } from './cwd.js';
 
 const coverageFolder = 'coverage';
 
-const convertToForwardSlashPaths = (pathStr) => pathStr.replace(/\\/g, '/');
-const addSep = (p) => `${p}${path.sep}`;
+const hasErrorMessage = (e: unknown): e is { message: string } =>
+  (e as { message: string }).message !== undefined;
```
askoufis, line 26: Casting inside this function sort of defeats its purpose. It's sort of a 'mostly safe cast' already, so I'd typically do something like checking for the property at runtime instead. It's a little verbose but it's thorough.

```suggestion
const hasErrorMessage = (e: unknown): e is { message: string } =>
  typeof e === 'object' &&
  e !== null &&
  'message' in e &&
  typeof e.message === 'string';
```

Ditto for the error code type guard.

askoufis (approved):
Happy to merge once the type guard stuff is addressed.

PR #1140, opened 2024-12-17
Refactor config folder to ts files.

DanDroryAu:
Change the `config` folder to typescript.


packages/sku/tsconfig.json
```diff
@@ -14,5 +14,11 @@
...
   },
   "include": ["./src/**/*"],
-  "exclude": ["node_modules", "**/@loadable"]
+  "exclude": [
+    "node_modules",
+    "**/@loadable",
+    "**/*.test.ts",
+    "**/*.test.tsx",
+    "**/*.test.js"
+  ]
```
askoufis, line 23: This was undone in the previous PR. Was it intentionally changed back in this PR?

eslint.config.js
```diff
@@ -19,6 +19,8 @@ module.exports = [
       '!test/test-cases/*/*.test.js',
       'fixtures/*/*',
       '**/@loadable/**/*',
+      '**/*.test.js',
+      '**/*.test.ts',
```
askoufis, line 23: This was undone in the previous PR. Was it intentionally changed back in this PR?

DanDroryAu: This was a merge issue. I'm rolling it back.

packages/sku/src/config/storybook/index.cjs
```diff
@@ -1,11 +1,12 @@
+// @ts-check
 /**
  * This file and all its dependencies must be CJS
  * https://github.com/storybookjs/storybook/pull/23018
  */
 
 /** @typedef {import("@storybook/react-webpack5").StorybookConfig} StorybookConfig */
 
-/** @type {NonNullable<StorybookConfig['webpackFinal']>} */
+/** @type {StorybookConfig['webpackFinal']} */
```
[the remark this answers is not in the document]
DanDroryAu: I'm getting this Type-error here

```TS1065: The return type of an async function or method must be the global Promise<T> type.
```
I think `NonNullable<T>` has an issue with the union return type of `Configuration | Promise<Configuration>` of the webpackFinal type.

I assume we use NonNullable to filter out the null values of the returned Configuration type of `webpackFinal`? I might have to change the type here a bit by reconstructing it and returning a `NonNullable<Configuration> | Promise<NonNullable<Configuration>>`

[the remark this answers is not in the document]
askoufis: Oh right, I guess it's just complaining because of the union, which is valid for a non-async function, but invalid for an async function.

The annotation was only there so the type of `webpackFinal` would be based off the actual config type storybook expects. It's a bit annoying to do that now, so I'd be happy with just `Promise<Configuration>`.

PR #1141, opened 2024-12-23
V14 entry ts refactor

DanDroryAu:
Change entry and config into .ts files.

Add type declarations for webpack globals.


packages/sku/src/context/configSchema.ts
```diff
@@ -1,5 +1,6 @@
 import Validator from 'fastest-validator';
 
+// @ts-expect-error
```
jahredhope, line 3: Can this have a reason, even just "Types not done yet" or something?

packages/sku/src/context/defaultSkuConfig.ts
```diff
@@ -1,15 +1,15 @@
 import browserslistConfigSeek from 'browserslist-config-seek';
 import { join } from 'node:path';
 import isCompilePackage from '../lib/isCompilePackage.js';
+import type { SkuConfig } from '../../sku-types.d.ts';
 
-const defaultDecorator = (a) => a;
+const defaultDecorator = (a: any) => a;
```
jahredhope, line 6: Should this have a comment explaining the any?

packages/sku/src/entry/client/index.tsx
```diff
@@ -8,6 +8,7 @@ import clientContextKey from '../clientContextKey.js';
 
 if (process.env.NODE_ENV === 'development') {
   if (typeof client !== 'function') {
+    // eslint-disable-next-line
```
jahredhope, line 11: Can this have an explanation as well please?

packages/sku/src/entry/server/server.ts
```diff
@@ -14,7 +15,7 @@ const serverOptions = serverExports({ publicPath });
 
 const { middleware, onStart, renderCallback } = serverOptions;
 
-const app = express();
+const app: Express = express();
```
jahredhope, line 18: If express has types why does this need to be declared?

jahredhope (commented):
Nice work, a few random thoughts

fixtures/translations/.gitignore
```diff
@@ -5,7 +5,7 @@
 .eslintcache
 .prettierrc
 coverage/
-dist/
+dist-ssr/
```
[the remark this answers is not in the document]
DanDroryAu: Seems to be yeah.

packages/sku/src/entry/client/index.tsx
DanDroryAu: Yeah I'll add it in. 

ESM doesn't want require however this file is being loaded in via webpack which needs the require. I can probably refactor this to import `dedent` at the top and use it that way.

packages/sku/src/context/defaultSkuConfig.ts
DanDroryAu: This should be a T

packages/sku/src/context/configSchema.ts
DanDroryAu: We can either do this or workaround type it by using `as unknown as Validator.default` for the validator type. It fixes typescript typing and it still acts like it should.

packages/sku/src/entry/server/server.ts
DanDroryAu: It throws up a typescript error telling me to declare the type.

```
TS2742: The inferred type of app cannot be named without a reference to
`.pnpm/@types+express-serve-static-core@4.19.6/ node_modules/@types/ express-serve-static-core`
. This is likely not portable. A type annotation is necessary.
```

Which is why I typed it.

packages/sku/src/types/global.d.ts
```diff
@@ -0,0 +1,30 @@
...
+type CSPHandler = {
+  registerScript: (script: string) => void;
+  createCSPTag: () => string;
+  handleHtml: (html: string) => string;
+};
+
+type SkuCSP = {
+  enabled: boolean;
+  extraHosts: string[];
+};
```
askoufis, line 30: IMO these types don't need to be global. Can they be defined (or potentially derived) inline where they're used, or at least defined in a non-global typescript file?

packages/sku/src/entry/makeExtractor.tsx
```diff
@@ -42,8 +39,14 @@ export default (stats, publicPath, csp) => {
...
-  return {
-    /** @type RenderCallbackParams['getHeadTags'] */
+  const extractorContext: Pick<
+    RenderCallbackParams,
+    | 'getHeadTags'
+    | 'flushHeadTags'
+    | 'getBodyTags'
+    | 'SkuProvider'
+    | 'extractor'
+  > = {
```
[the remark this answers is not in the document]
DanDroryAu: If I annotate it on the function as the return type the `extractorContext` functions aren't typed. I can define it as a separate type and use satisfies here.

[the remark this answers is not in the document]
DanDroryAu: It works if I directly return the extractorContext rather than defining a constant.

packages/sku/src/types/global.d.ts
DanDroryAu: I've moved these to the csp.ts file. The globals list does have a reference of it but I've imported it from the csp.ts file.

askoufis (approved):
LGTM, though CI is now failing.

DanDroryAu:
> LGTM, though CI is now failing.

This is because of the global.d.ts importing the type `SkuCSP`. I've moved that back since its not being used in the csp.js file. The other type moved.

PR #1146, opened 2025-01-31
add changeset for esm change

DanDroryAu:
Add changeset outlining the update to ESM and notifying users to update files.

### DIFF
--- a/.changeset/brown-teachers-teach.md
+++ b/.changeset/brown-teachers-teach.md
@@ -23,8 +23,4 @@ If you are using `require()` to import `sku/webpack-plugin` you will have to cha
 
 ___
 
-`sku/config/eslint` is a new entrypoint that exposes `sku`'s ESLint configuration.
-
-`sku@14` now uses `eslint` flat config and it will try to migrate your `.eslintignore` and `.eslintrc` files automatically.
-
-If a custom `eslint` configuration is required and you want to use the `sku/config/eslint` you will need to convert your `eslint.config.js` file to ESM.
+`sku/config/eslint` is a new entrypoint that exposes `sku`'s flat ESLint configuration. It now only supports ESM.
--- a/.changeset/brown-teachers-teach.md
+++ b/.changeset/brown-teachers-teach.md
@@ -6,20 +6,10 @@ Migrating `sku` to ESM from commonjs
 
 **BREAKING CHANGE**:
 
-`sku` now uses ESM modules instead of commonjs.
-
-If your project is not already using ESM modules you will have to make the following changes:
-
+Most of [`sku`s API entrypoints](https://seek-oss.github.io/sku/#/./docs/api) are now ESM. Consumers that use the following API entrypoints may need to convert some of their configuration files to ESM:
 ___
 
-`sku.config` files must now be written in TypeScript or ESM.
-
-If you are already using a `sku.config.ts` file no changes are required.
-If you are using a `sku.config.js` file you will either have to move to a `ts` or `mjs` file and update the `sku.config` accordingly.
-
-___
-
-`sku/webpack-plugin` is now an ESM module and has to be imported as such.
+`sku/webpack-plugin` is now an ES module and has to be imported as such.
 
 If you are using `require()` to import `sku/webpack-plugin` you will have to change it to `import`.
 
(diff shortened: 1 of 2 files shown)


.changeset/brown-teachers-teach.md
```diff
@@ -0,0 +1,11 @@
...
+'sku': major
+---
+
+Migrating `sku` to ESM from commonjs
+
+**BREAKING CHANGE**:
+
+`sku` now uses ESM modules instead of commonjs. Consumers will need to update their `sku.config.js` and any other configuration files that use `sku` to use ESM syntax.
+
+
```
askoufis, line 11: IMO this needs to be much more specific about what's happening and what's required for consumers.

E.g. most sku consumers are using `sku.config.ts` files, so won't need to do anything. Only consumers that use `js` _and_ CJS will need to do anything.

Additionally, we should have details for each entrypoint on what is/isn't required to migrate. While it seems a bit tedious, simply mentioning "any other configuration files that use sku`" will inevitably result in some consumers forgetting/ignoring a file and then asking support questions that could've been answered in the release notes.

DanDroryAu: After having tested the v14 snapshot on some projects I've got a much better understanding about whats required so I can certainly increase the specificity of the notes.

.changeset/brown-teachers-teach.md
```diff
@@ -0,0 +1,37 @@
...
+If your project is not already using ESM modules you will have to make the following changes:
+
+___
+
+`sku.config` files must now be written in TypeScript or ESM.
+
+If you are already using a `sku.config.ts` file no changes are required.
+If you are using a `sku.config.js` file you will either have to move to a `ts` or `mjs` file and update the `sku.config` accordingly.
+
```
askoufis, line 19: I don't want to give the impression that consumers need to do anything drastic like set `type: "module"` or something, which this sort of implies. Everyone consuming sku already _writes_ ESM (usually via TypeScript), so their application code shouldn't need to change at all.

I've also realised that we don't actually support loading `.mjs` files by default. I'll make a PR to add support for this (and refactor the config file resolution a bit too).

```suggestion
`sku` has been rewritten from CommonJS to ESM

**BREAKING CHANGE**:

Application code should be unaffected by this change. However, `sku.config` files are now loaded in an ESM context. This means they must either be authored in TypeScript (inside a `.ts` file) or in ESM (inside either a `.js` or `.mjs` file).

Additionally, most of [`sku`s API entrypoints](https://seek-oss.github.io/sku/#/./docs/api) are now ESM. Consumers that use the following API entrypoints may need to convert some of their configuration files to ESM:
```

.changeset/brown-teachers-teach.md
```diff
@@ -0,0 +1,37 @@
...
++ import SkuWebpackPlugin from 'sku/webpack-plugin';
+```
+
+> [!NOTE]
+> The file that is importing `sku/webpack-plugin` must also be an ESM module so further changes may be required.
+
+___
+
+`sku@14` uses `eslint` flat config and it will try to migrate your `.eslintignore` and `.eslintrc` files automatically.
+If a custom eslint configuration is required and you want to use the `sku/config/eslint` it will have to be ESM.
```
askoufis, line 37: Should we mention the `jest-preset` entrypoint too, even though there are barely any consumers of it?

.changeset/brown-teachers-teach.md
```diff
@@ -0,0 +1,37 @@
+---
+'sku': major
+---
+
```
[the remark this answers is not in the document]
DanDroryAu: The RLA Acronyms.

.changeset/brown-teachers-teach.md
DanDroryAu: I'll have a test of this in a package that consumes it. I couldn't read any information regarding the `preset` field and if it needs to be `ESM` or `commonjs`.

askoufis: The docs specify that any format is supported. Having dived into [the implementation](https://github.com/jestjs/jest/blob/68dca2a321b9887660d641c160ff6b94552c9dac/packages/jest-config/src/normalize.ts#L148) a few months ago for another sku-related reason, it seems that the preset module doesn't have to be the same format as config.

https://github.com/jestjs/jest/blob/bacb7de30d053cd87181294b0c8a8576632a8b02/packages/jest-util/src/requireOrImportModule.ts#L45-L74

If the preset module is `.mjs`, then a dynamic `import` is used. Otherwise `require` is used, and if that fails because it required a `.js` file that contained ESM, then it tries a dynamic `import` again. So I _think_ an ESM preset should work regardless of whether the jest config is CJS or ESM.

.changeset/brown-teachers-teach.md
```diff
@@ -0,0 +1,30 @@
...
+> [!NOTE]
+> The file that is importing `sku/webpack-plugin` must also be an ES module so further changes may be required.
+
+___
+
+`sku/config/eslint` is a new entrypoint that exposes `sku`'s ESLint configuration.
+
+`sku@14` now uses `eslint` flat config and it will try to migrate your `.eslintignore` and `.eslintrc` files automatically.
+
+If a custom `eslint` configuration is required and you want to use the `sku/config/eslint` you will need to convert your `eslint.config.js` file to ESM.
```
askoufis, line 30: The migration is already mentioned in a separate changeset, so not sure if needs mentioning here. Additionally, we don't want to confuse consumers on how to set up custom eslint config. If they're using sku, they should use `dangerouslySetESLintConfig`. `sku/config/eslint` should only be used by consumers that are just using plain `eslint` rather than `sku lint`.

```suggestion
`sku/config/eslint` is a new entrypoint that exposes `sku`'s flat ESLint configuration. It now only supports ESM.
```

askoufis (approved):

PR #1178, opened 2025-02-13
add createViteConfig function to the vite service.

DanDroryAu:
Add the `createViteConfig` function to the `vite` service folder.

This is the boilerplate configuration for all inline vite sdk calls in the service.


packages/sku/src/services/vite/createConfig.ts
```diff
@@ -0,0 +1,78 @@
+import { builtinModules } from 'node:module';
+import type { InlineConfig } from 'vite';
+
+import react from '@vitejs/plugin-react';
```
askoufis, line 4: Can we use `@vitejs/plugin-react-swc` instead? Appreciate I might be throwing a spanner in the works if you've been testing with the regular react plugin this whole time, but we already have a dependency on SWC, so we might as well lean into that and get the perf benefits from using it for react too.

askoufis (commented):
I know I said previously to split PRs up, but this is almost too isolated and granular. E.g., what's `VITE_SKU_CONTEXT` used for? I can't know this until you raise a subsequent PR that uses it, at which point I have to refer to an already-merged PR (this PR) to connect the dots.

Feel like I'm being petty here but it matters how PRs are split up, not just that they _are_ split up.

PR #1179, opened 2025-02-13
Vite start command

DanDroryAu:
Add `sku start` command for experimental `vite` support. 

This can be tested in the new `vite-text-app` fixture.


packages/sku/src/services/vite/createConfig.ts
```diff
@@ -0,0 +1,68 @@
...
+    ],
+    resolve: {
+      alias: {
+        __sku_alias__clientEntry: skuContext.paths.clientEntry,
+        __sku_alias__serverEntry: skuContext.paths.serverEntry,
+        __sku_alias__renderEntry: skuContext.paths.renderEntry,
+      },
+    },
+    define: {
+      'process.env': process.env,
```
askoufis, line 44: Vite already has [its own envar conventions](https://vite.dev/guide/env-and-mode.html#env-variables-and-modes), so IMO the only envar we should expose here is `process.env.NODE_ENV` since braid relies on it (and it's documented sku feature too).

packages/sku/src/services/vite/createConfig.ts
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  return {
+    root: process.cwd(),
+    plugins: [
+      react(),
+      vanillaExtractPlugin({
+        unstable_pluginFilter: (plugin) =>
+          !plugin.name.startsWith('vite:react-'),
+      }),
```
askoufis, line 33: As per my comment on your previous PR #1178, this shouldn't be necessary as the VE plugin no longer passes through any plugins by default.

packages/sku/src/services/vite/plugins/skuViteMiddlewarePlugin.ts
```diff
@@ -0,0 +1,43 @@
+import type { SkuContext } from '@/context/createSkuContext.js';
+import type { Plugin } from 'vite';
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+
+export const skuViteMiddlewarePlugin = (skuContext: SkuContext): Plugin => ({
+  name: 'vite-plugin-sku-server-middleware',
+  configureServer(server) {
+    return () => {
```
askoufis, line 10: `configureServer` can return `void`, so returning a function isn't necessary here.

packages/sku/src/services/vite/plugins/skuViteMiddlewarePlugin.ts
```diff
@@ -0,0 +1,43 @@
...
+          skuContext.sites.find((skuSite) => skuSite.host === hostname) || '';
+
+        const isHtml = req.url === '/index.html';
+        if (isHtml) {
+          const renderEntry = require.resolve('../entries/vite-render.jsx');
+          const clientEntry = require.resolve('../entries/vite-client.jsx');
+
+          const { viteRender } = await server.ssrLoadModule(renderEntry);
+
+          const renderedHtml = await viteRender({
```
askoufis, line 24: `viteRender` has type `any` here. I think this is a spot where a typecast would prevent headaches. Having a centralized `Render` type would help keep these in sync.

packages/sku/src/services/vite/createConfig.ts
```diff
@@ -0,0 +1,68 @@
+import { createRequire, builtinModules } from 'node:module';
+import type { InlineConfig } from 'vite';
+
+import react from '@vitejs/plugin-react';
```
askoufis, line 4: Any reason not to use `@vitejs/plugin-react-swc` here instead?

packages/sku/src/services/vite/createPreRenderedHtml.ts
askoufis, line 1: I can't get linting to apply to this file for some reason.

askoufis:
What are your thoughts on just putting this in `master` instead of a feature branch? I'm hesitant to make any non-trivial changes to `master` when I know there's a large feature branch in the works, so would honestly just prefer we release the feature, call it experimental, and iterate on it over time.

packages/sku/src/services/vite/plugins/skuViteMiddlewarePlugin.ts
DanDroryAu: https://vite.dev/guide/api-plugin.html#configureserver

This states that by returning a function I ensure that my middleware is run after the internal middlewares. This makes it easy for this middleware to understand which specific call returns the `index.html` file since it will be configured then.

DanDroryAu:
> What are your thoughts on just putting this in `master` instead of a feature branch? I'm hesitant to make any non-trivial changes to `master` when I know there's a large feature branch in the works, so would honestly just prefer we release (or don't, doesn't really matter that much) the feature, call it experimental, and iterate on it over time.

We can do that, I do feel like everything should be a minor bump though because of new functionality. We have to specify that any future changes to the experimental work can be breaking without any major version bump.

packages/sku/src/services/vite/createViteServer.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  return await createServer({
+    ...createViteConfig({
+      skuContext,
+      plugins: [skuViteMiddlewarePlugin(skuContext)],
+    }),
+    server: {
+      allowedHosts: skuContext.sites
+        .map((site) => site.host ?? false)
+        .filter((host) => typeof host === 'string'),
```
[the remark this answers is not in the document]
DanDroryAu: If the object of `site` is this:

```
site = {
   host: 'HostName',
};
```

Wouldn't this just filter everything out?

`host` wont ever be of type `string` without the `map` first trying to turn the array into a `Array<string>`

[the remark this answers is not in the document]
DanDroryAu: If empty strings should be removed the best thing would be to change the `??` into `||`. The filter checking for type of string ensures the returned array is of type `string[]` which is what the config expects.

I can destructure over dot chain in the map.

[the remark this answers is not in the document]
askoufis: Right, ignore me. Forgot that `site` can be a `string` or an object.

packages/sku/src/services/vite/createViteServer.ts
```diff
@@ -0,0 +1,23 @@
+import { skuViteMiddlewarePlugin } from '@/services/vite/plugins/skuViteMiddlewarePlugin.js';
+import { createViteConfig } from '@/services/vite/createConfig.js';
+import type { SkuContext } from '@/context/createSkuContext.js';
+import { createServer } from 'vite';
+
+export const createViteServer = async (skuContext: SkuContext) => {
+  const base = process.env.BASE || '/';
+  const root = process.cwd();
```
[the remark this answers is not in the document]
DanDroryAu: `base` is the public path where all content gets served from. In the sku config it's the root of the static assets. When checking `build` `start` or `serve` in a webpack setting the only thing that changes is the path to the static assets in the index.html. 

So I don't believe this is supposed to go into the base of the application because the effect becomes that the application gets hosted on ``${hostname}/#{base}``

I'll remove root.

askoufis (approved):

PR #1180, opened 2025-02-17
Vite serve ssr command

DanDroryAu:
Add `start-ssr` command to the `vite` program.

This can be tested in the `vite-test-app` fixture. 

`start-ssr` is slightly different from `start` in that it will go through an express server rather than the Vite server itself.

`start` should only create the `index.html` for the initial requested route to mimic pre-render behaviour (which is still to come).


fixtures/vite-test-app/src/server.jsx
```diff
@@ -0,0 +1,17 @@
...
+    console.log('SSR rendered', renderContext);
+
+    return renderToPipeableStream(
+      <StrictMode>
+        <App />
+      </StrictMode>,
+      options,
+    );
+  },
+};
```
askoufis, line 17: Should make a type for the Vite server entry so we can add a `satisfies ViteServerEntry` here.

askoufis (commented):
Just a single comment about adding a type for the server entrypoint.

Additionally, I'm honestly not convinced that the `@` path alias is providing value to the codebase. A fair few import paths would be shorter without it, especially in this PR.

IMO sku's package structure doesn't have that many deeply nested branches to make it worthwhile having an alias to the root.

DanDroryAu:
> Additionally, I'm honestly not convinced that the `@` path alias is providing value to the codebase. A fair few import paths would be shorter without it, especially in this PR.
> 
> IMO sku's package structure doesn't have that many deeply nested branches to make it worthwhile having an alias to the root.

https://www.npmjs.com/package/eslint-plugin-no-relative-import-paths

Would this plugin with a relative depth of 1-2 satisfy that?

fixtures/vite-test-app/src/server.jsx
DanDroryAu: The file is currently `jsx`. I can convert it to `tsx` and do both entries.

askoufis: > The file is currently jsx

Oh, I missed that 😅. We should probably convert all the fixture files to ts/tsx, but yeah can start with these ones so we can define and use the entrypoint types.

askoufis:
> > Additionally, I'm honestly not convinced that the `@` path alias is providing value to the codebase. A fair few import paths would be shorter without it, especially in this PR.
> > IMO sku's package structure doesn't have that many deeply nested branches to make it worthwhile having an alias to the root.
> 
> https://www.npmjs.com/package/eslint-plugin-no-relative-import-paths
> 
> Would this plugin with a relative depth of 1-2 satisfy that?

Maybe? What problem are we trying to solve with the path alias and/or the eslint plugin?

I'm happy to persist with the alias, with the eslint plugin too if we want, but I feel like they're made to address issues in much larger repos with much deeper folder structures, both of which don't seem to be the case in `sku`.

Are long import paths that big of a deal? Something like `../../../../../foo.js` becoming `@/foo.js` seems beneficial in isolation, but the tradeoff is a short relative import like `../bar.js` might become `@/some/really/long/path/to/bar.js`.

Path aliases feel more useful in front-end repos that have a flat `components` directory and separate `App` directory that contains pages/layouts. Or maybe a separate utils directory. Aliasing the root src directory just means there's now 2 different ways to do the same thing, with negligible benefit to either.

askoufis (approved):

PR #1119, opened 2024-12-03
Add support for WebP and AVIF image formats

adenj:
- Added support for webp and avif image formats.
- Updated documentation to reflect the supported image types: bmp, gif, jpeg, png, svg, webp, and avif.
- Added a note on browser support for webp and avif formats, including a fallback strategy using the `<picture>` element.

PR #1181, opened 2025-02-18
Fix `webp` and `avif` support, test asset loading in fixture

askoufis:
Looks like some changes from https://github.com/seek-oss/sku/pull/1119 got missed when the change was merged into the v14 branch, so webpack currently fails to load `webp` and `avif` images.

I've added the missing changes, as well `avif` support for the jest file mock. Additionally, each supported image format is now tested within the `public-path` fixture.


jahredhope (approved):

packages/sku/src/config/jest/jest-preset.js
```diff
@@ -27,7 +27,7 @@ export default jestDecorator({
   ],
   moduleFileExtensions: ['js', 'json', 'ts', 'tsx'],
   moduleNameMapper: {
-    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga|svg)$':
+    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga|svg|avif)$':
```
DanDroryAu, line 30: **Question:** Should `bmp` be in here as well?

Is it ok to load a `bmp` without a filemock?

askoufis: Oh, didn't notice that it was missing. Automerge beat you to it. I'll make another PR.

DanDroryAu: I'll ✅ that when it shows up

PR #1182, opened 2025-02-18
Escape stray appostraphe in app template

askoufis:
All other apostraphes in the template are escaped as `&rsquo;`.


packages/sku/template/src/App/NextSteps.tsx
```diff
@@ -113,7 +113,7 @@ export function NextSteps({ environment }: NextStepsProps) {
                 <Stack space="gutter">
                   <Heading level="3">Brand resources</Heading>
                   <Text>
-                    To align your project to SEEK's brand standards it is
+                    To align your project to SEEK&rsquo;s brand standards it is
```
DanDroryAu, line 116: His ninja days are done 🥷.

DanDroryAu (approved):

PR #1183, opened 2025-02-18
add vite `build` command

DanDroryAu:
Add `build` command to the set of vite experimental commands. 

`build` will use the `routes` object in the `sku.config.ts` to build out the static sites for those routes. 

This can be tested by running `build` in the `vite-test-app` fixture and running `serve` afterwards to test it. `serve` is still using the original serve method from `sku`.

This does not yet support full lazy loading which will come in a future pr.


packages/sku/src/services/vite/preload/collector.ts
```diff
@@ -0,0 +1,245 @@
...
+  preloads: Map<string, Preload>;
+  chunk: string;
+  nonce?: string;
+}) => {
+  const ext = extname(chunk).substring(1);
+  let as;
+  let mimeType;
+
+  switch (ext) {
+    case 'png':
```
askoufis, line 194: This will need to support any image formats sku supports. Looks like `bmp` and `avif` are missing, and worth adding `jpeg` too.

askoufis (commented):
I can't load the `nz` site after running `sku build && sku serve`. Looks like the site path is nested an extra level for some reason.

image

Haven't had a chance to take a good look at the preloading implementation just yet. Or is it worth wait until full lazy loading is supported in a future PR?

DanDroryAu:
> I can't load the nz site after running sku build && sku serve. Looks like the site path is nested an extra level for some reason.
> Haven't had a chance to take a good look at the preloading implementation just yet. Or is it worth wait until full lazy loading is supported in a future PR?

This is the same behavior as the multiple-routes fixture. It does the same thing so maybe we need to have a look at both of those.

askoufis:
> > I can't load the nz site after running sku build && sku serve. Looks like the site path is nested an extra level for some reason.
> > Haven't had a chance to take a good look at the preloading implementation just yet. Or is it worth wait until full lazy loading is supported in a future PR?
> 
> This is the same behavior as the multiple-routes fixture. It does the same thing so maybe we need to have a look at both of those.
> 
> Also, currently the collector only gets all of the files from the manifest and doesn't do anything with lazy loading yet. The CollectorContext and plugin are still missing. I'm only using it to inject the correct scripts in the index.html from manifest.
> 
> I'll add lazy loading and the corresponding plugin in as a separate pr.

NVM, that's just how that fixture (and the vite one) are configured. On the NZ domain `/` doesn't work, you have to go to `/nz`. 🤦

askoufis (approved):

PR #1184, opened 2025-02-20
Vite lazy load plugin

DanDroryAu:
BLOCKED BY: https://github.com/seek-oss/sku/pull/1183

Add vite preload

Vite preload has a few components. 

- The `loadable` function that wraps a component in a lazy load that we can use to pre-load it before we generate render the app (thus skipping suspense boundaries).
- The `Collector` class which gathers all imported component scripts and files that need to be imported to render the page and returns the correct tags.
- The `LoadableProvider` component which we use to register extra lazy loaded component dependencies.
- The `preloadAll` function which pre-loads all lazy components before we call a render function.

This is meant to replace/substitute `@loadable` for `vite` applications.

Some notes: The implementation in the plugin is using babel to parse the AST and inject the correct module Id into the hook to register the component. This code can likely be simplified.


packages/sku/src/services/vite/plugins/skuVitePreloadPlugin.ts
```diff
@@ -0,0 +1,150 @@
...
+      }
+      let ast: ReturnType<typeof parse>;
+
+      // TODO: Do some better checking. Not every function thats called `loadable` is the correct one.
+      // TODO: Make sure that the import is correct as well to failsafe.
+      // Find dynamic imports
+      if (code.includes(' import(')) {
+        ast = parse(code, {
+          sourceType: 'module',
+          plugins: ['jsx', 'typescript'],
```
askoufis, line 69: Is there a reason to use plugins over presets here? We already declare dependencies on the react and typescript presets, so I'd expect to just re-use those.

Additionally, similar to what we do [sku's babel config](https://github.com/seek-oss/sku/blob/94508aaf8b3ea014b2a8d22ad1be0f1accc6912f/packages/sku/src/config/babel/babelConfig.ts), it would be better to `require.resolve` the path to the preset/plugin rather than rely on babel to resolve them I we've had issues in the past when just passing plugins as strings.

packages/sku/src/services/vite/plugins/skuVitePreloadPlugin.ts
```diff
@@ -0,0 +1,150 @@
...
+          };
+        }
+      }
+
+      return null;
+    },
+
+    buildEnd() {
+      if (debug) {
+        const s = lazyImportedModules.difference(injectedModules);
```
askoufis, line 142: Worth noting that `Set.difference` support was added [in Node 22](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference#browser_compatibility), which is above sku's current minimum supported node version. That being said, we could also just require node v22 when using sku's vite mode. It's experimental, so we can do whatever we want 😎.

packages/sku/src/services/vite/plugins/skuVitePreloadPlugin.ts
DanDroryAu: I'm not to familiar with all the ways you can parse AST so I don't have a particular reason. If you've got some tips on why one thing works over the other that would be helpful

packages/sku/src/services/vite/preload/PreloadContext.tsx
```diff
@@ -0,0 +1,12 @@
+import { createContext, useContext } from 'react';
+import type { Collector, ModuleId } from './collector.js';
+
+const context = createContext<null | Collector>(null);
+
+export const LoadableProvider = context.Provider;
+
+export const useRegisterComponent = (moduleId: ModuleId) => {
+  const components = useContext(context);
+  if (!components) return;
```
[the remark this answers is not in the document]
DanDroryAu: I have to make it a bit more complex because the `client.tsx` files are usually not wrapped inside a loadable component which would then error when it's hydrating.

I am going to rename it to `collector` though because that's what it returns and components is a bit confusing.

packages/sku/src/services/vite/plugins/skuVitePreloadPlugin.ts
askoufis: I think the issue is related to PNPM not hoisting dependencies, which I think Babel doesn't account for if it tries to resolve a plugin/preset string, hence why we've used `require.resolve` in the past.

There's not an AST-specific reason to use plugins over presets. Presets are simply collections of plugins. In this case, I guess since you're only parsing JSX and TS and don't need to transform it (e.g. adding the jsx runtime), then the syntax plugins are likely all you need. It would still be a good idea to declare dependencies on [`@babel/plugin-syntax-typescript`](https://www.npmjs.com/package/@babel/plugin-syntax-typescript) and [`@babel/plugin-syntax-jsx`](https://www.npmjs.com/package/@babel/plugin-syntax-jsx) though.

Worth also considering that we can likely get an already-parsed AST from rollup via [`this.getModuleInfo`](https://rollupjs.org/plugin-development/#this-getmoduleinfo). The tradeoff would be that the AST is in a different format (specifically ESTree), which isn't easily compatible with babel's AST format, so would require a few dependencies to work with. But I'd only consider this if this transformation becomes a performance issue.

packages/sku/package.json
```diff
@@ -72,12 +80,16 @@
   },
   "dependencies": {
     "@babel/core": "^7.21.8",
+    "@babel/generator": "^7.26.9",
+    "@babel/plugin-syntax-jsx": "^7.25.9",
+    "@babel/plugin-syntax-typescript": "^7.25.9",
```
DanDroryAu, line 85: need to remove

packages/sku/src/services/vite/loadable/PreloadContext.tsx
```diff
@@ -0,0 +1,19 @@
...
+const context = createContext<null | Collector>(null);
+
+export const LoadableProvider = context.Provider;
+
+export const useRegisterComponent = (moduleId: ModuleId) => {
+  const collector = useContext(context);
+  if (!collector) {
+    if (import.meta.env.SSR) {
+      throw new Error(
+        'useRegisterComponent must be used inside LoadableProvider',
```
DanDroryAu, line 13: `useRegisterComponent` is hidden so I should rename this to `loadable` because thats more apparent to the end user

packages/sku/src/types/types.ts
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line
+/// <reference types="vite/client" />
```
askoufis, line 2: Should this be removed?

DanDroryAu: Yeah, oversight. `client.d.ts` is being loaded in already.

askoufis (approved):

PR #1186, opened 2025-02-21
add esbuild plugin for vanilla-extract css loading issue.

DanDroryAu:
Add a new `fixViteVanillaExtractDepScanPlugin` to the vite config. 

This plugin stops an issue where esbuild changes the order of import for vanilla-extract files.


packages/sku/package.json
```diff
@@ -123,6 +132,7 @@
     "didyoumean2": "^7.0.2",
     "ensure-gitignore": "^1.1.2",
     "env-ci": "^11.1.0",
+    "esbuild": "^0.24.2",
```
askoufis, line 135: Since we're only using `esbuild` for the `Plugin` type, can this be moved to a dev dependency?

askoufis (approved):

PR #1189, opened 2025-02-24
Vite experimental flag and changeset

DanDroryAu:
This is blocked by previous PR's going into the experimental branch.

The work that can be reviewed in this pr:
- Changeset
- experimentalBundler.option.ts
- the build/start/start-ssr `action.ts` and `command.ts` files.

Add changeset and flag for experimental-bundlers


packages/sku/src/program/commands/start/start.action.ts
```diff
@@ -8,14 +8,21 @@ export const startAction = async (
...
   },
   command: Command,
 ) => {
   const environment = command.parent?.opts()?.environment;
-  if (skuContext.bundler === 'vite') {
+  if (skuContext.bundler === 'vite' && !experimentalBundler) {
+    throw new Error(
+      'The `vite` bundler is experimental. If you want to use it please use the `--experimental-bundler` flag.',
+    );
+  }
```
askoufis, line 24: It feels a bit tedious, but should we also have a check for the inverse of this? I.e. `--experimental-bundler` is set but `bundler: 'vite'` is not set in sku config?

.changeset/flat-papayas-suffer.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Migrating your render entrypoint.
+
+The interface for the render entrypoint has seen some more significant changes.
+
+- The `renderToString` function has been replaced with `renderToPipeableStream`.
+- The `renderApp` function has been renamed to `render`.
+- The `SkuProvider` component has been removed from the `renderApp` function. Instead, you have to use the `LoadableProvider` from the `sku/vite/loadable` export.
+- The `preloadAll` function has been added. Awaiting this function ensures that all `loadable` imported files will be preloaded before sku renders the page.
+- The `renderDocument` function has been removed. It's functionality will be added in a future update.
```
askoufis, line 57:
```suggestion
- The `renderDocument` function has been removed. Its functionality will be added in a future update.
```

.changeset/flat-papayas-suffer.md
```diff
@@ -0,0 +1,171 @@
...
+export default {
+  bundler: 'vite', // sku will use vite as the bundler.
+  ...
+} satisfies SkuConfig;
+```
+
+### Migrating your client entrypoint
+
+With `vite` the `index.html` will be generated with a different id for the root element.
+Make sure that your client entrypoint exports a function that hydrates the root element as the default export.
```
askoufis, line 30: I assume this won't be necessary once `renderDocument` is supported in vite. Maybe worth noting that this is a temporary change.

.changeset/flat-papayas-suffer.md
```diff
@@ -0,0 +1,171 @@
...
+};
+```
+
+### Migrating your render entrypoint.
+
+The interface for the render entrypoint has seen some more significant changes.
+
+- The `renderToString` function has been replaced with `renderToPipeableStream`.
+- The `renderApp` function has been renamed to `render`.
+- The `SkuProvider` component has been removed from the `renderApp` function. Instead, you have to use the `LoadableProvider` from the `sku/vite/loadable` export.
```
askoufis, line 55: Is `LoadableProvider` always necessary, or is it only useful if you're using `loadable`?

.changeset/flat-papayas-suffer.md
```diff
@@ -0,0 +1,171 @@
...
+-      </html>
+-    `;
+-  },
+- } satisfies Render;
++ } satisfies ViteRender;
+```
+
+### Migrating from `sku/@loadable/components` to `sku/vite/loadable`
+
+The `sku/@loadable/components` package has been replaced with `sku/vite/loadable`.
```
askoufis, line 122:
```suggestion
The `sku/@loadable/components` entrypoint has been replaced with `sku/vite/loadable`.
```

.changeset/flat-papayas-suffer.md
```diff
@@ -0,0 +1,171 @@
...
+
+The new `sku/vite/loadable` package relies on the `react <Suspense>` component for the loading fallback.
```
askoufis, line 124:
```suggestion
The new `sku/vite/loadable` entrypoint relies on the React's [`<Suspense>`](https://react.dev/reference/react/Suspense) component for the loading fallback.
```

.changeset/flat-papayas-suffer.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Using vite types
+
+If `vite/client` reference types are needed in your project, you can import them from `sku/vite/client`.
+
+```typescript
+/// <reference types="sku/vite/client" />
+
+// ... rest of your application
+````
```
askoufis, line 153:
````suggestion
```
````

File: .changeset/flat-papayas-suffer.md (hunk 4 above)
Comment (DanDroryAu) [reply .changeset/flat-papayas-suffer.md:55]: Only needed if you use `loadable` which is what the SkuProvider did I believe? I'll mention it more explicitly.

File: packages/sku/src/program/commands/start/start.action.ts (hunk 1 above)
Comment (DanDroryAu) [reply packages/sku/src/program/commands/start/start.action.ts:24]: I'll pop this in a separate commander.js hook. This will muddy the action handlers otherwise and then we can just remove the hook and option when it's time to do so.

File: packages/sku/src/program/commands/serve/serve.command.ts [hunk 8]
```diff
@@ -12,6 +12,6 @@ serveCommand
   .addOption(siteOption)
   .addOption(portOption)
   .action(async ({ site, port, skuContext }, command) => {
-    const environment = command.parent.opts()?.environment;
+    const { environment } = command.optsWithGlobals();
```
DanDroryAu, line 15: Small little cleanup I've added in after reading about options. This way we don't have to go up multiple commands to find the correct `parent` that has the right options. Globals get merged in with other local options and since `--environment` is global we can get it from this function call.

askoufis (approved):

PR #1192, opened 2025-02-27
Experimental vite support

DanDroryAu:


askoufis:
Unsure if I'm doing something wrong, but running `sku start --experimental-bundler` in the `vite-test-app` fixture works, but accessing `http://dev.seek.com.au:8303/` does not. The browser doesn't even connect to anything, I don't even get a 404 response. `build` works fine though 🤔.

askoufis (approved):

PR #1200, opened 2025-02-28
add hmr telemetry plugin

DanDroryAu:
Add plugin for HMR telemetry.

This may be a potential candidate for all the telemetry in vite in future.


packages/sku/src/services/vite/plugins/skuViteHMRTelemetry.ts
```diff
@@ -0,0 +1,45 @@
+import type { Plugin } from 'vite';
+import provider from '@/services/telemetry/index.js';
+import debug from 'debug';
+
+const log = debug('sku:metrics');
```
jahredhope, line 5: [Elsewhere](https://github.com/seek-oss/sku/blob/bed85cdf3133dd5750982df2b536b5e66054c883/packages/sku/src/services/telemetry/index.ts#L11) we use `sku:telemetry` would this be better grouped together?

jahredhope: Nevermind, didn't realise we already use `sku:metrics` [in the webpack plugin](https://github.com/seek-oss/sku/blob/bed85cdf3133dd5750982df2b536b5e66054c883/packages/sku/src/services/telemetry/index.ts#L11).

Seems less than ideal as I'd think you'd want to see them both together, but not worth changing in this PR

jahredhope (approved):

PR #1203, opened 2025-03-11
`vite build (experimental)`: allowing specific cjs package interop and support publicPath values

williamlark:
**Allowing specific cjs package interop.**

The Vite compiler enforces ESM modules to be used, so any CJS dependencies will fail the build command. To make testing the Vite compiler easier, we have allowed some cjs libraries to be interpreted as modules. For this release, `@apollo/client` and `lodash` have been added.

**Support publicPath values**

The Vite compiler did not consider the configured `publicPath` in the sku config. Now it does for absolute and relative paths (e.g., _assets_, _/assets_, _./assets_). URLs are still unsupported.

### DIFF
--- a/packages/sku/src/services/vite/helpers/createConfig.ts
+++ b/packages/sku/src/services/vite/helpers/createConfig.ts
@@ -2,6 +2,7 @@ import { createRequire, builtinModules } from 'node:module';
 import type { InlineConfig } from 'vite';
 
 import react from '@vitejs/plugin-react-swc';
+import { cjsInterop } from 'vite-plugin-cjs-interop';
 import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';
 
 import type { SkuContext } from '@/context/createSkuContext.js';
@@ -36,6 +37,9 @@ export const createViteConfig = ({
   return {
     root: process.cwd(),
     plugins: [
+      cjsInterop({
+        dependencies: ['@apollo/client', 'lodash'],
+      }),
       react(),
       vanillaExtractPlugin(),
       skuVitePreloadPlugin(),
--- a/packages/sku/src/services/vite/helpers/createConfig.ts
+++ b/packages/sku/src/services/vite/helpers/createConfig.ts
@@ -61,6 +61,7 @@ export const createViteConfig = ({
       ssr: configType === 'ssr' || configType === 'ssg',
       manifest: configType === 'client',
       ssrManifest: false,
+      assetsDir: skuContext.publicPath.replace(/(^\/|\/$)/g, ''),
       rollupOptions: {
         input: input[configType],
         output: {
--- a/packages/sku/src/services/vite/helpers/createConfig.ts
+++ b/packages/sku/src/services/vite/helpers/createConfig.ts
@@ -35,6 +35,7 @@ export const createViteConfig = ({
   };
 
   return {
+    base: './',
     root: process.cwd(),
     plugins: [
       cjsInterop({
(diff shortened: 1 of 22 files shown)


packages/sku/src/services/vite/helpers/html/createPreRenderedHtml.ts
```diff
@@ -28,6 +28,7 @@ export function createPreRenderedHtml({
       clientContext = await render.provideClientContext({
         site,
         url,
+        environment: process.env.NODE_ENV,
```
askoufis, line 31: This isn't where `environment` should come from. This environment is actually the [sku environment](https://seek-oss.github.io/sku/#/./docs/configuration?id=environments) the page is being rendered in. I don't think our vite implementation supports environments just yet, so would remove this until we have proper support for pre-rendering multiple environments.

packages/sku/src/services/vite/helpers/prerenderRoutes.ts
```diff
@@ -17,6 +17,8 @@ export const prerenderRoutes = async (skuContext: SkuContext) => {
     const render = (await import(resolve('./dist/render/render.js'))).default;
     const loadableCollector = createCollector({
       manifest,
+      // ? We could set this as a SKU config. Hardcoding for now to force relative paths.
+      base: './',
```
askoufis, line 21: At this point you might as well take this value from the loaded sku config. It's `skuContext.publicPath`.

packages/sku/src/services/vite/helpers/createConfig.ts
```diff
@@ -57,6 +61,7 @@ export const createViteConfig = ({
       ssr: configType === 'ssr' || configType === 'ssg',
       manifest: configType === 'client',
       ssrManifest: false,
+      assetsDir: skuContext.publicPath.replace(/(^\/|\/$)/g, ''),
```
askoufis, line 64: The leading slash is important here as it distinguishes between a root-relative path and a document-relative path.

```suggestion
      assetsDir: skuContext.publicPath.replace(/\/$/, ''),
```

packages/sku/src/services/vite/loadable/helpers/preloadUtils.ts
```diff
@@ -14,14 +14,14 @@ const getNonce = (nonce?: string) => (nonce ? ` nonce="${nonce}"` : '');
 
 const tagTypes = {
   stylesheet: ({ href, nonce }: Preload) =>
-    `<link rel="stylesheet" href="/${href}" crossorigin${getNonce(nonce)} />`,
+    `<link rel="stylesheet" href="./${href}" crossorigin${getNonce(nonce)} />`,
```
askoufis, line 17: This is sort of replacing one problem with another. Now any href is always relative.

We'd ideally support any type of path, e.g. all of these should work: `/static/foo`, `./`, `/`. I think with the suggestion in my other comment, combined with making all of these just `href="${href}", any `publicPath` to work.

askoufis:
This change needs a changeset.

packages/sku/src/services/vite/helpers/createConfig.ts
williamlark: I added the forward slash removal because Vite doesn't support leading slashes in the assetDir `patterns can be neither absolute nor relative paths. If you want your files to be stored in a subdirectory, write its name without a leading slash like this: subdirectory/pattern.`.

askoufis (approved):

PR #1205, opened 2025-03-13
Fix deployment

askoufis:
My change in https://github.com/seek-oss/sku/pull/1202 resulted in the URL used for publishing the docs site to github pages being incorrect, resulting in [this error](https://github.com/seek-oss/sku/actions/runs/13827265072/job/38684667294#step:7:33). I've modified the deployment script to account for this change.

Have also updated the `gh-pages` dep as it was quite out of date, and converted the deploy script to ESM.

### DIFF
--- a/docs/deploy-docs.js
+++ b/docs/deploy-docs.js
@@ -1,6 +1,10 @@
-const ghpages = require('gh-pages');
+import { publish } from 'gh-pages';
 
-const repoUrl = require('../package.json').repository.url;
+import packageJson from '../package.json' with { type: 'json' };
+
+const {
+  repository: { url: repoUrl },
+} = packageJson;
 
 const GITHUB_TOKEN = process.env.GITHUB_TOKEN;
 const tokenRegex = GITHUB_TOKEN ? new RegExp(GITHUB_TOKEN, 'g') : null;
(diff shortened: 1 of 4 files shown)


docs/deploy-docs.js
```diff
@@ -11,17 +15,15 @@ const log = function (message) {
...
+const publishConfig = {
+  repo: GITHUB_TOKEN
+    ? repoUrl.replace('git+https://', `https://${GITHUB_TOKEN}@`)
+    : repoUrl,
+  logger: log,
+  dotfiles: true,
 };
 
-ghpages.publish('.', makeConfig(), function (err) {
+await publish('.', publishConfig, function (err) {
```
askoufis, line 26: This function returns a promise now.

docs/deploy-docs.js
```diff
@@ -11,17 +15,15 @@ const log = function (message) {
...
 
-const makeConfig = function () {
-  return {
-    repo: GITHUB_TOKEN
-      ? repoUrl.replace('https://', `https://${GITHUB_TOKEN}@`)
-      : repoUrl,
-    logger: log,
-    dotfiles: true,
-  };
+const publishConfig = {
```
askoufis, line 18: No good reason this was a function, so it's now an object.

jahredhope (approved):

PR #1211, opened 2025-03-17
`vite`: Emit more accurate HMR metric

askoufis:
The current metric being sent by timing the elapsed period between two `handleHotUpdate` hooks is emitting HMR metrics that are sub 1ms, e.g. `0.05`. Not only is this suspiciously fast relative to [Vite's quoted HMR time of ~50ms](https://vite.dev/guide/features), but our internal telemetry provider rounds this value to `0`, which is unhelpful.

This plugin aims to implement [this suggestion](https://github.com/vitejs/vite/issues/14186#issuecomment-1722492373) by listening on hot updates in the client and then sending an event to the server so it can send off the telemetry. The downside to this approach is that multiple clients (i.e. browser tabs) may be connected to the server, so multiple events will be received by the server. This is simple enough to dedupe, so we will only ever send one HMR update metric no matter how many clients are connected.


packages/sku/src/services/vite/plugins/skuViteHMRTelemetry.ts
```diff
@@ -1,45 +1,61 @@
...
             type,
           });
-        },
+        });
       },
     },
   ];
 }
+
+const skuHmrTelemetryClient = dedent/* html */ `
```
askoufis, line 49: I think it's possible to point the `src` attribute of the injected `script` to a local file, and vite will inject it into the client. However, I like the idea of having everything related to a plugin in a single file, especially for relatively small plugins like this.

.changeset/famous-clocks-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+`vite`: Emit more accurate HMR metric
```
jahredhope, line 5: Changeset could include more details of the value "Record metrics as observed by browser"

askoufis: https://github.com/seek-oss/sku/pull/1211/commits/8d7e7308a3468e4a9d4add5065d2dac648ffa4cc

askoufis:
@jahredhope This plugin is only used when creating the dev server, so it's already scoped to dev-time and does not get included in a `sku build`.

.changeset/famous-clocks-rule.md
```diff
@@ -0,0 +1,7 @@
+---
+'sku': patch
+---
+
+`vite`: Emit more accurate HMR rebuild timing metrics
+
+Previously this metric was captured server-side by relying on plugin hooks firing in the correct order.This proved to be inaccurate, so we now capture this metric client-side and send the data back to the dev server.
```
DanDroryAu, line 7:
```suggestion
Previously this metric was captured server-side by relying on plugin hooks firing in the correct order. This proved to be inaccurate, so we now capture this metric client-side and send the data back to the dev server.
```

Small nitpick (forgot the space)

packages/sku/src/services/vite/plugins/skuViteHMRTelemetry.ts
```diff
@@ -1,45 +1,74 @@
...
-          startTime = performance.now();
+}): Plugin {
+  const hmrUpdateTimestamps = new Set();
+
+  return {
+    name: 'vite-plugin-sku-hmr-telemetry',
+    transformIndexHtml: {
+      // Vite needs to process the script in order to pick up the HMR context
+      order: 'pre',
+      handler: () => [
```
DanDroryAu, line 27: **praise:** nice solution by adding this into the head tag. No need to bother the client with this.

DanDroryAu (approved):

PR #1214, opened 2025-03-19
`vite`: Emit initial start time telemetry for `sku start`

askoufis:
This takes a similar approach to #1211, though it's not quite the same. The issue with `sku start` in a Vite world is that a user could theoretically start the CLI, but then open the initial route in the browser 10 minutes later. Since Vite bundles on-request in dev mode, we can't simply time how long it takes for that initial request to finish, we also need to account for the gap between running `sku start` and actually requesting a route.

This PR implements a solution by tracking 2 separate times: the amount of time taken for the vite dev server to start listening, and how long it takes between the first request being sent and the browser triggering the `load` event. We inject a small client-side script that sets an event listener on the `load` event and sends a websocket message to the dev server. _Only_ when this event is received to we calculate the total time for CLI start and initial request and then emit telemetry.

Unfortunately I couldn't isolate the entire implementation to just a single vite plugin, as we have to mark the initial load time from within the request middleware so we can ensure it starts as soon as we get a request in. An alternative I considered was injecting a separate middleware in the `skuViteStartTelemetry` plugin just before the request middleware, but that relies on the ordering of middlewares which doesn't seem like a good idea IMO.

For reasons I don't understand ([potentially involving `renderToPipeableStream`](https://github.com/vitejs/vite-plugin-react/issues/222)), injecting the script via `transformIndexHtml` during `start-ssr` doesn't seem to work, so for now this telemetry is only emitted for Vite `sku start`, not `sku start-ssr`. We not be able to use `transformIndexHtml` to inject the script during SSR, and might need to just inject it manually during the pre-render.

In the `vite-test-app` fixture, CLI start time is ~4.5s, and initial page load time is ~1-2s, for a total "initial start" time of ~5.5-6.5s.

### DIFF
--- a/packages/sku/src/services/telemetry/metricsMeasurers.ts
+++ b/packages/sku/src/services/telemetry/metricsMeasurers.ts
@@ -1,0 +1,42 @@
+import debug from 'debug';
+
+const log = debug('sku:metrics');
+
+const skuStartMarkName = 'skuStartComplete';
+const skuStart = {
+  mark: () => {
+    performance.mark(skuStartMarkName);
+  },
+  measure: () => {
+    const result = performance.measure('skuStart', {}, skuStartMarkName);
+
+    log(`Sku dev server start took ${Math.round(result.duration)}ms`);
+
+    return result;
+  },
+};
+
+const initialPageLoadMarkName = 'initialPageLoadStart';
+const initialPageLoad = {
+  mark: () => {
+    performance.mark(initialPageLoadMarkName);
+  },
+  isInitialPageLoad: true,
+  measure() {
+    this.isInitialPageLoad = false;
+
+    const result = performance.measure(
+      'initialPageLoad',
+      initialPageLoadMarkName,
+    );
+
+    log(`Initial page load took ${Math.round(result.duration)}ms`);
+
+    return result;
+  },
+};
+
+export const metricsMeasurers = {
+  skuStart,
+  initialPageLoad,
+};
--- a/packages/sku/src/services/vite/index.ts
+++ b/packages/sku/src/services/vite/index.ts
@@ -41,8 +41,8 @@ export const viteService = {
     const server = await createViteServerSsr({
       skuContext,
     });
+    server.listen(skuContext.port.server);
 
-    await server.listen(skuContext.port.server);
     if (skuContext.sites.length > 1) {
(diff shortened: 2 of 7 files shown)


packages/sku/src/services/telemetry/metricsMeasurers.ts
```diff
@@ -0,0 +1,42 @@
+import debug from 'debug';
+
+const log = debug('sku:metrics');
+
+const skuStartMarkName = 'skuStartComplete';
+const skuStart = {
```
askoufis, line 6: This module is an attempt to unify our metric measurements into a single place. It might even be worth emitting telemetry from here as well, so that the raw `provider` API is abstracted away into something more type-safe and explicit.

The distinction I've made between `metrics` and `telemetry` is that the metrics are the actual measurements and values, whereas the telemetry is the communication of those metrics to somewhere.

packages/sku/src/services/vite/index.ts
```diff
@@ -41,8 +41,8 @@ export const viteService = {
     const server = await createViteServerSsr({
       skuContext,
     });
+    server.listen(skuContext.port.server);
 
-    await server.listen(skuContext.port.server);
```
askoufis, line 45: This `server.listen` doesn't need to be `await`ed.

packages/sku/src/services/vite/plugins/skuViteStartTelemetry.ts
```diff
@@ -0,0 +1,55 @@
...
+      if (metricsMeasurers.initialPageLoad.isInitialPageLoad) {
+        const { duration: skuStartDuration } =
+          metricsMeasurers.skuStart.measure();
+
+        const { duration: initialLoadDuration } =
+          metricsMeasurers.initialPageLoad.measure();
+
+        const initialStartTime = skuStartDuration + initialLoadDuration;
+
+        provider.timing('start.webpack.initial', initialStartTime, {
```
DanDroryAu, line 40: **question:** how does the timing tool know that this is `vite` over `webpack`?

DanDroryAu (approved):
Nice work. Looking forwards to expending this set of metrics.

askoufis:
Tested in an internal project, confirmed that telemetry is flowing through.

PR #1219, opened 2025-03-24
Update root resolution ordering, support `.jsx` and `.cjs`

askoufis:
Vite is a bit stricter about what can go inside a `.js` file, disallowing any JSX syntax. While we want to align with this restriction, currently consumers aren't able to as we don't support importing `.jsx` files without an extension. This PR fixes that, while also enabling root resolution of `.jsx` files too. `.cjs` was thrown in there too as it's supported by `sku` in other ways already (e.g. linting, babel transformation, etc.).

Additionally, as the file extensions are checked in the order defined by the array, it makes sense to put the most common file extensions first. For most internal apps, this would be TypeScript files, hence the re-ordering. The `m` and `c` prefixed extensions are less common, so they were de-prioritized relative to their less-specific versions.

This will also help unblock ongoing vite testing work.

### DIFF
--- a/packages/sku/src/services/webpack/config/plugins/sku-webpack-plugin/index.ts
+++ b/packages/sku/src/services/webpack/config/plugins/sku-webpack-plugin/index.ts
@@ -16,6 +16,7 @@ import validateOptions, {
   type SkuWebpackPluginOptions,
 } from './validateOptions.js';
 import targets from '@/config/targets.json' with { type: 'json' };
+import { rootResolutionFileExtensions } from '@/config/fileResolutionExtensions.js';
 
 class SkuWebpackPlugin implements WebpackPluginInstance {
   options: SkuWebpackPluginOptions;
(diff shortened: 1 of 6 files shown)


packages/sku/src/services/webpack/config/utils/index.ts
```diff
@@ -1,13 +1,8 @@
+import { extensions } from '@/config/fileResolutionExtensions.js';
+
 export * from './loaders.js';
 export { resolvePackage } from './resolvePackage.js';
 
-// TODO: Figure out a better way to share this config. This is currently copy-pasted from
```
DanDroryAu, line 4: We finally figured it out!

packages/sku/src/services/webpack/config/plugins/sku-webpack-plugin/index.ts
```diff
@@ -176,21 +177,14 @@ class SkuWebpackPlugin implements WebpackPluginInstance {
...
     } else {
-      if (!compiler.options.resolve.extensions.includes('.ts')) {
-        compiler.options.resolve.extensions.push('.ts');
-      }
+      const dedupedExtensions = new Set([
+        ...rootResolutionFileExtensions,
+        ...compiler.options.resolve.extensions,
+      ]);
 
-      if (!compiler.options.resolve.extensions.includes('.tsx')) {
```
DanDroryAu, line 191: I believe there was a request sometime back that was fiddling with this to get `.jsx` to work. This is a nice fix.

DanDroryAu (approved):

packages/sku/src/services/webpack/config/utils/index.ts
askoufis: Well, kind of. These extensions are in eslint, but we didn't want to tie ourselves to eslint in case the list changed without us knowing. IMO it's still an unsolved problem.

PR #1220, opened 2025-03-24
Test `braid-design-system` fixture with vite

askoufis:
The `braid-design-system` fixture is now also tested with the vite bundler (except the `sku test` test, for now).

This involves a few steps:
- Creating a separate render entrypoint and sku config for vite
- Changing the react root element id as it is currently hardcoded as `root` for vite
- Sanitizing snapshot HTML so the working directory doesn't show up in snapshots
- Making a small change to the `setupTestHosts` script to ensure the dev hostname can also point to `::1`, which seems to be the address vite hosts the server on in CI

Hopefully with this change, and the ones in #1216 and #1218, we can now test the majority, if not all, of our fixtures with vite, in addition to webpack.

> [!NOTE]
> Currently there is a bug in the vite implementation where critical CSS is not present in the static render.
> This results in large post-hydration diffs in snapshot tests.
> This is a known issue.


askoufis:
Aaaaaaaaand we're back to mismatched snapshots between local and CI 😭

tests/braid-design-system.test.js
```diff
@@ -24,55 +24,62 @@ function getLocalUrl(site) {
...
 
-    it('should generate the expected files', async () => {
-      const files = await dirContentsToObject(distDir);
-      expect(files).toMatchSnapshot();
+    describe('build', () => {
+      let process;
+
+      beforeAll(async () => {
+        await runSkuScriptInDir('build', appDir, args);
+        process = await runSkuScriptInDir('serve', appDir, args);
```
DanDroryAu, line 61: **nitpick:** `serve` doesn't need the extra args. I think it should complain about the `--experimental-bundler` flag only being used for `build | start`

DanDroryAu (approved):
small comment but can be a fast follow.

LGTM!

jahredhope (approved):

tests/braid-design-system.test.js
askoufis: Good pickup. [Removed the `args`.](https://github.com/seek-oss/sku/pull/1220/commits/d2df1e51c96f336b9062484b4d858b27825b3fdf)

PR #1221, opened 2025-03-24
`setup-hosts`: Add ipv6 hosts alongside ipv4

williamlark:
When running the `setup-hosts` command, only ipv4 hosts were configured. This works most of the time, but some bundlers (such as Vite) prefer serving on ipv6 ports. This change adds the ipv6 localhost equivalent to the hosts file alongside v4 to increase compatibility with bundlers that prefer it.

I've also added some tests to make sure we don't break this accidentally and add to the documentation-as-code.

### DIFF
--- a/.changeset/slimy-pots-press.md
+++ b/.changeset/slimy-pots-press.md
@@ -2,4 +2,4 @@
 'sku': minor
 ---
 
-setup-hosts: ipv6 hosts are now added alongside their v4 counterpart.
+`setup-hosts` ipv6 hosts are now added alongside their ipv4 counterpart
--- a/.changeset/slimy-pots-press.md
+++ b/.changeset/slimy-pots-press.md
@@ -1,0 +1,5 @@
+---
+'sku': minor
+---
+
+setup-hosts: ipv6 hosts are now added alongside their v4 counterpart.
--- a/packages/sku/src/utils/contextUtils/hosts.test.ts
+++ b/packages/sku/src/utils/contextUtils/hosts.test.ts
@@ -1,0 +1,151 @@
+import { createSkuContext } from '@/context/createSkuContext.js';
+import { checkHosts, setupHosts } from './hosts.js';
+import { jest } from '@jest/globals';
+
+describe('setupHosts', () => {
+  beforeEach(() => {
+    // Silence the logging for the tests
+    jest.spyOn(global.console, 'log').mockImplementation(() => {});
+    jest.spyOn(global.console, 'error').mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('should set site-specific hosts', async () => {
+    const context = createSkuContext({});
+    const mockSetHosts = jest.fn(async () => {});
+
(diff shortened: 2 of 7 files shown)


packages/sku/src/utils/contextUtils/hosts.test.ts
```diff
@@ -0,0 +1,151 @@
+import { createSkuContext } from '@/context/createSkuContext.js';
+import { checkHosts, setupHosts } from './hosts.js';
+import { jest } from '@jest/globals';
+
+describe('setupHosts', () => {
+  beforeEach(() => {
+    // Silence the logging for the tests
+    jest.spyOn(global.console, 'log').mockImplementation(() => {});
+    jest.spyOn(global.console, 'error').mockImplementation(() => {});
```
askoufis, line 9: I don't _mind_ suppressing logs, but honestly it's just more code to write, and if there aren't _that_ many logs I don't really see the value TBH.

.changeset/slimy-pots-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': minor
+---
+
+setup-hosts: ipv6 hosts are now added alongside their v4 counterpart.
```
askoufis, line 5: Some small nitpicks:

For command-related changes, or really anything that involves code, e.g. a function, library name, anything you'd type in a terminal, I tend to them in inline code.

Additionally, we tend to treat the first line in a changeset like a dot point, in that we don't end it with a full stop unless it contains multiple sentences (though I try not to do that if possible). It ultimately serves as a summary of the change, and _if_ there's a need for further explanation, it should be present in further paragraphs below.

```suggestion
`setup-hosts` ipv6 hosts are now added alongside their ipv4 counterpart
```

packages/sku/src/utils/contextUtils/hosts.test.ts
williamlark: I just don't like that they clog up the output, and particually when I'm expecting an error in the tests it makes it look like something went wrong. I'd rather keep them silenced in this test, but I can see why you might keep them on for long running tests. 

CleanShot 2025-03-25 at 11 42 40

.changeset/slimy-pots-press.md
williamlark: Thank you. Good to know!

packages/sku/src/utils/contextUtils/hosts.test.ts
askoufis: I guess for me the thing I'm looking for is the test to pass or fail. The presence, or lack, of console output isn't really of much concern (in this case), so I tend not to suppress it as that's 4 lines of code I don't need to write.

williamlark: Personally I think a couple lines of code to save far more noise in the the terminal output is fine here. I know in CI it's not an issue but its a nicer experience locally when writing with watch mode on and not seeing all the noise in the console 😄 I'd still prefer to keep them suppressed here if you're okay with that.

askoufis (approved):

PR #1222, opened 2025-03-25
Match vite render API to the webpack api

DanDroryAu:
## Outline

the initial implementation of experimental `vite` support came with a changed API. In order to get feature parity and make the transition from `webpack` to `vite` smooth in the future the API of experimental `vite` should match that of the current `webpack` API. This PR aims to make that improvement.

The larger change here is in the `vite` middleware that runs the dev server and transforms the index.html file when requested. 

## Changes

Add additional log information for `rendering` `middleware` and `loadable`.

Change vite render api to match that of webpack.
Change `suspense` fixture, allowing both vite and webpack bundling.
Change `suspense` test to run for both vite and webpack.
Change type for `Render` and added a new `LibraryRender` type.

Remove sections of the docs that are outdated regarding vite migration.
Remove the `vite-test-app` fixture. This fixture was only there for internal testing. With these new changes we can start applying `vite` on the existing fixtures.


packages/sku/src/types/types.ts
```diff
@@ -72,16 +36,18 @@ export interface Server {
...
   route: string;
-  environment: string;
-  site: string;
-  language: string;
-  libraryName: string;
-  libraryFile: string;
+  environment: string | undefined;
+  site: string | undefined;
+  language: string | null;
+}
```
askoufis, line 45: These changes are going to flow through to every webpack app that uses the `Render` type. I'd consider these breaking changes as I reckon many apps have taken these to always be defined, but now may not be available.

I hate to fracture the API more, but maybe a separate `ViteRender` type is necessary for now?

askoufis: Or can we lie and say they're always defined, and then do a breaking change later?

DanDroryAu: youre right this would be a breaking change since for `library` we would now need a `LibraryRender` I was experimenting to see if I could do a type that would infer but there is no way to know what it is when we import the render object via the alias.

packages/sku/src/types/types.ts
```diff
@@ -97,7 +66,7 @@ interface RenderDocumentProps<App> extends SharedRenderProps {
 }
 
 export interface Render<App = string> {
-  renderApp(p: RenderAppProps): Promise<App> | App;
+  renderApp?(p: RenderAppProps): Promise<App> | App;
```
askoufis, line 69: I'm confused. Why has this become optional all of a sudden?

packages/sku/src/services/vite/helpers/html/createPreRenderedHtml.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const clientContext =
+    (await render.provideClientContext?.({
+      ...renderContext,
+      app,
+      site,
+    })) || {};
+
+  function getHeadTags() {
+    return [
+      // ...loadableCollector.getAllLinks(),
```
askoufis, line 91: Is this meant to be commented out?

packages/sku/src/types/types.ts
DanDroryAu: A library doesn't need a `renderApp` which is why it should be optional. Check the library test fixture.

packages/sku/src/services/vite/helpers/html/createPreRenderedHtml.tsx
DanDroryAu: Yeah was commented out but needs to be removed. `.getAllLinks()` returns linkheaders.

packages/sku/src/types/types.ts
askoufis: Are we still keeping these optional?

DanDroryAu: you mean `libraryName` and `libraryFile`? I believe so.

askoufis: Sorry, meant `routeName`, `environment`, `site` and `language`. That's what I was talking about originally.

DanDroryAu: Aha! But their types were `string | undefined`. That confused me. Let me investigate why this change was made again. It's to make it match types but I dont know why. Somewhere in the route conversion functions it got a bit wonky.

askoufis: Yeah I'm assuming it was to better match the actual output. It makes sense as they can all likely be optional, but again, I think the assumption that they're always defined is likely baked into many apps, so I'd consider making them optional as a breaking change.

DanDroryAu: Yeah, it's to match them with the `createHtmlRenderPlugin` required types.

I'll add the changeset to notify.

askoufis: I'm suggesting that we shouldn't make this change. I understand that the types are more correct, but IMO it's quite a significant break. Every consumer currently assumes that these values are defined. If they suddenly might not be, every consumer will have to add logic to their render entry to handle this.

packages/sku/src/services/vite/plugins/skuViteMiddlewarePlugin.ts
```diff
@@ -2,52 +2,83 @@ import type { SkuContext } from '@/context/createSkuContext.js';
...
+        return;
+      }
+
+      log('Matching route found:', matchingRoute);
+
+      const language =
+        getLanguageFromRoute(req.url, matchingRoute, skuContext) || '';
+
+      const renderEntry = require.resolve('../entries/vite-render.js');
+      const clientEntry = require.resolve('../entries/vite-client.js');
```
askoufis, line 62: These can be hoisted to a higher scope. Currently they're being resolved (to the same path) on every request.

askoufis (commented):
Needs a changeset

askoufis (approved):

PR #1233, opened 2025-04-15
add batch pre-render for vite

DanDroryAu:
Change pre-render functionality for `vite start`. It now runs route pre-rendering in batches using `worker_threads`.

You can test this functionality via the new `many-routes` fixture.


packages/sku/src/services/vite/index.ts
```diff
@@ -20,7 +20,7 @@ export const viteService = {
     await build(createViteConfig({ skuContext }));
     await build(createViteConfig({ skuContext, configType: 'ssg' }));
     if (skuContext.routes) {
-      await prerenderRoutes(skuContext);
+      await startPrerenderWorkers(skuContext);
```
williamlark, line 23: nit: Just a suggestion, but rather than naming this based on how it runs (with workers) we could name based on what it does. `prerenderConcurrently` (or `concurrentlyPrerender`). Thoughts?

packages/sku/src/services/vite/helpers/prerenderRoutesWorker.ts
williamlark, line 1: I notice the name of this is `prerenderRoutesWorker.ts` suggesting it's a worker, when in fact it's running the workers. Could be clearer to call this `startPrerenderWorkers.ts` (or `prerenderConcurrently.ts` as per my other suggestion) to distinguish it as not the actual worker.

packages/sku/src/services/vite/helpers/prerenderBatch.ts
williamlark, line 1: suggestion: love the helper folder, but it does blur the meaning of certain functions.  Could this live in a `/workers` folder? Or at the very least have `worker` in the file name -- though I do prefer a worker folder IMO so you don't even need to add it to the name 🧹 Thoughts?

williamlark (commented):
Very nice 🎉  Do you have any numbers on roughly how much it speeds up pre-rendering?

packages/sku/src/services/vite/helpers/prerenderBatch.ts
DanDroryAu: Yeah I'll put them in a pre-render folder. Makes more sense.

fixtures/many-routes/src/render.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  renderDocument: ({ app, bodyTags, headTags }) => {
+    return html/* html */ `
+      <!DOCTYPE html>
+      <html>
+        <head>
+          <meta charset="UTF-8" />
+          <title>hello-world</title>
+          <meta name="viewport" content="width=device-width, initial-scale=1" />
+          <script src="https://code.jquery.com/jquery-3.5.0.slim.min.js"></script>
```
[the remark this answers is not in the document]
DanDroryAu: This is just a copy from a previous fixture. It's mostly there to show that imported <scripts> in the document get added.

DanDroryAu:
> Very nice 🎉 Do you have any numbers on roughly how much it speeds up pre-rendering?

I don't know how much it speeds up. It depends on the number of routes. I would imagine that it speeds it up to the amount of cores the CPU has. So with 12 cores each having 2 threads I would imagine a 23x increase in speed (one core for the main thread).

I don't know how to run a test for this. I can write a script that generates loads of routes to see the effect but I don't know if that is useful.

packages/sku/src/services/vite/helpers/prerender/prerenderConcurrently.ts
```diff
@@ -0,0 +1,77 @@
...
+      if (code === 0) {
+        resolve();
+      } else {
+        reject(new Error(`Worker stopped with exit code ${code}`));
+      }
+    });
+  });
+};
+
+export const startPrerenderWorkers = async (skuContext: SkuContext) => {
```
williamlark, line 41: nit 🤏 

```suggestion
export const prerenderConcurrently = async (skuContext: SkuContext) => {
```

williamlark (approved):
🚀 Great work

PR #1237, opened 2025-05-01
Sku codemod package

DanDroryAu:
## Changes

Add the `sku-codemod` package. This package holds relevant codemods that our users can use to automatically migrate their projects over.

Add the `transform-vite-loadable` codemod. This codemod attempts to migrate over all imports of `'sku/@loadable/component'` to the new `'sku/vite/loadable'` imports required for `vite`.

## Reasoning

When we make breaking changes to `sku` we can support our users migration efforts by providing codemods for large sweeping changes. 

## Testing steps

To test if the mod works you can navigate to the `fixtures/multiple-routes` folder and run the following command `npx sku-codemod@sku-codemods` and follow the prompts.

The expected outcome is two files changed.


package.json
```diff
@@ -60,6 +60,7 @@
     "eslint-import-resolver-typescript": "^3.7.0",
     "eslint-plugin-jsdoc": "^50.2.2",
     "eslint-plugin-unicorn": "^56.0.1",
+    "fs-fixture": "^2.6.0",
```
askoufis, line 63: Can this dep be moved to the `sku-codemod` package.json?

tests/sku-codemods/sku-codemods.test.ts
```diff
@@ -0,0 +1,86 @@
+import { describe, it } from 'vitest';
+import fs from 'node:fs/promises';
+import { runSkuCodemod } from '@sku-private/test-utils';
+import dedent from 'dedent';
+import { createFixture } from 'fs-fixture';
+
+describe('sku codemods', () => {
+  describe('"transform-vite-loadable" codemod', async () => {
+    const filesToTest = [
```
askoufis, line 9: Any reason this can't be refactored to use [`describe.for`](https://vitest.dev/api/#describe-for) for better integration with the test runner?

packages/sku-codemod/src/program/index.ts
```diff
@@ -0,0 +1,33 @@
...
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { Command } from 'commander';
+import { runTransform } from '../transform/runner.js';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+const { name, description, version } = JSON.parse(
+  fs.readFileSync(resolve(__dirname, '../../package.json'), 'utf-8'),
+);
```
askoufis, line 11: Adding a package.json export makes importing this info much nicer IMO, and is just generally good practice for packages. E.g.

```jsonc
// package.json
  "exports": {
  }
```

```suggestion
import packageJson from 'sku-codemode/package.json' with { type: 'json' };

const { name, description, version } = packageJson;
```

packages/sku-codemod/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "bin.js",
+    "dist",
+    "tsconfig.json"
+  ],
+  "homepage": "https://github.com/seek-oss/sku/packages/sku-codemod#readme",
+  "devDependencies": {
+    "@types/diff": "^7.0.1",
+    "@types/jscodeshift": "^17.3.0",
+    "@types/prompts": "^2.4.9",
+    "@types/terminal-kit": "^2.5.7",
```
askoufis, line 40: This dev dep looks unused (or at least it has no corresponding regular dep).

packages/sku-codemod/src/bin/sku-codemods.ts
askoufis, line 1: I don't feel like it's worth having an extra file just to import and run `program.parse`. Could `bin.js` look like this?

```sh
#!/usr/bin/env -S node --disable-warning=ExperimentalWarning
import { program } from './dist/program/index.js';

program.parse();
```

packages/sku-codemod/src/utils/test-utils.ts
```diff
@@ -0,0 +1,101 @@
+import fs from 'node:fs';
```
askoufis, line 1: Worth using `node:fs/promises` where possible.

packages/sku-codemod/src/transform/runner.ts
```diff
@@ -0,0 +1,178 @@
...
+  }>;
+};
+
+export const runTransform = async (
+  transform: string,
+  path: string,
+  options: Options,
+): Promise<void> => {
+  let transformer = transform;
+  console.log('path', path);
```
askoufis, line 33: Can this log be removed?

packages/sku-codemod/src/transform/runner.ts
```diff
@@ -0,0 +1,178 @@
...
+  options: Options;
+  jobs: Array<{
+    filePath: string;
+  }>;
+};
+
+export const runTransform = async (
+  transform: string,
+  path: string,
+  options: Options,
```
askoufis, line 30: IMO we should emit a log if the user is running a dry run, just so that there's feedback that no files will be modified.

package.json
DanDroryAu: The test is in the mono-repo root. Does it still work if its in the subfolder package.json

askoufis: I think it should be fine since the actual test file is inside `sku-codemod`, so when it resolves the dep it will first check that package, and then the root. The test runner itself being at the root shouldn't matter.

packages/sku-codemod/src/program/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  .helpOption('-h, --help', 'Display this help message.')
+  .option('-d, --dry', 'Dry run (no changes are made to files)')
+  .option(
+    '-p, --print',
+    'Print transformed files to stdout, useful for development',
+  )
+  .option('--verbose', 'Show more information about the transform process')
+  .action(runTransform)
+  .allowUnknownOption()
+  .enablePositionalOptions();
```
williamlark, line 33: What's the purpose of these two options enabled?

packages/sku-codemod/src/transform/runner.ts
```diff
@@ -0,0 +1,178 @@
...
+): Promise<void> => {
+  let transformer = transform;
+  console.log('path', path);
+  let directory = path;
```
williamlark, line 34: This is perhaps a style choice, but I can assure you that `let` is needed far less than you think, and can usually be written with `const` instead. I find you can catch encapsulations better when trying to write with less reassignable variables, and keeps code readable in a linear format. Debugging with `let` usually required more mental effort compared to using variables that can't be reassigned. 

Here's a suggestion for this code:

```suggestion
  const transformer = transform || (await promptTransformationType())
  const directory = path || (await promptDestination());
  console.log('path', directory);
```

packages/sku-codemod/src/transform/runner.ts
```diff
@@ -0,0 +1,178 @@
...
+          const files = globbySync([file, '!**/node_modules', '!**/dist'], {
+            absolute: true,
+          });
+          r(files);
+        }),
+    ),
+  );
+
+  return list.flat();
+};
```
williamlark, line 146: Can't globby do all this for you?

```suggestion
const getAllFiles = async (paths: string[]) => globby([...paths, '!**/node_modules', '!**/dist'], {
            absolute: true,
          });
```

packages/sku-codemod/src/utils/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+  title: string;
+  value: string;
+  version: string;
+};
+
+export const CODEMODS: Codemod[] = [
+  {
+    title: 'Transform all webpack loadable imports to Vite loadable imports',
+    value: 'transform-vite-loadable',
+    version: '0.0.1',
```
williamlark, line 11: What's the use of the version here? I image this wouldn't get updated much, or easily forgotten. Changesets versioning on the main package would be enough imo, but I may be missing some function here.

williamlark (commented):
Nice work! I tested it out and it works great 🎉  Left a few comments.


#### Testing notes 

* When running a dry run it still says `Changed files: x`. It could be clearer that files aren't being changed.

packages/sku-codemod/src/transform/runner.ts
DanDroryAu: Do you mean just logging out that this is a dry run? 

the `-p` command prints the diff log so combining them does a print/dry run.

askoufis: Yep, just a log saying "this is a dry run, no files will be affected". This helps distinguish a run that didn't modify any files because it didn't find anything to modify, vs a dry run where regardless of changes found, no files will be modified.

fixtures/multiple-routes/package.json
```diff
@@ -12,7 +12,8 @@
     "@sku-private/test-utils": "workspace:*",
     "@testing-library/react": "^14.0.0",
     "dedent": "^1.5.1",
-    "sku": "workspace:*"
+    "sku": "workspace:*",
+    "sku-codemod": "workspace:*"
```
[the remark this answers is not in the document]
DanDroryAu: It was for me to test them in these packages.

I can remove this.

package.json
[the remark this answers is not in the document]
DanDroryAu: I'll pop this in the tests package.json

packages/sku-codemod/src/program/index.ts
DanDroryAu: Positional options for for subcommands and such. But it was there before to allow optional comments to go to `tscodeshift` which has since been replaced with `ast-grep` I can change this.

packages/sku-codemod/src/utils/constants.ts
DanDroryAu: Yeah I copied this from the way `next.js` handles this. I thought it would be nice to update this in case we missed something in the future with these. But we can forgo this version value if we want.

packages/sku-codemod/src/codemods/transform-vite-loadable/__tests__/transform-vite-loadable.test.ts
```diff
@@ -0,0 +1,20 @@
+import { defineTest } from '../../../utils/test-utils.js';
+
+defineTest(__dirname, 'transform-vite-loadable', 'loadableNameFixture', {
```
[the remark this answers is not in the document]
DanDroryAu: It's was a copy of `tscodeshift`'s api. That api is based on `jest` and we're going to `vitest`. I can change the test util a bit to make it fit better. The API has gotten a bit muddy. 

In the future I do imagine that we do a lot of similar tests so having a good util that does most of the boilerplate stuff for us would be good to have.

package.json
DanDroryAu: The test file is inside `tests/sku-codemods` so I'll put this inside the `tests/package.json`

askoufis: Oh my bad, yeah that's where it should live.

packages/sku-codemod/src/transform/runner.ts
DanDroryAu: I'll look at refactoring this.

williamlark (approved):

packages/sku-codemod/README.md
```diff
@@ -0,0 +1,31 @@
+[![npm](https://img.shields.io/npm/v/sku-codemod.svg?style=flat-square)](https://www.npmjs.com/package/sku-codemod)
+
+# Codemods for `sku`
+
+## Usage
+
+```sh
+pnpm dlx sku-codemod <codemod> [files] [options]
+```
```
askoufis, line 9: Should it also be mentioned that `pnpm dlx sku-codemod` run without `codemod` or `files` will show prompts for those options?

askoufis (approved):
Looking good! Just one small comment.

PR #1238, opened 2025-05-05
`tests`: Running sku with `execa` in test suits

williamlark:
Our tests run many instances of sku that are started within child process. We were using `cross-spawn` to do so, as well as a custom script to start a worker directly. This API was inconsistent and lacked good error handling, but worked fine nonetheless.

This change aims to make the error handling from child-process within out tests better and aligns the two ways we start sku in the test suit. This change uses [execa](https://github.com/sindresorhus/execa) to run sku, and changes the way we kill the process within out tests by using AbortControllers.


fixtures/braid-design-system/package.json
```diff
@@ -3,10 +3,14 @@
   "private": true,
   "type": "module",
   "dependencies": {
-    "braid-design-system": "^33.1.0",
+    "braid-design-system": "^33.6.0",
```
askoufis, line 6: I would avoid updating braid here, as it's causing snapshot updates, which are just noise in this PR.

test-utils/package.json
```diff
@@ -12,19 +12,19 @@
...
     "serve-handler": "^6.1.3",
-    "tree-kill": "^1.2.1",
     "wait-on": "^8.0.1",
     "webpack-stats-plugin": "^1.0.3"
   },
-  "skuSkipValidatePeerDeps": true
+  "skuSkipValidatePeerDeps": true,
+  "dependencies": {
+    "execa": "^9.5.2"
+  }
```
askoufis, line 29: Can this be moved to a dev dependency with the rest of the others? Also, would it be possible to swap to [`tinyexec`](https://www.npmjs.com/package/tinyexec)? I'd prefer it over `execa`, which has [quite a lot of deps](https://npmgraph.js.org/?q=execa), and we likely don't need any of its fancy features.

askoufis: Happy for this to be done in a separate PR so it doesn't block your vocab work.

test-utils/process.ts
```diff
@@ -60,31 +45,14 @@ type SkuScript =
...
   cwd: string,
-  args?: string[],
-  options?: SpawnOptions,
-): Promise<
-  ChildProcess | { stdout: string; stderr: string; child: ChildProcess }
-> {
-  const processOptions = {
+  args: string[] = [],
+  options?: Options,
+) {
```
DanDroryAu, line 53: **Suggestion:** would it be beneficial to turn the parameters into a single object rather than positional arguments?

I see you have to fill out an empty array for `args` a lot in the function calls in the tests.

vitest.config.ts
```diff
@@ -1,14 +1,17 @@
 import { defineConfig } from 'vitest/config';
 import tsconfigPaths from 'vite-tsconfig-paths';
 
+export const TEST_TIMEOUT = 50_000;
```
DanDroryAu, line 4: **praise:** I like it. Keeps it in sync.

tests/public-path.test.js
```diff
@@ -4,6 +4,7 @@ import path from 'node:path';
 import { runSkuScriptInDir, waitForUrls } from '@sku-private/test-utils';
 
 import { createRequire } from 'node:module';
+import { createCancelSignal } from '@sku-private/test-utils/process.ts';
```
DanDroryAu, line 7: **issue:** importing from `.ts` file directly.

Can you change this to import it from the barrel file like the other `test-utils`?

tests/multiple-routes.test.js
```diff
@@ -9,6 +9,7 @@ import {
 import { getAppSnapshot } from '@sku-private/vitest-utils';
 
 import { createRequire } from 'node:module';
+import { createCancelSignal } from '@sku-private/test-utils/process.ts';
```
DanDroryAu, line 12: See my other comment.

test-utils/process.ts
DanDroryAu: This also means I should re-factor some of my test utils for codemods ;)

DanDroryAu (commented):
Nice stuff. 

Reducing these flaky tests is a much needed productivity boost. Left some small comments. let me know if you need elaboration on them.

🥇

test-utils/package.json
williamlark: Can do! I've swapped it over to [nano-spawn](https://github.com/sindresorhus/nano-spawn) since it's even smaller than tinyexec (and maintained by the execa guy)

test-utils/process.ts
williamlark: Good call. I'll keep the script and directory as parameters since their always required, but move the args into the options object.

tests/library-file.test.js
```diff
@@ -30,15 +31,15 @@ describe('library-file', () => {
...
     beforeAll(async () => {
-      server = await runSkuScriptInDir('start', appDir);
+      runSkuScriptInDir('start', appDir, [], { cancelSignal: signal });
       await waitForUrls(devServerUrl);
     });
 
     afterAll(async () => {
-      await server.kill();
+      cancel();
     });
```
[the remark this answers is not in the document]
williamlark: I think this could be good. I might leave it in this PR just to keep this moving, but I can explore a little on what we could do to reduce this everywhere.

fixtures/braid-design-system/package.json
askoufis: Oh, completely missed that you converted the braid fixture to typescript, and that's why the snapshot has updated.

tests/__snapshots__/braid-design-system.test.ts.snap
askoufis, line 1: The `.js.snap` file should be deleted now that the test file is `.ts` right?

williamlark: Yes, good call 👍

DanDroryAu (approved):

PR #1239, opened 2025-05-06
Vite docs update

DanDroryAu:
Add migration docs for `vite`.

There is no link to the docs page in the side bar.


packages/sku-codemods/package.json
askoufis, line 1: Seems this was accidentally committed.

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,95 @@
+## Migration guide for `vite`
+
+This document outlines the steps to migrate your `sku` project to use `vite` as the bundler. The migration process is relatively straightforward and should not require any major changes to your codebase.
+
+### Migrate to `sku/vite/loadable`
+
+> [NOTE!]: If you're not using `sku/@loadable/component` you can skip this step.
+
```
askoufis, line 8: I'm not sure if these work on our docs site. Best to just use a regular blockquote.
```suggestion
> If you're not using `sku/@loadable/component` you can skip this step.

```

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,95 @@
+## Migration guide for `vite`
```
askoufis, line 1:
```suggestion
## Migrating to `vite`
```

Additionally, these headings all need to be 1 level higher. The name of the document will not be automatically rendered as an `h1` unfortunately.

docs/docs/vite-migration.md
askoufis, line 1: As a general rule we've tried to write markdown with 1 sentence per line (i.e. newlines after every full stop). This makes it easier to comment on diffs as well as read the raw markdown.

I want to enforce this with tooling eventually, but haven't gotten around to it.

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,95 @@
...
+{
++   "type": "module",
+    "name": "my-project",
+    ...
+}
+```
+
+### Update jsx file extensions.
+
+In order to use `vite` as the bundler you will need to update your file extensions to `.jsx` or `.tsx` on files that contain JSX. This is a requirement of `vite` and is not specific to `sku`.
```
askoufis, line 73: `.js` -> `.jsx` is the relevant file extension change here. TS files with JSX will already be `.tsx`, that restriction exists currently without vite.

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Update `sku.config.ts` file
```
askoufis, line 75: There's overlap here with the existing `Vite support` document. Is the intention to replace that document entirely, or separate this document as the explicit migration guide, and make the other document about features/differences between webpack and vite?

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,95 @@
...
+import type { SkuConfig } from 'sku';
+
+export default {
+  __UNSAFE_EXPERIMENTAL__bundler: 'vite', // sku will use vite as the bundler.
+  ...
+} satisfies SkuConfig;
+```
+
+The `dangerouslySetWebpackConfig` option is not supported in `vite` and will be ignored.
+We've currently got no means to change the vite config but are open to requests for this feature.
```
askoufis, line 91: This is a bit misleading. We _do_ have the means to change the config, we just haven't exposed any user API for it.

```suggestion
There is currently no equivalent `dangerouslySetViteConfig` API available.
Support for this API may not be implemented unless there is good reason to do so.
```

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,95 @@
...
+In order to use `vite` as the bundler you will need to update your project to `esm`. This can be done by changing the `type` field in your `package.json` file to `module`.
+
+```diff
+// package.json
+{
++   "type": "module",
+    "name": "my-project",
+    ...
+}
+```
```
askoufis, line 69: Feel like this is underselling this step a lot. I'd say for the majority of apps it will involve _much_ more than just setting `"type": "module"`. 

Teams will need to rename `.js` files to `.cjs` if they're still using `require`, or migrate those files to ESM. They might need to add `--experimental-vm-modules` to their `test` scripts. Dependencies may need to be updated to be compatible with ESM. It's honestly a huge unknown that I'm quite worried about.

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,95 @@
...
+- import loadable from 'sku/@loadable/component';
++ import { loadable } from 'sku/vite/loadable';
+```
+
+We've provided a codemod to help you out with this process. Note that it may have unexpected results, so please review the changes it makes.
+
+```bash
+pnpm dlx sku-codemod transform-vite-loadable .
+# optionally, you can use the --dry and --print flag to see what changes will be made without actually modifying any files.
+pnpm dlx sku-codemod transform-vite-loadable . -dp
```
askoufis, line 23: Worth using the non-shorthand syntax here.

```suggestion
pnpm dlx sku-codemod transform-vite-loadable . --dry --print
```

docs/docs/vite-migration.md
DanDroryAu: I want to make this the `migration` guide and the other one about features indeed. I haven't had a look at the `vite-support` docs so let me check the overlap.

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,108 @@
...
+  // the following options are unchanged.
+  provideClientContext: ...,
+  renderDocument: ...,
+};
+```
+
+### Update server entrypoint
+
+`sku` in `vite` mode does not support `ssr` rendering yet.
+Stay tuned for server rendering support in the future.
```
williamlark, line 62: I feel like this should be added to the top of a file, perhaps under something along the lines of a `App requirements`/`Support apps`/`Compatible apps` header. Somewhere we can list up front which apps and/or features we support so people don't start migrating an app without reading the full doc, and then find out it doesn't support it yet.

This would also make updating this doc easier, since all the "to do"s and missing features are listed at the top.

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+Once the project is updated to `esm` you will need to update your imports to use the new `import` syntax.
+Any files that use `require` will need to be changed from `.js` to `.cjs`.
+Dependencies may need to be updated to `esm` compatible versions.
+
+### Jest support
+
+`jest` does not [officially](https://jestjs.io/docs/ecmascript-modules) support `esm` yet.
+In order to run `jest` or `sku test` you will need to use the `--experimental-vm-modules` flag to enable `esm` support.
```
williamlark, line 85: A code snippet here would be nice for the visual learners.

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Update jsx file extensions.
+
+In order to use `vite` as the bundler you will need to update your `.js` file extensions to `.jsx` files that contain `JSX`.
+This is a requirement of `vite` and is not specific to `sku`.
```
williamlark, line 90: I reckon we can reduce saying "In order to use `vite` as the bundler" since that the purpose of the guide anyway. A small diff in the docs could be nice too (makes it easy to scan the changes needed)

```suggestion
## Update jsx file extensions.

Javascript files containing JSX will need to be renamed to use the`.jsx` file extension.
This is a requirement of `vite` and is not specific to `sku`.

```diff
- App.js
+ App.jsx
```
```

File: docs/docs/vite-migration.md [hunk 11]
```diff
@@ -0,0 +1,108 @@
...
+
+## Update `sku.config.ts` file
+
+In order to use `vite` as the bundler you will need to update your `sku.config.ts` file to specify the `vite` bundler.
+This can be done by adding the `__UNSAFE_EXPERIMENTAL__bundler` field to your config file and setting its value to `vite`.
```
williamlark, line 95:
```suggestion
## Update `sku.config.ts` file

The vite bundler can be enabled via your projects `sku.config.ts` file. 
Add the `__UNSAFE_EXPERIMENTAL__bundler` field to your config file and setting its value to `vite`.
```

docs/docs/vite-migration.md
```diff
@@ -0,0 +1,108 @@
...
+import type { SkuConfig } from 'sku';
+
+export default {
+  __UNSAFE_EXPERIMENTAL__bundler: 'vite', // sku will use vite as the bundler.
+  ...
+} satisfies SkuConfig;
+```
+
+There is currently no equivalent `dangerouslySetViteConfig` API available.
+Support for this API may not be implemented unless there is good reason to do so.
```
williamlark, line 108: Similar to my other comments, this could be moved to the top of the guide so people know they can't (or may have issues to) upgrade their project upfront.

williamlark (commented):
Nice work! A few small things.

Also, I noticed we're missing mention of the `--experimental-bundler` flag that needs to be passed to the command, unless we're planning on getting rid of that.

PR #1241, opened 2025-05-09
allow ts-config-paths to be respected in vite

DanDroryAu:
Add the `tsConfigPaths` plugin to vite.

This change makes sure that our `tsconfig.json` is respected when it comes to path resolution.


packages/sku/src/services/vite/helpers/createConfig.ts
```diff
@@ -9,6 +9,7 @@ import type { SkuContext } from '@/context/createSkuContext.js';
 import skuVitePreloadPlugin from '../plugins/skuVitePreloadPlugin/skuVitePreloadPlugin.js';
 import { fixViteVanillaExtractDepScanPlugin } from '@/services/vite/plugins/esbuild/fixViteVanillaExtractDepScanPlugin.js';
 import { outDir, renderEntryChunkName } from './bundleConfig.js';
+import tsconfigPaths from 'vite-tsconfig-paths';
```
askoufis, line 12: `vite-tsconfig-paths` should be added as a dep within `sku`. It's currently only present as a monorepo (top-level) dep.

askoufis (approved):

DanDroryAu:
This pr requires a vocab plugin update to ensure it runs before the ts paths plugin.

packages/sku/package.json
```diff
@@ -112,7 +112,7 @@
     "@vocab/core": "^1.6.2",
     "@vocab/phrase": "^2.0.1",
     "@vocab/pseudo-localize": "^1.0.1",
-    "@vocab/vite": "^0.2.0",
+    "@vocab/vite": "0.2.1",
```
askoufis, line 115: @DanDroryAu This should stay pinned.

DanDroryAu: I'll fast follow this

PR #1243, opened 2025-05-12
Emit sku config path in generated ESLint config, don't redundantly load ESLint config

askoufis:
Currently sku both loads its own ESLint config programmatically _and_ lets ESLint load the config from `eslint.config.mjs`. Given that these are both the same configs, this is redundant. Removing the programmatic loading from `runESlint` simplifies a few functions as they no longer need to be passed a `skuContext`.

To support this change, we now need to emit the resolved sku config path to the user's eslint config during `sku configure`. I think we were getting a way with _not_ doing this in the past because we were loading the config programmatically. Now that we're not doing this we have to emit the config path.


.changeset/witty-dodos-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+`lint|format`: Don't redundantly load ESLint config
```
jahredhope, line 5: I like "Avoid", just a suggestion.

```suggestion
`lint|format`: Avoid redundantly loading ESLint config
```

jahredhope (approved):
Is good

jahredhope (approved):

PR #1244, opened 2025-05-12
Add `resolveComponent` to the `loadable` options api.

DanDroryAu:
# Outline:

The `@loadable` package has the [resolveComponent](https://loadable-components.com/docs/api-loadable-component/#optionsresolvecomponent) option that allows users to specify which component the `loadable` function should use.

This PR is adding that option to the `vite/loadable` function.

# Changes:

- Add the `resolveComponent` option to the `loadable` options.
- Updated `multiple-routes` fixture to use the `resolveComponent` option.

### DIFF
--- a/.changeset/stale-moles-nail.md
+++ b/.changeset/stale-moles-nail.md
@@ -1,0 +1,31 @@
+---
+'sku': minor
+---
+
+`vite/loadable`: Add the `resolveComponent` option to the `loadable` function. The `loadable` function returns the `default` export by default. By using the `resolveComponent` function you can specify the correct component for `loadable` to use.
+
+**example usage**
+
+```typescript
+// src/MyComponent.tsx
+import React from 'react';
+
+export const MyComponent = () => {
+  return <div>Hello, world!</div>;
+};
+
+// src/App.tsx
+import { loadable } from 'sku/vite/loadable';
+
+const MyComponent = loadable(() => import('./MyComponent'), {
+  resolveComponent: (module) => module.MyComponent,
+});
+
+export const App = () => {
+  return (
+    <div>
+      <MyComponent />
+    </div>
+  );
+};
+```
--- a/.changeset/stale-moles-nail.md
+++ b/.changeset/stale-moles-nail.md
@@ -8,8 +8,6 @@
 
 ```typescript
 // src/MyComponent.tsx
-import React from 'react';
-
 export const MyComponent = () => {
   return <div>Hello, world!</div>;
 };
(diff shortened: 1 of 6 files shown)


packages/sku/src/services/vite/loadable/loadable.tsx
```diff
@@ -17,13 +17,23 @@ export type PreloadableComponent<T extends ComponentType<any>> = T & {
...
-  factory: () => Promise<{ default: T }>,
+  factory: () => Promise<{ default: T } & Record<string, T>>,
   options?: {
     fallback?: NonNullable<ReactNode> | null;
+    resolveComponent?: (module: { default: T } & Record<string, T>) => T;
   },
   moduleId: ModuleId = '', // Gets set via the plugin
 ): PreloadableComponent<T> {
-  const ReactLazyComponent = lazy(factory);
+  const getDefaultExport = (module: { default: T } & Record<string, T>) =>
```
askoufis, line 27: Feel like this could have a better name that's not _only_ related to default exports:
```suggestion
  const getResolvedOrDefaultExport = (module: { default: T } & Record<string, T>) =>
```

fixtures/multiple-routes/src/handlers/AsyncComponent.jsx
```diff
@@ -1 +1 @@
-export default () => <span>Some special async content</span>;
+export const AsyncComponent = () => <span>Some special async content</span>;
```
askoufis, line 1: Can we test both a default and named export? Just so we can cover all our bases.

.changeset/stale-moles-nail.md
```diff
@@ -0,0 +1,31 @@
...
+'sku': minor
+---
+
+`vite/loadable`: Add the `resolveComponent` option to the `loadable` function. The `loadable` function returns the `default` export by default. By using the `resolveComponent` function you can specify the correct component for `loadable` to use.
+
+**example usage**
+
+```typescript
+// src/MyComponent.tsx
+import React from 'react';
```
askoufis, line 11: Probably not worth including the `react` import these days. Sku consumers don't need it anymore at least.

```suggestion
```

fixtures/multiple-routes/src/handlers/AsyncComponent.jsx
DanDroryAu: The default exports should be covered by the `App` itself. It has two default imports via loadable. Or do you want two specific tests for this?

askoufis: Oh if `App` uses loadable with default exports then yeah it's fine. Thought this was the only loadable component.

askoufis (approved):

PR #1246, opened 2025-05-12
Add ESLint rule to enforce dynamic action module imports, consolidate CLI command exports

askoufis:
- Added an ESLint rule that bans regular imports of `*.action.js` modules, which should hopefully ensure they stay dynamically imported
- Cleaned up some exports by exporting declarations directly
- Updated the `plop` templates to reflect these export changes for any future commands that may be added to the CLI

### DIFF
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -80,6 +80,19 @@ module.exports = [
       'no-process-exit': 'off',
       'no-sync': 'off',
 
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: [
+            {
+              group: ['**/*.action.js'],
+              message:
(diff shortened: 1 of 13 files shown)


eslint.config.cjs
```diff
@@ -80,6 +80,19 @@ module.exports = [
...
+        {
+          patterns: [
+            {
+              group: ['**/*.action.js'],
+              message:
+                'Please dynamically import action modules to keep the CLI fast',
+            },
+          ],
+        },
+      ],
```
DanDroryAu, line 94: **praise:** nice lint rule and custom warning

DanDroryAu (approved):

PR #1247, opened 2025-05-14
`vite`: Ensure `publicPath` is prepended to URLs inside CSS assets

askoufis:
The primary goal of this PR was to get the `publicPath` test suite working with vite, as well as to ensure that the `publicPath` fixture generates the expected output.

Before these changes, we were setting [`base`](https://vite.dev/guide/build.html#public-base-path) to `/`. `base` is effectively Vite's version of `publicPath`. This was fine for any assets handled by the loadable collector as it would prepend paths with `publicPath`, however asset import URLs and CSS [`url`](https://developer.mozilla.org/en-US/docs/Web/CSS/url_function) paths inside VE modules were not being handled at all.

This turned out to be caused by the VE Vite plugin forwarding the `base` config through to the VE compiler. More details can be found in the fix PR https://github.com/vanilla-extract-css/vanilla-extract/pull/1583.

After this change, I also noticed that for webpack we set `publicPath` to `/` _only_ during `start` scripts. From what I can tell this is primarily done to provide better DX during local dev as it means your site's paths won't be prepended with your configured `publicPath`.

Additionally, this seems to have also fixed a bug where `modulepreload` link tags were being generated with incorrect (not accounting for `publicPath`) paths. See the `multiple-routes` snapshots for the change.

### DIFF
--- a/fixtures/public-path/sku.config.js
+++ b/fixtures/public-path/sku.config.js
@@ -1,8 +1,9 @@
 import { makeStableHashes } from '@sku-private/test-utils';
 
 export default {
-  publicPath: '/static',
-  target: 'dist/static',
+  clientEntry: 'src/client.jsx',
+  renderEntry: 'src/render.jsx',
+  publicPath: '/static/my-app',
   port: 4001,
   dangerouslySetWebpackConfig: (config) => makeStableHashes(config),
 };
--- a/tests/public-path.test.js
+++ b/tests/public-path.test.js
@@ -5,6 +5,8 @@ import {
   runSkuScriptInDir,
   waitForUrls,
   createCancelSignal,
+  getPort,
+  dirContentsToObject,
 } from '@sku-private/test-utils';
 
 import { createRequire } from 'node:module';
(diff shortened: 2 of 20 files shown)


tests/public-path.test.js
```diff
@@ -15,26 +17,45 @@ const appDir = path.dirname(
...
+        });
+
+        it('should create valid app with no unresolved resources', async ({
+          expect,
+        }) => {
+          const app = await getAppSnapshot({ url, expect });
+          expect(app).toMatchSnapshot();
+        });
+
+        it('should generate the expected files', async ({ expect }) => {
```
askoufis, line 54: This test ensures we snapshot CSS output so we can ensure asset URLs are correct.

fixtures/public-path/sku.config.js
```diff
@@ -1,8 +1,9 @@
 import { makeStableHashes } from '@sku-private/test-utils';
 
 export default {
-  publicPath: '/static',
-  target: 'dist/static',
```
askoufis, line 5: I assume the intent of this config was to output static assets into `/static`, but practically this isn't how our apps are typically configured.

williamlark (approved):

PR #1249, opened 2025-05-14
Update ESM-only deps

askoufis:
`sku`s CLI is published as ESM, so the renovate config that disabled updates for packages that became ESM-only has been removed, and the ESM-only deps have been updated to their latest versions.


.github/renovate.json5
```diff
@@ -86,21 +86,5 @@
...
-    {
-      groupName: 'ESM-only packages',
-      matchUpdateTypes: 'major',
-      matchPackageNames: [
-        'chalk',
-        'escape-string-regexp',
-        'find-up',
-        'get-port',
-        'indent-string',
-        'lint-staged',
```
williamlark, line 98: Do we use lint staged?  I noticed only the types package is in devDeps

williamlark (approved):
Does this need a Changeset?

.github/renovate.json5
askoufis: Yep https://github.com/seek-oss/sku/blob/1e55aaad062346010171ec7d0c56a5f18b4b416f/packages/sku/package.json#L157

PR #1250, opened 2025-05-15
`vite start`: Don't clean up local SSL certs every time the dev server starts

askoufis:
Sorry @williamlark. 

While it would be nice to handle, the logic to clean up existing SSL certs makes it necessary to trust the certs in the browser them every time the dev server starts. Removing this logic aligns the vite behaviour with the webpack behaviour.

While it would be nice to handle the situation where an apps hosts change, this doesn't happen that often so IMO isn't worth optimizing for (at the expense of a poorer dev experience). **EDIT**: it turns out that this is really only an issue if you delete the host from the `hosts` config. That host still has a valid certificate, but since it's only used for local dev this isn't a huge concern.

Also did a tiny refactor to pull the conditional plugin logic into a getter.

### DIFF
--- a/packages/sku/src/services/vite/helpers/server/createViteServer.ts
+++ b/packages/sku/src/services/vite/helpers/server/createViteServer.ts
@@ -6,7 +6,7 @@ import { createViteConfig } from '../createConfig.js';
 import skuViteHMRTelemetryPlugin from '@/services/vite/plugins/skuViteHMRTelemetry.js';
 import { skuViteStartTelemetryPlugin } from '../../plugins/skuViteStartTelemetry.js';
 import { getAppHosts } from '@/utils/contextUtils/hosts.js';
-import { getHttpsDevServerPlugin } from '../../plugins/skuViteHttpsDevServer.js';
+import { skuViteHttpsDevServer } from '../../plugins/skuViteHttpsDevServer.js';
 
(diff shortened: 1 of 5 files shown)


packages/sku/src/services/vite/helpers/server/createViteServer.ts
```diff
@@ -22,7 +22,7 @@ export const createViteServer = async (skuContext: SkuContext) =>
           target: 'node',
           type: 'static',
         }),
-        skuContext.httpsDevServer && skuViteHttpsDevServer(skuContext),
+        getHttpsDevServerPlugin(skuContext),
```
williamlark, line 25: Is the name change needed? Would be nice to keep it the same as the others.

```suggestion
        skuViteHttpsDevServer(skuContext),
```

williamlark (approved):

packages/sku/src/services/vite/helpers/server/createViteServer.ts
askoufis: I feel like the original name wasn't really correct though. It never returned a server, it was really a getter for some plugins. It's still that, so I figured I'd name it as such. Honestly the `skuVite` prefix is redundant, and I want to rename the others as well. Happy to undo the rename in this PR and rename everything separately though.

PR #1251, opened 2025-05-15
Rename vite plugins

askoufis:
Followup from #1250.

IMO the `skuVite` prefix on these plugins is redundant. They're already in a `vite/plugins` directory. Also default export bad :caveman:.

Additionally, `skuViteHttpsDevServer` does not return a dev server, but rather is a getter for a plugin, so it has been renamed to reflect that.


jahredhope (approved):

DanDroryAu (approved):
`skuVite` is dead, long live the plugins!

packages/sku/src/services/vite/helpers/server/createViteServer.ts
```diff
@@ -1,28 +1,29 @@
...
           type: 'static',
         }),
-        skuViteHMRTelemetryPlugin({
+        // eslint-disable-next-line new-cap
+        HMRTelemetryPlugin({
           target: 'node',
           type: 'static',
         }),
-        skuViteHttpsDevServer(skuContext),
+        getHttpsDevServerPlugin(skuContext),
```
williamlark, line 26: I feel like this should still just be `httpsDevServerPlugin`. Vite plugins can be nullable, so defining it as a getter is more of an implementation detail.

williamlark (approved):

packages/sku/src/services/vite/helpers/server/createViteServer.ts
askoufis: Yeah good point.

PR #1252, opened 2025-05-15
Convert `sku-webpack-plugin` tests to TypeScript

askoufis:
We should go full typescript 😎. Sick of not having types ✨ _sometimes_ ✨ .


jahredhope (approved):
idk, I'm not sold on TypeScript. What's wrong with Flow?

PR #1254, opened 2025-05-15
Convert all test utils to typescript

askoufis:
Notable changes:
- Replaced [`node-dir`](https://www.npmjs.com/package/node-dir) with [`klaw`](https://www.npmjs.com/package/klaw). We use `klaw` in an internal repo and its API is a bit nicer than `node-dir`s (and it has 0 deps).
- Added `type: "module"` to the `styling` and `typescript-css-module` fixtures to keep `tsc` happy (note that these fixtures are not tested against vite yet)
- Added `tsx` as a dep so we can run a script written in TypeScript (`--experimental-strip-types` is only in node 22 😭)

Everything else is pretty much a change to make `tsc` happy

Friendship ended


tsconfig.json
```diff
@@ -8,8 +8,7 @@
     // Output incremental build info to node_modules
     "outDir": "node_modules/.cache/typescript",
     "incremental": true,
-    "resolveJsonModule": true,
-    "types": ["vitest-puppeteer"]
```
askoufis, line 12: This didn't seem to actually do anything. In the one file that `tsc` complained about, `test-utils/storybook.ts`, a side-effect import of `import 'vitest-puppeteer'` was enough to inject the necessary global types.

tests/typescript-css-modules.test.ts
```diff
@@ -76,7 +71,7 @@ describe.sequential('typescript-css-modules', () => {
         args: ['--config=sku-ssr.config.ts'],
       });
       run('node', {
-        args: ['server'],
+        args: ['server.cjs'],
```
askoufis, line 74: Now that the `typescript-css-modules` fixture is `type: "module"`, it outputs a `server.cjs` file, but `node server` tries to find `server.js`, so we need an explicit file extension here.

test-utils/dirContentsToObject.ts
```diff
@@ -0,0 +1,50 @@
+import { relative } from 'node:path';
+import { readFile } from 'node:fs/promises';
+
+import klaw from 'klaw';
```
DanDroryAu, line 4: claw-the

DanDroryAu (approved):

PR #1260, opened 2025-05-20
Actually test vite in the `public-path` test

askoufis:
Accidentally forgot to pass the bundler args in the `public-path` test, so we were testing the webpack build twice. Snapshot have been updated accordingly and they look fine to me.

### DIFF
--- a/tests/public-path.test.ts
+++ b/tests/public-path.test.ts
@@ -20,42 +20,40 @@ const appDir = path.dirname(
 const distDir = path.resolve(appDir, 'dist');
 
 describe('public path', () => {
-  describe.concurrent.each(['vite', 'webpack'])(
-    'bundler: %s',
-    async (bundler) => {
-      const port = await getPort();
-      const args = ['--strict-port', `--port=${port}`];
-      if (bundler === 'vite')
-        args.push(
-          ...['--experimental-bundler', '--config', 'sku.config.vite.js'],
-        );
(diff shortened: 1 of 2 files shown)


tests/public-path.test.ts
```diff
@@ -20,42 +20,40 @@ const appDir = path.dirname(
...
-
-      describe('build and serve', () => {
-        const url = `http://localhost:${port}`;
-        const { cancel, signal } = createCancelSignal();
-
-        beforeAll(async () => {
-          await runSkuScriptInDir('build', appDir);
-          runSkuScriptInDir('serve', appDir, { args, signal });
-          await waitForUrls(url);
+  describe.each(['vite', 'webpack'])('bundler: %s', async (bundler) => {
```
williamlark, line 23: Does work fine, though I'm not sure if we should put `.sequential` here or not. If these run concurrently and try to build over each other there could be issues. There isn't any problems now, so approving, but something to monitor. I couldn't see in the Vite docs if it runs concurrently or sequential by default for `.each` 🤷

williamlark (approved):

tests/public-path.test.ts
askoufis: `concurrent` was breaking the tests, I assume because both the `webpack` and `vite` build artifacts were overwriting/fighting each other. So it _seems_ like `sequential` is the default.

PR #1261, opened 2025-05-20
Strip assertions during `vite` builds

askoufis:
During production webpack builds, assertions are stripped from the bundle. Currently this is handled by https://github.com/unassert-js/babel-plugin-unassert. Vite builds need to emulate this behaviour.

For Vite there are 3 viable options: 
1. the babel plugin we already use
2. https://github.com/unassert-js/rollup-plugin-unassert
3. https://github.com/unassert-js/swc-plugin-unassert

Given that we likely want to run a few more babel plugins during vite builds, I chose to opt for the babel plugin. I would've liked to use the SWC plugin, but the plugin doesn't quite support the features we need (specifically custom assert methods and modules).

There are 2 easy ways to run babel plugins in Vite:
1. https://www.npmjs.com/package/vite-plugin-babel
2. https://www.npmjs.com/package/@vitejs/plugin-react

While option 1 seems like the obvious choice, by switching from `@vitejs/plugin-react-swc` to `@vitejs/plugin-react`, we can pass the babel plugin directly to it (`@vitejs/plugin-react-swc` only supports SWC plugins).

We may one day want to switch entirely to SWC, but until then it will be much easier to achieve feature parity with webpack by sticking to babel.

### DIFF
--- a/tests/assertion-removal.test.ts
+++ b/tests/assertion-removal.test.ts
@@ -26,7 +26,7 @@ assert(skuConfig.serverPort, 'skuConfig has serverPort');
 const backendUrl = `http://localhost:${skuConfig.serverPort}`;
 
 describe('assertion-removal', () => {
-  describe.for(['webpack'])('bundler %s', (bundler) => {
+  describe.for(['vite', 'webpack'])('bundler %s', (bundler) => {
     describe('build', async () => {
       const port = await getPort();
       const url = `http://localhost:${port}`;
--- a/tests/assertion-removal.test.ts
+++ b/tests/assertion-removal.test.ts
@@ -7,6 +7,7 @@ import {
   startAssetServer,
   run,
   createCancelSignal,
+  getPort,
 } from '@sku-private/test-utils';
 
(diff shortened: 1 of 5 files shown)


tests/assertion-removal.test.ts
```diff
@@ -25,30 +26,39 @@ assert(skuConfig.serverPort, 'skuConfig has serverPort');
 const backendUrl = `http://localhost:${skuConfig.serverPort}`;
 
 describe('assertion-removal', () => {
-  describe('build', () => {
-    const url = 'http://localhost:8239';
-    const { cancel, signal } = createCancelSignal();
+  describe.for(['webpack'])('bundler %s', (bundler) => {
```
williamlark, line 29:
```suggestion
  describe.for(['vite', 'webpack'])('bundler %s', (bundler) => {
```

williamlark (approved):

tests/assertion-removal.test.ts
askoufis: Good catch! https://github.com/seek-oss/sku/pull/1261/commits/f43245e4ba0f743519283838cf2fdc7b5471a443

PR #1263, opened 2025-05-21
Fix CLI docs typo

askoufis:
My spellchecker doesn't work inside backticks 😭.


jahredhope (approved):
LGTM, as long as the buid passes

PR #1267, opened 2025-05-30
Adding test for svg images

williamlark:
Our public path test checks that an assortment of images are imported by sku correctly. SVGs were missing from this test, and testing these will be useful when we add SVG optimization to the Vite bundler. 

I've also modified the error output of the process command. It will now spit out proper errors when it closes from anything other than an abort signal, making debugging failing tests easier. An example of the new error output for a failing test while debugging is below:

CleanShot 2025-05-30 at 10 04 51


test-utils/process.ts
```diff
@@ -24,9 +24,9 @@ export const run = async (
   } catch (error) {
     if (error instanceof SubprocessError) {
       if (error.cause instanceof Error && error.cause.name === 'AbortError') {
-        console.log(error.cause.cause);
         return;
       }
+      console.error(error.output);
```
askoufis, line 29: 👏

askoufis (approved):

PR #1271, opened 2025-06-03
`sku test`: add `vitest` experimental support.

DanDroryAu:
## Copilot summary

This pull request introduces experimental support for the `vitest` test runner in the `sku` framework, along with associated changes to configuration, dependencies, and tooling. The most important changes include adding the `vitest` runner, updating the `sku` test command to support both `jest` and `vitest`, and creating a new package for the `vitest` runner. Additionally, several configuration and dependency updates were made to support this integration.

### Experimental `vitest` Support:

* Added a new experimental configuration option `__UNSAFE_EXPERIMENTAL__testRunner` in `sku.config.ts` to allow switching between `jest` and `vitest` as the test runner. [[1]](diffhunk://#diff-2f4a6b374e16490714bd0e13cca89d4c0e0bb891e31212bbda85b31173734d55L1-R3) [[2]](diffhunk://#diff-79b5cff5cca2e8897ebb5d6196337cc83e185d6c32fa01b618aeabf4edfcd90bR1-R4) [[3]](diffhunk://#diff-7eba55a3c5050670cd645ca7c1da7e6d9898eefb531d7ccac78b9e1901f08be9R121-R129)
* Implemented a new `vitestHandler` in `vitest-test-handler.ts` to handle `vitest` execution, including dynamic installation of the `@sku-lib/vitest` package if not already installed.
* Updated the `testAction` logic in `test.action.ts` to delegate test execution to either `jest` or `vitest` based on the configuration. [[1]](diffhunk://#diff-91573e325f14805612e623dbe313d15cf3f16fbc8757e46d9cdc12e7be9b2fc8L1-R5) [[2]](diffhunk://#diff-91573e325f14805612e623dbe313d15cf3f16fbc8757e46d9cdc12e7be9b2fc8L24-R21)

### New `@sku-lib/vitest` Package:

* Created a new `@sku-lib/vitest` package to encapsulate the `vitest` test runner logic, including its configuration and setup. [[1]](diffhunk://#diff-ad8a38817b45fa1cbf3728d70c2da9fb62d3540f236e561fa5c94725a9f71da2R1-R43) [[2]](diffhunk://#diff-5d0c83270bf116f14a9694aa377d727cf2ea5ce9c2131293e46350f4c1720a73R1-R33)
[...]

PR #1273, opened 2025-06-04
Vitest fixes

DanDroryAu:
Fix for the `cli` parsing part of the new `vitest` runner for `sku`.
Fix the peer dependencies for `@sku-lib/vitest`.
Fix error handling for the installation prompt part of `@sku-lib/vitest`.

### DIFF
--- a/packages/sku/src/program/commands/test/vitest-test-handler.ts
+++ b/packages/sku/src/program/commands/test/vitest-test-handler.ts
@@ -10,13 +10,9 @@ export const vitestHandler = async ({
   skuContext: SkuContext;
   args: string[];
 }) => {
+  let vitestImport = null;
   try {
-    const { runVitest } = await import('@sku-lib/vitest');
-    await runVitest({
-      setupFiles: skuContext.paths.setupTests,
-      args,
-    });
-    return;
+    vitestImport = await import('@sku-lib/vitest');
   } catch (e: any) {
     if (e.code !== 'ERR_MODULE_NOT_FOUND' || isCI) {
       console.error(e.message);
@@ -45,4 +41,12 @@ export const vitestHandler = async ({
     // Retry running Vitest after installation
     await vitestHandler({ skuContext, args });
   }
+  if (!vitestImport) {
+    console.error('Failed to import @sku-lib/vitest');
+    return;
+  }
+  await vitestImport.runVitest({
+    setupFiles: skuContext.paths.setupTests,
+    args,
+  });
 };
--- a/tests/__snapshots__/sku-test.test.ts.snap
+++ b/tests/__snapshots__/sku-test.test.ts.snap
@@ -14,14 +14,26 @@ All files |       0 |        0 |       0 |       0 |
 `;
 
 exports[`[vitest]: sku-test > should pass through unknown flags to vitest 1`] = `
-"  console.log
-    running setup test
+"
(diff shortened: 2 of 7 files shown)


tests/__snapshots__/sku-test.test.ts.snap
```diff
@@ -14,14 +14,26 @@ All files |       0 |        0 |       0 |       0 |
 `;
 
 exports[`[vitest]: sku-test > should pass through unknown flags to vitest 1`] = `
-"  console.log
-    running setup test
+"
```
askoufis, line 17: This snapshot output could be cleaned up with https://nodejs.org/api/util.html#utilstripvtcontrolcharactersstr

askoufis (commented):
`sku` needs to move it's `vitest` dep to a dev dep, and `@sku-lib/vitest` needs to properly depend on `vitest` and the VE plugin too.

tests/__snapshots__/sku-test.test.ts.snap
DanDroryAu: I've got to strip the timestamp out of this snapshot as well.

packages/sku/src/program/commands/test/vitest-test-handler.ts
```diff
@@ -45,4 +41,12 @@ export const vitestHandler = async ({
     // Retry running Vitest after installation
     await vitestHandler({ skuContext, args });
   }
+  if (!vitestImport) {
```
williamlark: Why does this need to live outside of the try block?

williamlark (commented):
Needs a chnageset too. Patch release?

askoufis:
> Needs a chnageset too. Patch release?

I mentioned in [the previous PR](https://github.com/seek-oss/sku/pull/1271#pullrequestreview-2890936128) that the release notes for `@sku-lib/vitest` should probably be split out to just mention that it's an initial release.

packages/sku/src/program/commands/test/vitest-test-handler.ts
DanDroryAu: It's still possibly undefined. I can probably clean this up because I need to return at the end of the vitest runner because it might run it twice.

DanDroryAu: I.E. on line 42 it will re-try to run it and afterwards still enter the failed state.

askoufis (approved):
Worth adding a "do not use in production" message of some kind to the `sku` changeset.

tests/sku-test.test.ts
```diff
@@ -28,6 +29,10 @@ describe.for(['vitest', 'jest'])('[%s]: sku-test', (testRunner) => {
     });
     const output = (child?.stdout as string).replaceAll(cwd(), 'sku');
 
-    expect(output).toMatchSnapshot();
+    expect(
+      stripVTControlCharacters(output)
+        .replaceAll(/(\d+\.?\d*)s|(\d*)ms/g, '0ms')
+        .replaceAll(/\b\d{1,2}:\d{2}:\d{2}\b/g, '00:00:00'),
```
williamlark, line 35: Probably better to use fake timers here so we don't need to mess with regex stuff

williamlark (approved):

tests/sku-test.test.ts
DanDroryAu: I tested this and it doesn't work. The output time is still the same.

PR #1274, opened 2025-06-05
split vite config creation into multiple configs

DanDroryAu:
The `vite` base config was getting complex with lots of conditional logic. This PR tries to remove some of the conditional logic and uses the `mergeConfig` function provided by `vite` to deep merge configurations.

This also fixes an issue where the `vite-client.js` file was not getting a hash affixed.


packages/sku/src/services/vite/helpers/config/baseConfig.ts
```diff
@@ -1,44 +1,25 @@
...
-    ssg: skuContext.paths.renderEntry,
-  };
+const getBaseConfig = (skuContext: SkuContext): InlineConfig => {
   const vocabConfig = getVocabConfig(skuContext);
-  const isStartCommand = Boolean(skuContext.commandName?.startsWith('start'));
-
-  const isProductionBuild = process.env.NODE_ENV === 'production';
-
   return {
-    base: isStartCommand ? '/' : skuContext.publicPath,
```
askoufis, line 41: This logic has to remain. It ensures that the local dev server doesn't fuss around with the `base` URL at all, but built assets adhere to `publicPath`. This aligns with the sku's webpack dev behaviour.

packages/sku/src/services/vite/helpers/server/createViteServerSsr.ts
```diff
@@ -38,23 +38,25 @@ export const createViteServerSsr = async ({
...
+              hmr: true,
+              allowedHosts: getAppHosts(skuContext).filter(
+                (host) => typeof host === 'string',
+              ),
+            },
+            plugins: [httpsDevServerPlugin(skuContext)],
+            appType: 'custom',
+            base,
+          },
+          false,
```
askoufis: Shouldn't this (and in other usages of `mergeConfig`) flag be `true` (or omitted since `true` is the default) since we're merging the config at the top level? https://vite.dev/guide/api-javascript#:~:text=for%20example%2C%20set%20false%20if%20you're%20merging%20two%20build%20options

packages/sku/src/services/vite/helpers/server/createViteServer.ts
```diff
@@ -1,35 +1,40 @@
...
-      host: 'localhost',
-      allowedHosts: getAppHosts(skuContext).filter(
-        (host) => typeof host === 'string',
-      ),
-    },
-  });
+  createServer(
+    mergeConfig(
+      createViteClientConfig(skuContext),
+      {
```
askoufis: Can this second config be inlined into `mergeViteClientConfig` (ditto for the ssr and ssg configs)?

packages/sku/src/services/vite/helpers/config/baseConfig.ts
DanDroryAu: I'm overriding the `base` with `/` on the server config though which should be the same behaviour. this way we don't need to know if it's a start command yes or no.

packages/sku/src/services/vite/helpers/server/createViteServer.ts
DanDroryAu: I've opted into defining all the configs in the `createConfig` file.

packages/sku/src/services/vite/helpers/config/baseConfig.ts
askoufis: Oh right, none of the functions in `createConfig` are used for `sku start`, so `base: '/'` doesn't get overridden. Makes sense 👍.

DanDroryAu: Yeah only in the `createViteServer` file do I override it with the correct base. However all of that config stuff will be in the `createConfig.ts` file after my next push.

packages/sku/src/services/vite/helpers/config/createConfig.ts
williamlark, line 1: `nit`: Could be nice for the configs to be sorted alphabetically with an eslint rule or something, but not that important haha

williamlark (approved):
Nice

packages/sku/src/services/vite/helpers/config/createConfig.ts
```diff
@@ -0,0 +1,122 @@
...
+        host: 'localhost',
+        middlewareMode: true,
+        hmr: true,
+        allowedHosts: getAppHosts(skuContext).filter(
+          (host) => typeof host === 'string',
+        ),
+      },
+      plugins: [httpsDevServerPlugin(skuContext)],
+      appType: 'custom',
+      base: process.env.BASE || '/',
```
askoufis: Can this `process.env.BASE` be removed? It's in the vite SSR templates, but we don't want to support it (`publicPath` should be the only way to configure `base`) so it shouldn't be used.

PR #1275, opened 2025-06-05
throw an error when the specified config file is not found

DanDroryAu:
`sku` now throws an error when the `--config` option is passed and the config file can't be read. 

`sku` would only let the user know of this mistake if they were logging with the `--debug` flag before and may cause unintended side effects.

This change makes sure that `sku` exits the program when it can't find the specified file.

### DIFF
--- a/.changeset/easy-beds-argue.md
+++ b/.changeset/easy-beds-argue.md
@@ -1,0 +1,10 @@
+---
+'sku': major
+---
+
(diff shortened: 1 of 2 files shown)


askoufis (commented):
Is this a change for later? Wasn't planning on breaking any time soon.

DanDroryAu:
We can keep it for later if we don't want to do a breaking change. I'll just close this PR.

PR #1276, opened 2025-06-05
adding some sku webpack apis into sku vites

williamlark:
Making a few changes to increase backwards-compatibility of sku webpack features. I'm not expecting that all of these will be kept or implemented in the way proposed, but am more raising to kick of discussions. Can keep any major chats outside of the PR comments. 

**Overview**

* Removed `tsconfig.json` from the published files. I don't think required to be published and keeps causing errors when using yalc.
* Adding **dangerouslySetViteConfig** to mimic the Webpack sku api. We're using this in Braid Design Systems docs site and need a way to support the custom changes it's making.
* Adding `babel-plugin-macros`. It was used in the Webpack config and is needed to get the Braid Design System docs working.
* Passing in `compilePackages` to the sku vite config. This was configurable in the webpack api and we should include a way to set it in Vite.
* Adding `skipPackageCompatibilityCompilation`. Another webpack api I'm adding in since Braid Design System docs was using it. Unsure if I've added it to the right place, or if its even needed. 
* Removing `builtinModules` from being added to Vites externals since it's added in by default by Vite.


packages/sku/src/context/defaultSkuConfig.ts
```diff
@@ -36,6 +36,7 @@ export default {
   dangerouslySetJestConfig: defaultDecorator,
   dangerouslySetESLintConfig: defaultDecorator,
   dangerouslySetTSConfig: defaultDecorator,
+  dangerouslySetViteConfig: defaultDecorator,
```
DanDroryAu: ☠️

packages/sku/src/services/vite/helpers/createConfig.ts
```diff
@@ -107,19 +109,20 @@ export const createViteConfig = ({
...
         'serialize-javascript',
         'used-styles',
         '@sku-lib/vite',
-        ...(configType === 'ssg' || configType === 'ssr'
-          ? ['sku/vite/loadable']
-          : []),
       ],
-      noExternal: ['@vanilla-extract/css', 'braid-design-system'],
+      noExternal: [
+        '@vanilla-extract/css',
```
askoufis: This a already marked as external by the Vanilla Extract plugin, so it should just be removed IMO.

packages/sku/src/types/types.ts
```diff
@@ -226,6 +226,22 @@ export interface SkuConfig {
...
+   *
+   * As sku creates two Vite configs (`client` & `server|render`), this function will actually run twice.
+   * If you only need to modify one of these configs, then you can check `config.mode` within.
+   *
+   * Sku provides no guarantees that its Vite configuration will remain compatible with any customizations made within this function.
+   * It is the responsibility of the user to ensure that their customizations are compatible with sku.
+   *
+   * @link https://seek-oss.github.io/sku/#/./docs/configuration?id=dangerouslysetviteconfig
+   */
+  dangerouslySetViteConfig?: (skuViteConfig: any) => any;
```
askoufis: We can probably provide a better type for this. Vite's config is well defined and sane, unlike webpacks config. Also, it would be nice to _only_ make this configurable if you've configured vite.

I would also prefer this be marked as unstable or something. I'm still not 100% convinced that it's needed (other than for debugging purposes).

askoufis: I guess this could have the same type as the `config` plugin hook?

DanDroryAu: The type is `InlineConfig` from `vite`.

packages/sku/src/services/vite/helpers/createConfig.ts
```diff
@@ -107,19 +109,19 @@ export const createViteConfig = ({
...
       },
+      exclude: skuContext.skuConfig.skipPackageCompatibilityCompilation,
     },
     ssr: {
       external: [
-        ...builtinModules,
         '@vanilla-extract/css/adapter',
         'serialize-javascript',
         'used-styles',
         '@sku-lib/vite',
```
askoufis: Could remove the vanilla extract adapter (can't mark entrypoints as external anyway) and used-styles (unused in sku) from here too if you want.

packages/sku/src/types/types.ts
```diff
@@ -226,6 +227,22 @@ export interface SkuConfig {
...
+   *
+   * As sku creates two Vite configs (`client` & `server|render`), this function will actually run twice.
+   * If you only need to modify one of these configs, then you can check `config.mode` within.
+   *
+   * Sku provides no guarantees that its Vite configuration will remain compatible with any customizations made within this function.
+   * It is the responsibility of the user to ensure that their customizations are compatible with sku.
+   *
+   * @link https://seek-oss.github.io/sku/#/./docs/configuration?id=dangerouslysetviteconfig
+   */
+  dangerouslySetViteConfig?: (skuViteConfig: InlineConfig) => any;
```
askoufis: Can this use [the type from the `config` plugin hook](https://vite.dev/guide/api-plugin#config), since it's just being passed into that hook anyway? This would provide `mode` and `command`, both of which can be used to scope the config changes to certain builds.

Also, I'd prefer this to be prefixed with `unstable` or something similar until we're sure we want to stabilize it.

packages/sku/src/types/types.ts
```diff
@@ -226,6 +227,22 @@ export interface SkuConfig {
...
   dangerouslySetWebpackConfig?: (skuWebpackConfig: any) => any;
 
+  /**
+   * This function provides a way to modify sku's Vite configuration.
+   * It should only be used in exceptional circumstances where a solution cannot be achieved by adjusting standard configuration options.
+   *
+   * Before customizing your Vite configuration, please reach out in [#sku-support](https://seek.enterprise.slack.com/archives/CDL5VP5NU) to discuss your requirements and potential alternative solutions.
+   *
+   * As sku creates two Vite configs (`client` & `server|render`), this function will actually run twice.
+   * If you only need to modify one of these configs, then you can check `config.mode` within.
```
askoufis: I'm not sure if this is correct. `mode` in the config is used to override the configured `--mode` flag. It's also typically either `production` or `development`, rather than `client|server|render`. As mentioned in my other comment, if the full `config` hook signature is given, that would enable declaring specific config for a specific build.

askoufis (commented):
This needs a changeset (or multiple) for the new features being added.

.changeset/dark-eagles-film.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+Adding `babel-plugin-unassert` to the Vite babel plugins
```
askoufis: `babel-plugin-unassert` was added previously.

```suggestion
Adding `babel-plugin-macros` to the Vite babel plugins
```

williamlark: Thank you. Good catch. Copied my PR description which was also wrong 🤦

askoufis (approved):
Some snapshot errors need addressing, but otherwise LGTM.

PR #1278, opened 2025-06-06
Revert addition of `--disable-warning` flag to sku binary

askoufis:
Alpine linux's `env` (from busybox) [doesn't support the `-S` flag](https://forum.gitlab.com/t/error-usr-bin-env-unrecognized-option-s-with-alpine-linux-image-causes-ci-script-to-fail/64063), which means we can't use it to suppress warnings in sku. There are methods to make it work, but they aren't worth implementing for the small value they add.

As we move on to newer Node versions (eventually), the need to suppress these warnings will go away.

### DIFF
--- a/.changeset/shaky-banks-melt.md
+++ b/.changeset/shaky-banks-melt.md
@@ -1,0 +1,7 @@
+---
+'sku': patch
+---
+
+Revert addition of `--disable-warning` flag to `sku` binary
+
+This flag was added to suppress certain warnings about experimental features being used by `sku`. However, the method use to set this flag is not compatible with certain distributions of Linux without additional configuration, so it has been removed.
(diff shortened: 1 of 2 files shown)


williamlark:
> > It looks like there was a [marked solution](https://forum.gitlab.com/t/error-usr-bin-env-unrecognized-option-s-with-alpine-linux-image-causes-ci-script-to-fail/64063/11) for this in the thread you linked. Is this an issue you've ran into or been reported by someone using sku?
> 
> Yes it was [reported internally](https://seekchat.slack.com/archives/CDL5VP5NU/p1749164695272549). While there is a solution, forcing all consumers to update their dockerfiles just to suppress a few warnings doesn't feel worth it to me.

Yea fair enough if someone's reported it. Good find

PR #1280, opened 2025-06-11
`vite (start)`: `initialPath` will now render when starting the server

williamlark:
The `initialPath` in the sku config makes the server start at a specific route when running `sku start`. This works in the experimental Vite bundler, but it would not show the path in the CLI output. This means that if your app has no root page and you CMD + Click to open the output url from the CLI you will be greeted with a 404 error. 

This PR adds the initial route to the CLI output so the correct url can be opened easier.

**Before**

CleanShot 2025-06-11 at 14 04 59

**After**

CleanShot 2025-06-11 at 14 06 25

### DIFF
--- a/packages/sku/src/services/vite/index.ts
+++ b/packages/sku/src/services/vite/index.ts
@@ -71,8 +71,10 @@ const printUrls = (
 ) => {
   const proto = skuContext.httpsDevServer ? 'https' : 'http';
   hosts.forEach((site) => {
-    console.log(
-      `${chalk.green('➜')}  ${chalk.bold('Local')}: ${chalk.cyan(`${proto}://${site}:${chalk.bold(skuContext.port.client)}`)}`,
+    const initialPath = chalk.dim(`${skuContext.initialPath}`);
+    const url = chalk.cyan(
+      `${proto}://${site}:${chalk.bold(skuContext.port.client)}${initialPath}`,
     );
+    console.log(`${chalk.green('➜')}  ${chalk.bold('Local')}: ${url}`);
   });
 };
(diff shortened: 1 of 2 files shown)


williamlark:
Opted to show it dimmed since the port and base path seem like more important information and make those values easier to scan. No strong feelings though if we'd like to mimic the webpack behaviour and show it as the same opacity as the rest of the url

packages/sku/src/services/vite/index.ts
```diff
@@ -71,8 +71,10 @@ const printUrls = (
 ) => {
   const proto = skuContext.httpsDevServer ? 'https' : 'http';
   hosts.forEach((site) => {
-    console.log(
-      `${chalk.green('➜')}  ${chalk.bold('Local')}: ${chalk.cyan(`${proto}://${site}:${chalk.bold(skuContext.port.client)}`)}`,
+    const initialPath = chalk.dim(`${skuContext.initialPath}`);
```
askoufis: Honestly not a fan of `dim` here. It makes it hard to see the text (at least in your screenshot). Does `vite` use `dim` here? Does it even support configuring an initial path?

williamlark: Doesn't seem like they support an initial path, only a base path which is not quite the same. I agree it makes it harder to read, but my intention was to make it less pronounced than the rest of the url since I don't expect people to read it in the first place (it's defined in their config and they will see it in the browser anyway).

Though, I'm not too fussed. It looks fine not dimmed (below). Happy to do that 👌 

CleanShot 2025-06-11 at 14 22 53

williamlark: In fact, the bolding of the port is a nice separator of the path. I think I prefer it not dimmed now that I look at it 😂

askoufis: While I'm not _too_ concerned with console a11y, `dim` just ruins contrast in the console, so I tend to avoid it.

williamlark: Done, _let there be light_

askoufis (approved):

PR #1283, opened 2025-06-15
Declare `@sku-lib/vite` as a dependency

askoufis:
Unless I'm missing something, I think `@sku-lib/vite` was accidentally made a dev dep instead of a regular dep. We don't dynamically import it, and it's not declared as a peer dependency, so we should declare it as a regular dependency. Have seen CI errors in renovate PRs caused by this.


askoufis:
Ci will probably fail until #1284 goes through.

DanDroryAu (approved):

PR #1284, opened 2025-06-16
Fix `sku-init` test pollution

askoufis:
This PR utilizes vitest [`projects`](https://vitest.dev/guide/projects.html#test-projects) (previously named `workspaces`) to separate tests by environment, as well as isolate `sku-init` from the rest of the tests as I believe it is occasionally causing test pollution because it runs `pnpm install`.

I think this is happening in #1283. It seems to only happen when there are new non-major versions of specifically the packages installed by `sku init`.

I've defined 3 projects in the vitest config:
- `sku`: for unit tests within the `sku` package that run in the default `node` environment
- `sku-puppeteer`: for e2e tests that require `puppeteer`
- `sku-init`: to isolate the `sku-init` test and run it last to ensure it doesn't pollute dependencies and cause issues in other tests

### DIFF
--- a/tests/sku-init/sku-init.test.ts
+++ b/tests/sku-init/sku-init.test.ts
@@ -44,6 +44,7 @@ describe('sku init', () => {
 
   afterAll(async () => {
     await fs.rm(projectDirectory, { recursive: true, force: true });
+    spawnSync('git', ['restore', 'pnpm-lock.yaml']);
     console.log(
       "Running 'pnpm install' to clean up lockfile after sku-init test...",
     );
--- a/tests/sku-init/sku-init.test.ts
+++ b/tests/sku-init/sku-init.test.ts
@@ -9,13 +9,14 @@ const __dirname = path.dirname(new URL(import.meta.url).pathname);
 const fixtureDirectory = __dirname;
 const projectName = 'new-project';
 const projectDirectory = path.join(fixtureDirectory, projectName);
+const lockFilePath = path.join(process.cwd(), 'pnpm-lock.yaml');
 
 describe('sku init', () => {
-  let stdout: string;
-  let stderr: string;
+  let pnpmLockFile: string;
 
   beforeAll(
     async () => {
+      pnpmLockFile = await fs.readFile(lockFilePath, 'utf-8');
       await fs.rm(projectDirectory, { recursive: true, force: true });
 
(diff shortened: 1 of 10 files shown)


vitest.config.ts
```diff
@@ -1,24 +1,53 @@
-import { defineConfig } from 'vitest/config';
+import { defaultExclude, defineConfig } from 'vitest/config';
 import tsconfigPaths from 'vite-tsconfig-paths';
 
 export const TEST_TIMEOUT = 50_000;
+const defaultInclude = '**/*.{test,spec}.?(c|m)[jt]s?(x)';
```
askoufis, line 5: I wanted to prepend this default include string with another path, so chose not to use `defaultInclude` from `vitest/config` as it's an array.

tests/sku-init/sku-init.test.ts
```diff
@@ -44,6 +44,7 @@ describe('sku init', () => {
 
   afterAll(async () => {
     await fs.rm(projectDirectory, { recursive: true, force: true });
+    spawnSync('git', ['restore', 'pnpm-lock.yaml']);
```
askoufis: I think in the past just running `pnpm install` after deleting the directory created by `sku init` was enough to restore the lockfile to its previous state, but this doesn't occur consistently, so I've opted to restore the lockfile before running `pnpm install`.

tests/sku-test.test.ts
```diff
@@ -31,6 +31,7 @@ describe.for(['vitest', 'jest'])('[%s]: sku-test', (testRunner) => {
 
     expect(
       stripVTControlCharacters(output)
+        .replaceAll(/v\d\.\d\.\d /g, '')
```
askoufis, line 34: This replaces the `vitest` version number in the snapshot so it doesn't need to be udpated every time `vitest` is updated.

tests/sku-init/sku-init.test.ts
askoufis: The downside of this is that it will destroy any lockfile changes you have in your working tree. Maybe we store the lockfile, run the tests, then write it back and `pnpm install`, just to ensure nothing changes?

askoufis: Implemented in https://github.com/seek-oss/sku/pull/1284/commits/b63bca875cf025bfe5d1ef2c927a96006e520cc3

PR #1285, opened 2025-06-17
Local eslint rule. Use scope expect not global expect.

DanDroryAu:
add an eslint rule that enforces the use of the local `expect` that is available on the `context` object in the `it|test` callback.

Correct usage:
```
import { it } from 'vitest';

it('tests', ({ expect }) => {
  // ... my test here ...
});
```

Incorrect usage:
```
import { it, expect } from 'vitest';

it('tests', () => {
  // ... my test here ...
});
```

If the global expect is still required it can be circumvented by renaming it.
```
import { it, expect as globalExpect } from 'vitest';

it('tests', () => {
  globalExpect(true).toBe(true);
});
```

If the Eslint AST parser allows me to manipulate and create a new AST node and write it back to the old one without having to do string manipulation I would like to know that as feedback so I can improve some of the function calls.


test-utils/eslint-plugin/use-vitest-local-expect.cjs
```diff
@@ -0,0 +1,122 @@
...
+        // Check if `expect` is locally imported
+        if (node.source.value === 'vitest') {
+          const importedExpect = node.specifiers.some(
+            (specifier) => specifier.local.name === 'expect',
+          );
+          if (importedExpect) {
+            // Find the expect specifier. This is where eslint will report the issue.
+            const expectSpecifier = node.specifiers.find(
+              (specifier) => specifier.local.name === 'expect',
+            );
```
askoufis: `importedExpect` and `expectSpecifier` are the same here right? Could join the two `if` statements if so.

test-utils/eslint-plugin/use-vitest-local-expect.cjs
```diff
@@ -0,0 +1,122 @@
...
+                fix(fixer) {
+                  return fixer.remove(node);
+                },
+              });
+            } else {
+              // If there are other imports, remove only `expect`
+              const newSpecifiers = node.specifiers
+                .filter((specifier) => specifier.local.name !== 'expect')
+                .map((specifier) => specifier.local.name)
+                .join(', ');
```
askoufis: This currently doesn't handle import renames. IMO it's worth trying to modify existing source code rather than reconstructing it from scratch.

```suggestion
              const { sourceCode } = context;
              const newSpecifiers = node.specifiers
                .filter((specifier) => specifier.local.name !== 'expect')
                .map((specifier) => sourceCode.getText(specifier))
                .join(', ');
```

test-utils/eslint-plugin/use-vitest-local-expect.cjs
```diff
@@ -0,0 +1,122 @@
...
+                  // Report and autofix to add `expect` to the context object
+                  context.report({
+                    node,
+                    message:
+                      '`expect` should be accessed from the `it` or `test` function callback.',
+                    fix(fixer) {
+                      return fixer.replaceText(
+                        contextParam[0],
+                        `{${contextParam[0].properties.map((prop) => prop.key.name).join(', ')}, expect}`,
+                      );
```
askoufis: Same here, renamed object properties aren't handled correctly.

```suggestion
                      const { sourceCode } = context;
                      const paramText = sourceCode.getText(contextParam[0]);
                      const newSource = `${paramText.slice(0, -1)}, expect}`;
                      return fixer.replaceText(contextParam[0], newSource);
```

test-utils/eslint-plugin/use-vitest-local-expect.cjs
```diff
@@ -0,0 +1,122 @@
+module.exports = {
```
askoufis, line 2: Having some types, even if they're not checked, will be handy.

```suggestion
/** @type {import("eslint").Rule.RuleModule} */
module.exports = {
```

test-utils/eslint-plugin/use-vitest-local-expect.test.js
askoufis, line 1: We could use https://github.com/antfu/eslint-vitest-rule-tester for these so we can at least integrate them into vitest.

test-utils/eslint-plugin/use-vitest-local-expect.test.js
```diff
@@ -0,0 +1,66 @@
...
+    );
+  });
+
+  it('invalid case: no expect accessed from context', async ({ expect }) => {
+    const { result } = await invalid({
+      code: "it('should do something', () => { expect(true).toBe(true); });",
+      errors: ['correctExpect'],
+    });
+
+    expect(result.output).toMatchSnapshot();
```
askoufis: Would inline snapshots be more useful here? Makes it easy to compare the input and expected values at a glance, especially since the output is so small.

test-utils/eslint-plugin/use-vitest-local-expect.test.js
```diff
@@ -0,0 +1,66 @@
...
+  it('invalid case: no expect but context provided', async ({ expect }) => {
+    const { result } = await invalid({
+      code: "it('should do something', ({coo}) => { expect(true).toBe(true); });",
+      errors: ['correctExpect'],
+    });
+
+    expect(result.output).toMatchSnapshot();
+  });
+
+  it('invalid case: expect imported', async ({ expect }) => {
```
askoufis, line 62: Should add some tests for the import rename and object property rename cases.

askoufis (approved):
LGTM, left some comments about testing.

PR #1286, opened 2025-06-18
add new cjsInterop key to the sku config

DanDroryAu:
Add the `__UNSAFE_EXPERIMENTAL__cjsInteropDependencies` key to the `sku.config.ts`. These dependencies will be used internally by the cjsInterop plugin.

### DIFF
--- a/packages/sku/src/types/types.ts
+++ b/packages/sku/src/types/types.ts
@@ -137,6 +137,23 @@ export interface SkuConfig {
    * @default: []
    */
   __UNSAFE_EXPERIMENTAL__cjsInteropDependencies?: string[];
+
+  /**
+   * This function provides a way to modify sku's Vite configuration.
+   * It should only be used in exceptional circumstances where a solution cannot be achieved by adjusting standard configuration options.
+   *
+   * Before customizing your Vite configuration, please reach out in [#sku-support](https://seek.enterprise.slack.com/archives/CDL5VP5NU) to discuss your requirements and potential alternative solutions.
+   *
+   * As sku creates two Vite configs (`client` & `server|render`), this function will actually run twice.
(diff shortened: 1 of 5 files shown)


packages/sku/src/types/types.ts
askoufis, line 1: Maybe we should align `__unstableDangerouslySetViteConfig` to follow this naming pattern? e.g. `__UNSAFE_EXPERIMENTAL__dangerouslySetViteConfig`.

askoufis (commented):
Needs a changeset.

packages/sku/src/types/types.ts
DanDroryAu: I left it as a minor change since its `experimental` but I can bump it to a major if we have to do it since the API technically changed.

askoufis: I wouldn't bump to a major for the rename. The entire point of the experimental/unstable naming is that we provide no guarantees about its stability, even across patch versions.

DanDroryAu: Yeah the changeset is set to a `minor` version bump.

packages/sku/src/types/types.ts
```diff
@@ -128,6 +128,32 @@ export interface SkuConfig {
    * @default: "jest"
    */
   __UNSAFE_EXPERIMENTAL__testRunner?: 'vitest' | 'jest';
+  /**
+   * An array of cjs dependencies that vite should treat as ESM.
+   * This is an experimental option that may change or be removed without notice.
```
askoufis: This message and the one in the changeset could be more accurate. See [the docs](https://github.com/cyco130/vite-plugin-cjs-interop#:~:text=automatically%20unwrap%20the%20default%20export%20from%20cjs) for the plugin.

PR #1288, opened 2025-06-23
Fix tab opening

williamlark:
When starting a dev server a browser is opened to the dev url with the initial path included. When closing and restarting the dev server, the previously opened tab should be reused when available. This is an intended feature that apparently used to work but broke for some reason. _Edit_:  Looks like this broke in the [v14 release](https://github.com/seek-oss/sku/releases/tag/sku%4014.0.0).

Digging into it, I found some errors that we were silencing. `__dirname` wasn't available since sku is now `type: module`. Also, the applescript wasn't published with our application so it would always fall back to opening in a new tab.

Changes I made to fix this:
* Moved `openChrome.applescript` and `bin.js` into a top level bin directory that is included with our package (included in `package.json:files`
* Minor changes to the openBrowser function. Took a couple things from [Vites implementation](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/openBrowser.ts) and added debug logging.
* The Vite bundler now uses `server.open` option to open the browser so we don't have to rely on our own openBrowser code.

### DIFF
--- a/.changeset/smart-horses-mix.md
+++ b/.changeset/smart-horses-mix.md
@@ -2,4 +2,6 @@
 'sku': patch
 ---
 
-`start`: Browser tabs will be reused when starting a server
+`start`: Properly reuses Chrome tabs when starting a server.
+
+Fixes an issue of always opening a new tab in Chrome when one could be reused.
--- a/.changeset/smart-horses-mix.md
+++ b/.changeset/smart-horses-mix.md
@@ -1,0 +1,5 @@
+---
+'sku': patch
+---
+
+`start`: Browser tabs will be reused when starting a server
--- a/packages/codemod/package.json
+++ b/packages/codemod/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "bin": "./bin.js",
   "scripts": {
-    "build": "tsc --project tsconfig.build.json",
+    "build": "tspc --project tsconfig.build.json",
     "prepare": "ts-patch install -s"
   },
   "dependencies": {
--- a/packages/codemod/package.json
+++ b/packages/codemod/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "bin": "./bin.js",
   "scripts": {
-    "build": "tspc --project tsconfig.build.json",
+    "build": "tsc --project tsconfig.build.json",
     "prepare": "ts-patch install -s"
   },
   "dependencies": {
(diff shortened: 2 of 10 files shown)


packages/codemod/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "bin": "./bin.js",
   "scripts": {
-    "build": "tspc --project tsconfig.build.json",
+    "build": "tsc --project tsconfig.build.json",
```
williamlark: `tspc` works, but I don't know why. Is there a difference? I didn't look like `tspc` was a different package and is just an alias for `tsc`

packages/sku/bin/openChrome.applescript
```diff
@@ -1,6 +1,9 @@
 (*
-File copied from create-react-app
-https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/openChrome.applescript
```
askoufis, line 3: Would prefer this reference to the original code stick around.

packages/codemod/package.json
williamlark: Ah I see it's [part of ts-patch](https://github.com/nonara/ts-patch/blob/master/package.json#L56). Reverted this change.

packages/sku/src/services/vite/helpers/config/createConfig.ts
```diff
@@ -91,6 +91,7 @@ export const createViteDevConfig = (skuContext: SkuContext) =>
         allowedHosts: getAppHosts(skuContext).filter(
           (host) => typeof host === 'string',
         ),
+        open: skuContext.initialPath || true,
```
askoufis: This should respect `process.env.OPEN_TAB` so we maintain compatibility with the existing behaviour.

askoufis (approved):

.changeset/smart-horses-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+`start`: Browser tabs will be reused when starting a server
```
felixhabib:
```suggestion
`start`: Ensure browser tabs are reused when starting a server.

Fixes an issue that would prevent tabs from being reused.
```

Just trying to make it clearer that this is a fix and not a new feature. Happy for more changes

askoufis: Should we mention `Chrome` browsers here? Don't want to give the impression that this is possible in all browsers.

williamlark: Sure, happy to be clearer here. I've added that it fixes a bug and called out chrome specifically

felixhabib: *Chromium not Chrome. should be updated

PR #1291, opened 2025-06-24
change loadable default type

DanDroryAu:
The `default` property on the `loadable` factory type is now optional.

This change also adds the `typescript-transform-paths` dep to the two new packages to ensure typescript works correctly.


.changeset/slimy-bats-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sku-lib/vite': patch
+---
+
+`loadable`: The factory type now has an optional `default` property. Not all components loaded will have a `default` export.
```
williamlark: `nit`: just making it clearer that the default prop has changed and not added 

```suggestion
`loadable`: The factory type now sets `default` as optional. Not all components loaded will have a `default` export.
```

williamlark (approved):

PR #1298, opened 2025-07-03
enable supportedBrowsers when running vite

DanDroryAu:
Enable the `supportedBrowsers` from the `skuConfig` to work with `vite`.

### DIFF
--- a/packages/sku/src/services/vite/helpers/browserslist-to-esbuild.ts
+++ b/packages/sku/src/services/vite/helpers/browserslist-to-esbuild.ts
@@ -1,0 +1,98 @@
+import browserslist from 'browserslist';
+
+// convert the browserslist field in package.json to
+// esbuild compatible array of browsers
+export default function browserslistToEsbuild(
+  browserslistConfig: string[] | undefined = undefined,
+  options = {},
+) {
+  let internalBrowserslistConfig = browserslistConfig;
+  if (!internalBrowserslistConfig) {
+    // the path from where the script is run
+    const path = process.cwd();
+
+    // read config if none is passed
+    internalBrowserslistConfig = browserslist.loadConfig({ path, ...options });
+  }
+
+  const SUPPORTED_ESBUILD_TARGETS = [
+    'es',
+    'chrome',
+    'edge',
+    'firefox',
+    'ios',
+    'node',
+    'safari',
+    'opera',
+    'ie',
+  ];
+
+  // https://github.com/eBay/browserslist-config/issues/16#issuecomment-863870093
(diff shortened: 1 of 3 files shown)


packages/sku/src/services/vite/helpers/browserslist-to-esbuild.ts
```diff
@@ -0,0 +1,98 @@
+import browserslist from 'browserslist';
```
askoufis, line 2: Can we add a comment referencing the original implementation?


```suggestion
// Copied from https://github.com/marcofugaro/browserslist-to-esbuild/blob/0875f1a4ec71f206a634ba406c3528ca5ecf04de/src/index.js
import browserslist from 'browserslist';
```

packages/sku/src/services/vite/helpers/browserslist-to-esbuild.ts
askoufis, line 1: Would be great to copy-pasta [these unit tests](https://github.com/marcofugaro/browserslist-to-esbuild/blob/0875f1a4ec71f206a634ba406c3528ca5ecf04de/test/test.js) too just in case we ever decide to diverge from the original behaviour, and to give more confidence in the implementation.

Wouldn't bother testing the `package.json` or `browserslistrc` stuff, happy to just test the pure implementation: `array` in -> `string` out.

DanDroryAu: It is done!

DanDroryAu: Scrap that. Conflicts when pushing. Give me a moment I'll resolve it.

askoufis (commented):
Seems like some snapshot changes made their way in from the merge conflict? Don't think the braid or sku plugin snapshot should change in this PR.

tests/__snapshots__/braid-design-system.test.ts.snap
askoufis, line 1: I assume these are caused by seek's browser targets.

askoufis (approved):

tests/__snapshots__/braid-design-system.test.ts.snap
DanDroryAu: I think whenever the package changes it uses that to determine the hash so the hash changes. I've made small changes to some files in the sku program before and then the hash changes even though it does nothing to the output.

askoufis: I think the main contributors are braid and VE package versions, but yeah there's likely other contributors too.

PR #1302, opened 2025-07-09
`sku init`: Install specific version of `sku`

askoufis:
Running `sku init` from a snapshot version of `sku` would result in the latest version of `sku` installed, rather than the version of the snapshot. This is now fixed.

### DIFF
--- a/packages/sku/src/services/telemetry/index.ts
+++ b/packages/sku/src/services/telemetry/index.ts
@@ -3,7 +3,7 @@ import os from 'node:os';
 import { requireFromCwd } from '@/utils/cwd.js';
 import isCI from '@/utils/isCI.js';
 import provider from './provider.js';
-import skuPackageJson from '../../../package.json' with { type: 'json' };
+import skuPackageJson from 'sku/package.json' with { type: 'json' };
 import debug from 'debug';
 
 import type { SkuContext } from '@/context/createSkuContext.js';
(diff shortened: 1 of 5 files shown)


packages/sku/src/services/telemetry/index.ts
```diff
@@ -3,7 +3,7 @@ import os from 'node:os';
 import { requireFromCwd } from '@/utils/cwd.js';
 import isCI from '@/utils/isCI.js';
 import provider from './provider.js';
-import skuPackageJson from '../../../package.json' with { type: 'json' };
+import skuPackageJson from 'sku/package.json' with { type: 'json' };
```
askoufis, line 6: `sku/package.json` is a more stable reference to the `package.json`, so if this file moves it won't need to be updated.

tests/sku-init/sku-init.test.ts
```diff
@@ -3,6 +3,7 @@ import path from 'node:path';
 import fs from 'node:fs/promises';
 import { spawnSync } from 'node:child_process';
 import { runSkuScriptInDir } from '@sku-private/test-utils';
+import skuPackageJson from '../../packages/sku/package.json' with { type: 'json' };
```
askoufis, line 6: The `tests` package doesn't install `sku` as a dependency so I used a relative path here.

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@tsconfig/node-lts/tsconfig.json",
   "compilerOptions": {
+    "module": "node18",
```
askoufis, line 4: `@tsconfig/node-lts` sets `node16` here, but we need at least `node18` in order to use [import attributes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with) in the `tests` package.

williamlark (approved):
Nice one

PR #1303, opened 2025-07-09
Update PNPM to v10, migrate `.npmrc` settings to to `pnpm-workspace.yaml`

askoufis:
This will make testing pnpm v10-specific config management (in future PRs) easier.


.npmrc
```diff
@@ -1,10 +0,0 @@
-# Make sure the default patterns are still included (https://pnpm.io/npmrc#public-hoist-pattern)
-public-hoist-pattern[]="*eslint*"
-public-hoist-pattern[]="*prettier*"
```
askoufis, line 3: These are no longer defaults in PNPM v10, but we don't even need them anymore since we depend directly on `prettier` and `eslint` (and eslint flat config makes hoisting eslint plugins redundant too).

tests/package.json
```diff
@@ -29,6 +29,7 @@
     "@sku-fixtures/ssr-hello-world": "workspace:*",
     "@sku-fixtures/storybook-config": "workspace:*",
     "@sku-fixtures/styling": "workspace:*",
+    "@sku-fixtures/suspense": "workspace:*",
```
askoufis, line 32: Seems like maybe our `hoistWorkspacePackages: false` config wasn't working until now. This is necessary because a test file imports this package's sku config.

tests/sku-init/__snapshots__/sku-init.test.ts.snap
```diff
@@ -95,6 +95,13 @@ exports[`sku init > should create package.json 1`] = `
     "sku": "VERSION_IGNORED",
   },
   "name": "new-project",
+  "pnpm": {
+    "onlyBuiltDependencies": [
+      "sku",
+      "@swc/core",
+      "esbuild",
+    ],
+  },
```
williamlark, line 104: What made this snapshot change? I don't see `onlyBuiltDependencies` defined in any files changed here

williamlark (approved):

tests/sku-init/__snapshots__/sku-init.test.ts.snap
askoufis: We have [existing logic](https://github.com/seek-oss/sku/blob/8ad61eae1e8c1ce4740d6e67d0ce8bf0ca27b310/packages/sku/src/program/commands/init/init.action.ts#L138-L148) for adding `onlyBuiltDependencies` to `package.json`, but only in PNPM v10.

That logic will be updated to use `pnpm-workspace.yaml` in another PR.

tests/sku-init/sku-init.test.ts
```diff
@@ -61,13 +61,6 @@ describe('sku init', () => {
     );
     const packageJson = JSON.parse(contents);
 
-    // `workspace:` prefix comes from the default `saveWorkspaceProtocol` value.
-    // See https://pnpm.io/settings#saveworkspaceprotocol.
-    // In NPM, Yarn, or PNPM if configured in such a way, this will be just the version.
-    expect(packageJson.devDependencies.sku).toEqual(
-      `workspace:${skuPackageJson.version}`,
-    );
-
```
askoufis, line 70: Ended up deleting this test because it stopped working. We can make it work by setting [`saveWorkspaceProtocol: true`](https://pnpm.io/9.x/npmrc#save-workspace-protocol), but I don't want to do that because in all other cases the default makes sense.

Happy for this to not be tested as it's relatively inconsequential because it's only relevant when running `sku init` with a snapshot version.

PR #1305, opened 2025-07-11
Remove unused vite apis

williamlark:
While experimental support with Vite is available, some APIs are still under development. Static site generation is our primary focus for the Vite bundler at this time, so SSR commands with the experimental Vite bundler flags will now throw an error. 

Config functions have been slightly renamed as well to match their use, and the two SSR ones have been removed.

When we start focusing on SSR we can restore these files again.


pnpm-workspace.yaml
```diff
@@ -1,13 +1,24 @@
...
+  - vite-test-utils
+  - packages/*
 
 hoistWorkspacePackages: false
+
 linkWorkspacePackages: true
+
+onlyBuiltDependencies:
+  - '@swc/core'
+  - core-js-pure
```
askoufis: AFAIK this logs a message and that's it. Not worth allowing.
```suggestion
```

PR #1307, opened 2025-07-16
Support React 19

williamlark:
Updating dependancies and peer dep ranges to support React 19. React 18 support is still available so this is being released as a patch. No breaking changes are expected.

Some other changes:
* Long running tests now define their own extended timeout. This is so not every test has to wait a whole minute just to see errors. Most tests will error earlier, so having to wait the extra time to compensate for a couple longer-running tests is a little overkill.
* Added React 18 fixture. Doing so has meant we now have to inject our workspace packages so that React versions don't clobber each other. Once we decide to drop React 18 support we can remove the fixture and the workspace injection settings.
* Sanitising the `.pnpm` location in snapshots. Our Braid snapshot uses it and it ends up with a large url which we don't need.
* testing-library now uses the sku bin from within a fixture, rather than the workspace one. This allows injected packages to work correctly.
* Storybook tests now use `--exact-port` to catch when a port doesn't close properly on those tests
* Moved `jest-test` and `lint-format` tests to the node folder since they don't use the browser


williamlark:
Opening for review, but still having some issues with the braid snapshot. Hopefully my recent change fixes it 🤞

fixtures/react-18/sku.config.vite.ts
```diff
@@ -0,0 +1,8 @@
+import type { SkuConfig } from 'sku';
+
+import baseConfig from './sku.config.js';
```
askoufis:
```suggestion
import baseConfig from './sku.config.ts';
```

package.json
```diff
@@ -67,7 +67,7 @@
     "eslint-vitest-rule-tester": "^2.2.0",
     "globals": "^16.3.0",
     "husky": "^9.0.0",
-    "jiti": "^2.4.2",
+    "jiti": "2.4.2",
```
askoufis: Do newer versions of `jiti` break something, or was this unpinned accidentally?

tests/browser/__snapshots__/ssr-hello-world.test.ts.snap
```diff
@@ -45,6 +45,11 @@ exports[`ssr-hello-world > build > default port > should generate a production s
   </head>
   <body>
     <div id="app">
+      <link
+        rel="preload"
+        as="image"
+        href="http://localhost:4000/a85d2645cc4d80a20b5b.png"
+      >
```
askoufis, line 52: Any idea where these `link`s came from?

tests/browser/__snapshots__/public-path.test.ts.snap
```diff
@@ -67,6 +67,11 @@ exports[`public path > bundler: vite > build and serve > should build, serve, an
   </head>
   <body>
     <div id="app">
+      <link
+        rel="preload"
+        as="image"
+        href="/static/my-app/large-image-CufAniQw.avif"
+      >
```
askoufis, line 74: Ditto.

package.json
williamlark: Ah yes, it breaks linitng entirely. I found this issue when I was doing the vocab work yesterday and found it happen on here too. Pinning to 2.4.2 seems to be needed at the moment.

CleanShot 2025-08-05 at 16 04 56@2x

askoufis: Oh I saw that too in another branch, had no idea it was related to `jiti`. Good find!

Maybe we should add a PNPM `override` instead, and that we we can easily comment it because it would be in `pnpm-workspace.yaml` instead of `package.json`. It's also a dependency inside `sku` too I think. Thoughts?

williamlark: How about we trial workspace catalogs for it? 💭  ([commit](https://github.com/seek-oss/sku/pull/1307/commits/2b97f10bae68efd2f76fd96217b82a8f4cccba6e#diff-18ae0a0fab29a7db7aded913fd05f30a2c8f6c104fadae86c9d217091709794cR34-R37))

pnpm-workspace.yaml
```diff
@@ -31,3 +31,7 @@ publicHoistPattern:
 injectWorkspacePackages: true
 syncInjectedDepsAfterScripts:
   - build
+
+catalog:
+  # 05/08/25: there is currently a bug in later versions of jiti that break eslint ts support. Pinning to 2.4.2 fixes the issue.
+  jiti: 2.4.2
```
askoufis, line 37: Nice

tests/browser/__snapshots__/ssr-hello-world.test.ts.snap
williamlark: I'm really not sure, though I did start the fixtures locally and they are definitely in the dom. Why these only appeared in the snapshot now, I'm not sure.

askoufis: Maybe it's a React 19 thing? They did ship [some preloading APIs](https://react.dev/reference/react-dom#resource-preloading-apis), but I wouldn't expect them to be called by default, nor would I expect the resulting `link` tags to end up in the `body` instead of the `head`. 🤔

askoufis: Honestly stumped by this. `renderToString(</App)` contains the `link` element, so _something_ is injecting it. It might be related to loadable components, but not sure. Overall it's not affecting hydration, and it's not really _wrong_, just non-standard and redundant in the vite implementation.

PR #1308, opened 2025-07-18
Adding cli-testing-library and reorganising files

williamlark:
This change is part of a larger set of changes to improve the CLI testing experience. This PR includes:

**Adding cli-testing-library to the repo and testing the `sku-test` fixture with it.**
The move to cli-teting-library will allow up to test the terminal output of the sku cli easily, and may help remove the need for some helper functions. There are no performance improvements, but should help development experience. 

This may have some changes as the work continues, but I feel like it's in a good starting spot.

**Reorganising the test-utils folder into seperate items.**  
As part of adding the testing library, I've reorganised the files in the test-utils workspace. The test-utils folder still exists while more work on the cli-testing-library implementation is being done, but will be cleaned up in a follow up PR.

I've also changed a few imports to reference the workspace packages, rather than diving into the filesystem manually.


private/eslint-plugin/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@sku-private/eslint-plugin",
+  "type": "module",
+  "main": "index.cjs",
+  "skuSkipValidatePeerDeps": true
```
williamlark: Do we need `skuSkipValidatePeerDeps` in the workspace packages? I wasn't sure and didn't add it to all of the new packages.

askoufis: I _think_ it should only be necessary inside workspace packages that install `sku` as a dep. So it shouldn't be in this one, but may be necessary in others.

We don't actually do peer dep validation for PNPM anymore as the existing implementation simply isn't compatible with PNPM's node modules, so it's disabled. That's a future thing to fix.

williamlark: That would make sense. I've removed it from a few of the new packages now

PR #1309, opened 2025-07-18
Replace `eslint-plugin-unicorn` with `eslint-plugin-n`

askoufis:
Some small eslint config and dep changes:
- Removed some seemingly unused ignore paths
- Use multi-segment globs
- Replace `eslint-plugin-unicorn` with `eslint-plugin-n`. They both have a "prefer node protocol" rule, but `eslint-plugin-n` has a smaller dep tree.
- Update various eslint-related deps


eslint.config.cjs
```diff
@@ -134,7 +132,7 @@ module.exports = [
     },
   },
   {
-    files: ['tests/**', 'test-utils/**'],
+    files: ['tests/**', 'test-utils/**', 'vite-test-utils/**'],
```
williamlark: Can remove `vite-test-utils`

askoufis: Oh woops, think this was a bad rebase.

PR #1310, opened 2025-07-18
Mark private packages as private

askoufis:
Private packages should be marked as such otherwise [changesets will attempt to publish them](https://github.com/seek-oss/sku/actions/runs/16360258785/job/46226719958#step:7:35). Luckily we own the `@sku-private` scope, and the token in the repo doesn't have permissions to publish to that scope, so nothing actually published.


williamlark (approved):
Phew, thank you. Totally missed this!

PR #1311, opened 2025-07-18
`init, configure`: Create/suggest recommended PNPM configuration

askoufis:
`sku init` will now run `pnpm add --config pnpm-plugin-sku` if PNPM >=10 is detected.

`sku configure` will log warnings about
- `.npmrc` files, with a suggesting to migrate to `pnpm-workspace.yaml`
- Missing `pnpm-plugin-sku` config dep
- A PNPM version < 10.13.0, which is the minimum version that supports automatically loading PNPM plugins.

These are hopefully slightly annoying-enough warnings to get people to action them, while also not being _too_ intrusive.

As of writing, the logs look like this:
image

I went with the `blue.bold` style for important files/commands, as that's what we use elsewhere in the CLI (though we're not exactly consistent about this), and I went with regular `bold` style for useful info, in this case a URL.


private/testing-library/index.ts
```diff
@@ -24,7 +25,7 @@ type SkuCommand =
 
 // Default configuration for cli-testing-library
 const DEFAULT_CONFIG: Partial<Config> = {
-  asyncUtilTimeout: 5000,
+  asyncUtilTimeout: 10000,
```
askoufis, line 28: 5000ms wasn't quite cutting it. I was often getting test failures when running all `sku-test` tests, but no failures when running tests individually, so I bumped up this timeout.

tests/sku-test.test.ts
```diff
@@ -1,11 +1,15 @@
...
+import { scopeToFixture, waitFor, cleanup } from '@sku-private/testing-library';
 
 const { render } = scopeToFixture('sku-test');
 
 const testRunners = ['vitest', 'jest'] as const;
 
+afterEach(async () => {
+  await cleanup();
+});
+
```
askoufis, line 12: I'm not sure we're getting this for free with `vitest` because we're not injecting globals, so I added a [cleanup](https://cli-testing.com/guides/api/#cleanup) call after each test.

vitest.config.ts
```diff
@@ -19,6 +19,7 @@ export default defineConfig({
     hookTimeout: TEST_TIMEOUT + 1000,
     testTimeout: TEST_TIMEOUT + 1000,
     exclude: [...defaultExclude, '**/fixtures/**'],
+    restoreMocks: true,
```
askoufis, line 22: This just seemed like a sane default, I don't know why it's not enabled by default in vitest.

tests/configure.test.ts
```diff
@@ -48,6 +48,7 @@ const removeAppDir = async (folder: string) =>
 describe('configure', () => {
   describe('default', () => {
     beforeAll(async () => {
+      await removeAppDir(appFolder);
```
askoufis, line 51: Depending on when you interrupt these tests, you _may_ end up with a dirty state the next time you run the test. This ensures the test directory is always clean.

packages/sku/src/services/packageManager/getPnpmConfigValue.ts
```diff
@@ -0,0 +1,20 @@
+import { execAsync } from '@/utils/execAsync.js';
+
+type PnpmConfigKey = 'public-hoist-pattern' | 'only-built-dependencies';
+
+export const getPnpmConfigValue = async (
+  configKey: PnpmConfigKey,
+): Promise<string[] | null> => {
+  const rawResult = await execAsync(`pnpm config get ${configKey}`);
```
askoufis: There's a little bit of overhead to running this twice during every `configure` invocation, but in my experience it's relatively minimal.

packages/sku/src/services/packageManager/pnpmConfig.ts
```diff
@@ -0,0 +1,156 @@
...
+export const skuPnpmConfig = [
+  'publicHoistPattern:',
+  ...skuPublicHoistPatterns.map((dep) => `  - ${dep}`),
+  'onlyBuiltDependencies:',
+  ...skuOnlyBuiltDependencies.map((dep) => `  - ${dep}`),
+].join('\n');
+
+export type PnpmConfigAction = 'create' | 'update' | 'log' | 'noop';
+
+export const createOrUpdatePnpmConfig = async ({
```
williamlark: Could this be seperate functions? `createPnpmCOnfig`, `updatePnpmConfig`, etc. (you wouldn't need the noop one). I feel like this function is trying to do to much

packages/sku/src/services/packageManager/pnpmConfig.ts
```diff
@@ -0,0 +1,156 @@
...
+  pnpmWorkspaceFileExists: boolean;
+}): Promise<PnpmConfigAction> => {
+  if (!pnpmWorkspaceFileExists && !npmrcExists) {
+    return 'create';
+  }
+
+  const publicHoistPatterns =
+    (await getPnpmConfigValue('public-hoist-pattern')) ?? [];
+  const onlyBuiltDependencies =
+    (await getPnpmConfigValue('only-built-dependencies')) ?? [];
```
williamlark: suggestion: if we're going to need both, could we call them in a `Promise.all`? It's minor, but if we're calling this function twice it could be nice to do it simultaneously.

packages/sku/src/services/packageManager/pnpmConfig.ts
```diff
@@ -0,0 +1,156 @@
...
+
+export const detectPnpmConfigFiles = async (rootDir: string) => {
+  const npmrcExists = await exists(path.join(rootDir, '.npmrc'));
+
+  if (npmrcExists) {
+    console.log(
+      chalk.yellow(
+        "Found '.npmrc' file. It is strongly recommended to migrate all .npmrc configuration to 'pnpm-workspace.yaml' and gitignore '.npmrc'. See https://pnpm.io/settings for more information.",
+      ),
+    );
```
williamlark: suggestion: What do you think about taking our changelogs approach to sentences and having these as separately logged lines?


```suggestion
    console.log(chalk.yellow("Found '.npmrc' file."));
    console.log(chalk.yellow("It is strongly recommended to migrate all .npmrc configurations to 'pnpm-workspace.yaml' and gitignore '.npmrc'. See https://pnpm.io/settings for more information."));
```

packages/sku/src/services/packageManager/pnpmConfig.ts
```diff
@@ -0,0 +1,155 @@
...
+    if (pnpmWorkspaceFileExists) {
+      return 'update';
+    }
+
+    return 'create';
+  }
+
+  // In all other cases, simply log a message instructing the user to update the config manually
+  return 'log';
+};
```
williamlark: This is smart, but I think it would be simpler to just call the functions you need straight in here.

```ts
if (publicHoistPatterns.length === 0 && onlyBuiltDependencies.length === 0) {
    if (pnpmWorkspaceFileExists) {
      return updatePnpmConfig(...);
    }

    return createPnpmConfig(...);
  }
```

packages/sku/src/services/packageManager/pnpmConfig.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  // There is potentially conflicting config somewhere, so log a message instructing
+  // the user to update the config manually
+  if (action === 'log') {
+    console.log(
+      chalk.yellow(
+        'Potentially conflicting PNPM config detected. Please update your pnpm-workspace.yaml manually with the following config:',
+      ),
+    );
```
williamlark: suggestion: Same here, could be two different lines.

tests/sku-init/__snapshots__/sku-init.test.ts.snap
```diff
@@ -112,6 +112,8 @@ exports[`sku init > should create pnpm-workspace.yaml 1`] = `
 "publicHoistPattern:
   - eslint
   - prettier
+onlyBuiltDependencies:
+  - sku
```
williamlark: I've noticed that pnpm will warn about `core-js-pure` whenever I install the sku workspace dependancies. Only by adding `ignoredBuiltDependencies: core-js-pure` to the workspace will it stop. Are there any packages we might need to add to `ignoredBuiltDependencies` for consumers so they don't see warnings on first install?

williamlark (commented):
Nice work. Some minor comments

packages/sku/src/services/packageManager/pnpmConfig.ts
williamlark: Looks like you beat me to it haha

tests/sku-init/__snapshots__/sku-init.test.ts.snap
askoufis: I think configuring `onlyBuiltDependencies` implicitly disallows any non-listed dependencies to be built, so I'm confused why you're seeing warnings in the sku repo about this.

I'm also not quite sure about the difference between [`ignoredBuiltDependencies`](https://pnpm.io/settings#ignoredbuiltdependencies) and [`neverBuiltDependencies`](https://pnpm.io/settings#neverbuiltdependencies). They seem the same to me.

askoufis:
Updated CLI output:
image

packages/sku/src/services/packageManager/pnpmConfig.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    [false, false, true],
+    [false, true, false],
+    [false, true, true],
+    [true, false, false],
+    [true, false, true],
+    [true, true, false],
+    [true, true, true],
+  ];
+
+  describe.for(testCases)(
```
williamlark: nit: maybe a personal preference, but I find the template string table a bit easier to scan and read. Would be nice, but not strongly fussed either way. 

```suggestion
  describe.for`
  npmrcExists   |  hasRecommendedPnpmVersionInstalled   | pnpmPluginSkuInstalled
  ${false}      | ${false}                              | ${false}
  ...
`(
```

packages/sku/src/services/packageManager/getPnpmConfigDependencies.ts
```diff
@@ -0,0 +1,17 @@
+import { execAsync } from '@/utils/execAsync.js';
+
+export const getPnpmConfigDependencies = async (): Promise<string[]> => {
+  const rawResult = await execAsync(`pnpm config get --json`);
```
williamlark, line 7: Why not fetch the `config-dependencies` explicitly?

askoufis:
(Hopefully) final log output:

image

pnpm-workspace.yaml
```diff
@@ -5,6 +5,9 @@ packages:
   - packages/*
   - private/*
 
+configDependencies:
```
askoufis, line 8: `pnpm-plugin-sku` doesn't really affect the sku repo at all, and it means internal testing/usage of `sku` won't see the warnings added in this PR.

williamlark (approved):

packages/sku/src/services/packageManager/getPnpmConfigDependencies.ts
askoufis: Oh yeah, for whatever reason it returns `[object Object]` whenever I try. Seems like a PNPM bug honestly. I'll add a comment here, and raise a PNPM issue too.

askoufis: https://github.com/pnpm/pnpm/issues/9797

packages/sku/src/services/packageManager/pnpmConfig.test.ts
askoufis: I use both representations depending on the test. I agree that the template string table would be much more readable for this one.

PR #1313, opened 2025-07-21
`vite build`: Generate source maps when sourceMapsProd: true is configured

askoufis:
Converted the `source-maps` fixture to ESM and TypeScript and added a `vite build` test to the `source-maps` tests.

The feature itself simply involved setting [`build.sourcemaps`](https://vite.dev/config/build-options.html#build-sourcemap).


fixtures/source-maps/sku.config.ts
```diff
@@ -0,0 +1,12 @@
+import { makeStableHashes } from '@sku-private/test-utils';
+import type { SkuConfig } from 'sku';
+
+export default {
+  clientEntry: 'src/client.tsx',
+  renderEntry: 'src/render.tsx',
+  publicPath: '/static/source-maps',
+  port: 8303,
+  target: 'dist',
+  sourceMapsProd: process.env.SKU_SOURCE_MAPS_PROD === 'true',
```
askoufis, line 10: Decided to make this configurable via an envar to avoid having to define 4 different sku configs for the `source-maps` tests: 1 for each combination of `vite/webpack` and `sourceMapsProd: true/false`.

fixtures/source-maps/src/App.tsx
askoufis, line 1: Various unrelated test snapshots that run `sku build` with Vite have been updated as they now generate sourcemaps (because `sourceMapsProd` defaults to `true`).

williamlark (approved):

PR #1314, opened 2025-07-21
Add `config.target` support for build output directories via Vite 

adenj:
Enable Vite bundler to respect `config.target` for build output directories and replace hard-coded `'dist'` paths throughout Vite implementation


.changeset/public-onions-post.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': minor
+---
+
+Vite bundler now respects the `target` config property for specifying custom build output driectories
```
williamlark: nit

```suggestion
`vite (build)`: bundler now respects the `target` config property for specifying custom build output directories
```

williamlark (commented):
Amazing 💯  Is there a fixture that tests this already? Would be great to enable it for vite if there is, or add one if there isn't any at all.

packages/sku/src/services/vite/index.ts
```diff
@@ -7,20 +7,26 @@ import {
...
     if (skuContext.routes) {
       await prerenderConcurrently(skuContext);
     }
-    await cleanTargetDirectory(`${process.cwd()}/dist/render`, true);
-    await cleanTargetDirectory(`${process.cwd()}/dist/.vite`, true);
+    await cleanTargetDirectory(outDir.ssg, true);
+    await cleanTargetDirectory(
+      path.join(skuContext.paths.target, '.vite'),
+      true,
+    );
```
williamlark: suggestion: could be nice to encapsulate this logic with the other `createOutDir` stuff. Maybe there can be a little function exported on that object that does the joining for us.


```suggestion
    await cleanTargetDirectory(outDir.join('.vite'), true);
```

fixtures/custom-output-target/src/App.tsx
```diff
@@ -0,0 +1 @@
+export default () => <p>Hello there!</p>;
```
williamlark, line 1: 👋

williamlark (commented):
Just realised, theres a fixture but no test for it? 🧪

williamlark (approved):

PR #1315, opened 2025-07-21
`sku test`: Don't configure twice

askoufis:
`configure` is already run inside [the `test` action](https://github.com/seek-oss/sku/blob/1ea7bdc72eadabe4c7faf77f39c51b57d4fcd3f0/packages/sku/src/program/commands/test/test.action.ts#L15), so this second call inside `runJestTests` is redundant.

### DIFF
--- a/packages/sku/src/program/commands/test/jest-test-handler.ts
+++ b/packages/sku/src/program/commands/test/jest-test-handler.ts
@@ -3,7 +3,6 @@ import { run } from 'jest';
 
 import isCI from '@/utils/isCI.js';
 import { runVocabCompile } from '@/services/vocab/runVocab.js';
-import { configureProject } from '@/utils/configure.js';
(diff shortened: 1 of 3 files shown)


packages/sku/src/program/commands/test/jest-test-handler.ts
```diff
@@ -16,7 +15,6 @@ export const runJestTests = async (
   },
   { args = [] }: { args: string[] },
 ) => {
-  await configureProject(skuContext);
   await runVocabCompile(skuContext);
```
williamlark: Looks like the vocab compile is running twice too. Could remove this as well

askoufis: Oh good catch! Fixed in https://github.com/seek-oss/sku/pull/1315/commits/9bac79df6e806db52f6414f3022d49791b289062.

PR #1317, opened 2025-07-24
Cli testing refactor

williamlark:
This is quite a large PR focus primarily around refactoring our testing setup in sku. The goal was to introduce `cli-testing-library` to the system to make testing CLI output and commands easier. This library makes reading CLI output simple, and is is has a familiar API as the rest of the `testing-library` ecosystem. 

A few other changes were introduced too, helping cut bout 20s off of local test runs (benchmarked from my own machine) and about 10s from the CLI. 

## Full list of changes

* Moved the `sku-init` changes to their own fixtures folder. I've also removed the restoring of the lock file in the cleanup since running the test doesn't make any changes to the global lock file anymore.
* Rearranged the tests into `browser/` and `node/` folders. This allows us to make the definition explicit in the file system rather than declaring this via variables in the vitest config. 
* Removed the `startAssetServer` command in favour of `http-server` within each fixture. This allows each fixture to be in charge of how an asset server is started, and means less code for us to maintain.
* Adding some common output to both the Webpack and Vite commands (e.g., `Starting development server...` when running `sku start`). This allows us to have some consistent wording in the terminal that we can check for in our tests.
* Built out the rest of the `testing-library` private package and implemented it in every test in the codebase. This is where the majority of the file changes come from.
* ignoring fixtures from the vitest watch mode. Some fixtures that generate folders make the watch mode enter a continuous loop, so I've ignored them all. We may want to only ignore specific ones, but I didn't have any issues with them all ignored while I was making changes.


vitest.config.ts
```diff
@@ -4,15 +4,13 @@ import tsconfigPaths from 'vite-tsconfig-paths';
...
-  'tests/configure.test.ts',
-];
-
 export default defineConfig({
   plugins: [tsconfigPaths()],
+  server: {
+    watch: {
+      ignored: ['**/fixtures/**'],
+    },
+  },
```
askoufis, line 13: I was wondering why vitest watch mode was looping forever. Glad it's fixed!

private/testing-library/render-cli.ts
```diff
@@ -0,0 +1,77 @@
...
+    exec: (
+      command: string,
+      args: string[] = [],
+      options: Partial<Omit<RenderOptions, 'cwd'>> = {},
+    ) =>
+      render(command, args, {
+        ...options,
+        cwd: appDir,
+      }),
+    joinPath: (...paths: string[]) => path.join(appDir, ...paths),
```
askoufis: Thoughts on `fixturePath` instead of `joinPath` here? Feel like it's a bit more descriptive of what the API is doing.
```suggestion
    fixturePath: (...paths: string[]) => path.join(appDir, ...paths),
```

private/testing-library/render-cli.ts
```diff
@@ -0,0 +1,77 @@
...
+  options: Partial<RenderOptions> = {},
+) => render(skuBin, [command, ...args], options);
+
+export const scopeToFixture = (fixtureFolder: string) => {
+  const appDir = path.dirname(
+    require.resolve(`../../fixtures/${fixtureFolder}/package.json`),
+  );
+
+  return {
+    render: (
```
askoufis: Could we call this `sku`/`skuCli`/`renderSku` or something? `render` is very generic here and gives no indication as to what's actually being "rendered".

tests/browser/sku-webpack-plugin.test.ts
```diff
@@ -0,0 +1,62 @@
...
+        {
+          spawnOpts: {
+            env: {
+              ...process.env,
+              NODE_ENV: 'production',
+            },
+          },
+        },
+      );
+      await waitFor(async () => {
```
askoufis:
```suggestion
      await waitFor(() => {
```

tests/browser/styling.test.ts
```diff
@@ -88,30 +69,27 @@ describe('styling', () => {
...
+      const storybook = await exec('pnpm', [
+        'storybook',
+        'dev',
+        '--ci',
+        '--port',
+        storybookPort.toString(),
+      ]);
+      globalExpect(
+        await storybook.findByText(
+          'Storybook 9.0.16 for react-webpack5 started',
```
askoufis: This text will probably change in a lockfile update when the `storybook` dep updates. Worth using a looser matcher if possible.

packages/sku/src/program/commands/start/webpack-start-handler.ts
```diff
@@ -215,10 +215,8 @@ export const webpackStartHandler = async ({
       appHosts?.[0]
     }:${availablePort}${initialPath}`;
 
-    console.log();
-    console.log(
-      chalk.blue(`Starting the development server on ${chalk.underline(url)}`),
-    );
+    console.log('Starting development server...');
+    console.log(chalk.blue(`Local: ${chalk.underline(url)}`));
```
askoufis, line 203: Something like `Development server URL` or similar would be more descriptive than `Local` here:
```suggestion
    console.log(chalk.blue(`Development server URL: ${chalk.underline(url)}`));
```

tests/browser/storybook-config.test.ts
```diff
@@ -94,71 +88,80 @@ describe('storybook-config', () => {
...
+          exitCode: 0,
+        });
       });
 
-      closeStorybookServer = await startAssetServer(
-        assetServerPort,
-        storybookDistDir,
-      );
-      await waitForUrls(storyIframeUrl);
+      const assetServer = await exec('npm', ['run', 'start:asset-server']);
```
askoufis:
```suggestion
      const assetServer = await exec('pnpm', ['run', 'start:asset-server']);
```

tests/browser/storybook-config.test.ts
```diff
@@ -196,8 +199,13 @@ describe('storybook-config', () => {
     it('should render a ".mdx" file', async ({ expect }) => {
       const docsIframePath = '/iframe.html?viewMode=docs&id=docstest--docs';
       const docsIframeUrl = `${assetServerUrl}${docsIframePath}`;
+
+      const assetServer = await exec('npm', ['run', 'start:asset-server']);
```
askoufis:
```suggestion
      const assetServer = await exec('pnpm', ['run', 'start:asset-server']);
```

tests/browser/sku-webpack-plugin.test.ts
```diff
@@ -0,0 +1,62 @@
...
+        globalExpect(build.hasExit()).toMatchObject({
+          exitCode: 0,
+        });
+      });
+    });
+
+    it('should create valid app', async ({ expect }) => {
+      const assetServer = await exec('npm', ['run', 'start:asset-server']);
```
askoufis:
```suggestion
      const assetServer = await exec('pnpm', ['run', 'start:asset-server']);
```

tests/browser/sku-with-https.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      });
+
+      it('should support the supplied middleware', async ({ expect, task }) => {
+        skipCleanup(task.id);
+        const snapshot = await getAppSnapshot({
+          url: `${url}/test-middleware`,
+          expect,
+        });
+        expect(snapshot).toMatchSnapshot();
+      });
```
askoufis: If these two tests get merged into one, would it remove the need for `skipCleanup`?

tests/browser/__snapshots__/lint-format.test.ts.snap
```diff
@@ -50,30 +50,3 @@ const foo = [
 ];
 
 `;
-
-exports[`lint-format > sku lint > lint errors are reported: typescriptFile.ts 1`] = `
```
askoufis, line 54: I'd prefer to keep these snapshots if possible. Although we're technically testing the outputs of external CLIs, it at least indicates that they're running correctly. It's more insurance than anything else - we don't actually care that much about the output.

tests/browser/ssr-hello-world.test.ts
```diff
@@ -0,0 +1,124 @@
...
+      ).toBeInTheConsole();
+    });
+
+    describe('default port', () => {
+      it('should generate a production server based on config', async ({
+        expect,
+      }) => {
+        await node(['dist-build/server.cjs']);
+
+        const assetServer = await exec('npm', ['run', 'start:asset-server']);
```
askoufis:
```suggestion
        const assetServer = await exec('pnpm', ['run', 'start:asset-server']);
```

fixtures/sku-init/README.md
```diff
@@ -0,0 +1 @@
+The tests run `sku init` in this directory so this fixture intentionally relatively empty.
```
askoufis:
```suggestion
The tests run `sku init` in this directory so this fixture is intentionally relatively empty.
```

tests/browser/__snapshots__/ssr-hello-world.test.ts.snap
```diff
@@ -1,102 +1,5 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`ssr-hello-world > build > custom port > should generate a production server running on custom port 1`] = `
```
askoufis, line 3: This tests a feature of [the server entrypoint](https://github.com/seek-oss/sku/blob/251d4548fd89a0b1d80f1bbc9e2c7e1cc9c7e3c9/packages/sku/src/services/webpack/entry/server/index.ts#L7) itself, separate to the CLI, so IMO worth keeping.

askoufis (commented):
Huge effort! Mainly some small nitpicks, but otherwise this is awesome. Well done!

private/testing-library/render-cli.ts
williamlark: I like `sku` 👍  I'll change that up

tests/browser/styling.test.ts
williamlark: I've used a regex instead. `/Storybook \d+\.\d+\.\d+ for react-webpack5 started/`

packages/sku/src/program/commands/start/webpack-start-handler.ts
williamlark: I did this to copy the Vite output for some consistency. I don't think it's too confusing with `Local:` tbh

tests/browser/sku-with-https.test.ts
williamlark: Yea, it would. I considered doing so since I'd prefer not to cleanup, but I also like the seperate tests. I'll merge them since you picked up on it too 👌

Edit: Looks like I can hint the snapshots, so we can still tell them apart in the snapshot title

askoufis: Oh that's neat, didn't know about snapshot hinting.

packages/sku/src/program/commands/start/webpack-start-handler.ts
askoufis: Ah ok. In that case I disagree with Vite, but I guess the consistency is probably worth it.

tests/browser/__snapshots__/lint-format.test.ts.snap
williamlark: I removed them since I'm actually testing this output in the test itself. e.g., `findByText('src/typescriptFile.ts(1,7): error')`. Since we're only testing the CLI output I figured it can just be done in the test rather than a snapshot.

Though, I'm not really fussed. These snapshots do look nice so if you prefer them I can add them back in

tests/node/sku-init.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+const { sku, fixturePath } = scopeToFixture('sku-init');
+
+const projectName = 'new-project';
+const projectDirectory = fixturePath(projectName);
+
+describe('sku init', () => {
+  beforeAll(async () => {
+    await fs.rm(projectDirectory, { recursive: true, force: true });
+    await fs.writeFile(fixturePath('pnpm-workspace.yaml'), '');
```
williamlark, line 20: Creating a pnpm-workspace in this fixture means the dependancies will be installed scoped to this folder, and will generate a seperate pnpm-lock file. This is great because it means the install is totally isolated. 

We can also check the contents of this workspace file to make sure the sku pnpm config is being installed correctly; I've added an additional test with a snapshot down below (`should update the pnpm-workspace.yaml`). 

@askoufis

askoufis: Nice! Great that this is simpler and can run in parallel with the other tests now.

askoufis (commented):
I'm happy with the current changes, though the new `polyfill` tests will need a refactor to align with the new testing APIs.

tests/browser/__snapshots__/lint-format.test.ts.snap
askoufis: Oh I missed the assertions on the lint errors. Happy with that instead of the snapshots 👍

askoufis (approved):

PR #1318, opened 2025-07-24
Add fixture for testing `polyfills` behaviour

askoufis:
Adds a `polyfills` fixture that configures a local `polyfills.ts` and a 3rd-party dependency that both inject a global function. This global function is then called post-hydration, which is picked up by the `appSnapshot` diff.

> [!NOTE]
> Vite is not being tested yet for this fixture.

### DIFF
--- a/tests/__snapshots__/polyfills.test.ts.snap
+++ b/tests/__snapshots__/polyfills.test.ts.snap
@@ -1,0 +1,177 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`polyfills > bundler webpack > build > should create valid app 1`] = `
+"<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>
+      polyfill
+    </title>
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1"
+    >
+    <link
+      data-chunk="main"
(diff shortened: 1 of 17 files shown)


jahredhope (approved):
Great fixture. Love how clear the post hydrate diff in the snapshot is.

tests/__snapshots__/polyfills.test.ts.snap
```diff
@@ -0,0 +1,177 @@
...
+@@ -64,5 +64,11 @@
+       src="/main.js"
+     >
+     </script>
++    <p>
++      This node was injected by the 'injected' global function.
++    </p>
++    <p>
++      This node was injected by the '3rd-party-polyfill' dependency.
++    </p>
```
williamlark, line 173: suggestion: Is it possible for us to check that these are in the document with puppeteer rather than using a snapshot? Makes the test a bit more explicit on what we're testing here (most of the output isn't isn't needed)

williamlark (approved):

tests/__snapshots__/polyfills.test.ts.snap
askoufis: As discussed: we definitely could narrow down the output of some of our tests - they're overly verbose in some cases. However it's probably worth doing this holistically across all tests rather than in isolation.

PR #1319, opened 2025-07-25
tests: adding vitest eslint plugin

williamlark:
The Jest rules don't seem to be catching lint errors in our code, and since we are using vitest it makes sense we use that plugin instead.

### Changes

* Added the vitest eslint plugin to our eslint config
* changed the eslint file from `cjs` to `ts` (no reason for it to be a cjs file)
* Removed the jest plugin and rules from the seek eslint config. This may not be needed but we don't want these to override the vite stuff, and they don't need to be there.


package.json
```diff
@@ -65,6 +66,7 @@
     "eslint-vitest-rule-tester": "^2.2.0",
     "globals": "^16.3.0",
     "husky": "^9.0.0",
+    "jiti": "^2.4.2",
```
williamlark, line 69: jiti is needed for eslint ts file support

.d.ts
askoufis, line 1: Can this file have a proper name? Honestly didn't know it was possible to just have a bare `.d.ts` file but it's weirding me out 😅.

askoufis (approved):

PR #1321, opened 2025-07-25
`vite`: Add `polyfills` support

askoufis:
Virtual modules seemed like the easiest way to inject polyfills into the client entry. Rollup does support multiple entries [just like webpack](https://github.com/seek-oss/sku/blob/a51ee0f635213e0b23777e34e5903892a5d23359/packages/sku/src/services/webpack/config/webpack.config.ts#L72), however to ensure import order is maintained you still need to actual do the side-effect imports in the client entry. See [this FAQ section](https://rollupjs.org/faqs/#how-do-i-add-polyfills-to-a-rollup-bundle) for more details.

Since we want to mimic the existing `polyfills` API, I opted instead to create a plugin that resolves the `virtual:sku/polyfills` virtual module to a module containing side-effect imports for all the resolved polyfills. This seems to work well. The result in the output bundle is that the imported modules get inlined at the top of the main bundle, before everything else.

> [!NOTE]
> `polyfills.test.ts` mainly contains indentation changes. The only actual change was adding `vite` to the list of bundlers.

### DIFF
--- a/packages/sku/src/services/vite/plugins/polyfillsPlugin.ts
+++ b/packages/sku/src/services/vite/plugins/polyfillsPlugin.ts
@@ -1,0 +1,26 @@
+import type { Plugin } from 'vite';
+
+import type { SkuContext } from '@/context/createSkuContext.js';
+import { resolvePolyfills } from '@/utils/resolvePolyfills.js';
+
+export const polyfillsPlugin = (skuContext: SkuContext): Plugin => {
+  const virtualModuleId = 'virtual:sku/polyfills';
+  const resolvedVirtualModuleId = `\0${virtualModuleId}`;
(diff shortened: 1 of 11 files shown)


packages/sku/src/services/vite/plugins/polyfillsPlugin.ts
```diff
@@ -0,0 +1,26 @@
+import type { Plugin } from 'vite';
+
+import type { SkuContext } from '@/context/createSkuContext.js';
+import { resolvePolyfills } from '@/utils/resolvePolyfills.js';
+
+export const polyfillsPlugin = (skuContext: SkuContext): Plugin => {
```
askoufis, line 6: Stuck to the virtual modules conventions outlined [here](https://vite.dev/guide/api-plugin.html#virtual-modules-convention).

PR #1322, opened 2025-07-28
`init`: Populate `packageManager` field in `package.json` if the new app is at the repo root

askoufis:
The `packageManager` field in `package.json` will now be populated with the appropriate config based on the package manager used to execute `sku init`.

The implementation only includes `packageManager` if `sku init` wasn't run inside a monorepo.

This was tested with a snapshot, both for a new repo and in a monorepo. Our current `sku-init` test was not affected because the test is performed inside the `sku` monorepo. I believe the changes in #1317 will result in the `packageManager` field being visible in the `package.json` snapshot test in the `sku-init` suite.

### DIFF
--- a/packages/sku/src/program/commands/init/init.action.ts
+++ b/packages/sku/src/program/commands/init/init.action.ts
@@ -16,6 +16,8 @@ import {
   getPackageManagerInstallPage,
   getInstallCommand,
   isAtLeastPnpmV10,
+  rootDir,
+  packageManagerVersion,
 } from '@/services/packageManager/packageManager.js';
 import { write as prettierWrite } from '@/services/prettier/runPrettier.js';
 import { fix as esLintFix } from '@/services/eslint/runESLint.js';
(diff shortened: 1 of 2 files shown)


packages/sku/src/program/commands/init/init.action.ts
```diff
@@ -131,10 +133,18 @@ export const initAction = async (
     },
   };
 
+  const isRepoRoot = rootDir === null || rootDir === initDir;
```
askoufis, line 136: `rootDir` will be `null` inside a new repo.

`rootDir` will _not_ equal `initDir` in basically any monorepo situation. I can't think of a situation where `rootDir === initDir`, but this check is here just in case that ever does happen.

packages/sku/src/program/commands/init/init.action.ts
```diff
@@ -64,12 +66,12 @@ export const initAction = async (
   projectName: string,
   { verbose, skuContext }: { verbose: boolean; skuContext: SkuContext },
 ) => {
-  const root = path.resolve(projectName);
-  setCwd(root);
+  const initDir = path.resolve(projectName);
```
askoufis, line 69: I renamed this to help distinguish it from `rootDir`, which is the repo root detected by our package manager utils.

williamlark (approved):

PR #1324, opened 2025-07-29
`start`: Align webpack and vite `start` implementations

askoufis:
`vite`'s start implementation was not running `configure`, not was it validating peer deps.

This PR hoists common functionality between the webpack and vite start handlers into the start action. Additionally, some slight optimizations to promise handling and logging were made.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,8 +4,9 @@ coverage/
 dist/
 dist-ssr/
 dist-build/
-dist-start/
 storybook-static/
+custom-output-directory/
+dist-start/
 report/
 template/
 .changeset/*.md
--- a/packages/sku/src/program/commands/start/start.action.ts
+++ b/packages/sku/src/program/commands/start/start.action.ts
@@ -1,6 +1,10 @@
 import type { StatsChoices } from '@/program/options/stats/stats.option.js';
 import type { SkuContext } from '@/context/createSkuContext.js';
 import type { Command } from 'commander';
+import { configureProject, validatePeerDeps } from '@/utils/configure.js';
+import { watchVocabCompile } from '@/services/vocab/runVocab.js';
+import { checkHosts, withHostile } from '@/utils/contextUtils/hosts.js';
+import chalk from 'chalk';
 
 export const startAction = async (
   {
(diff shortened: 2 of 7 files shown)


.prettierignore
```diff
@@ -4,8 +4,9 @@ coverage/
 dist/
 dist-ssr/
 dist-build/
-dist-start/
 storybook-static/
+custom-output-directory/
```
askoufis, line 8: This is necessary for the new fixture added in #1314.

packages/sku/src/program/commands/start/start.action.ts
```diff
@@ -13,6 +17,17 @@ export const startAction = async (
...
   const { environment } = command.optsWithGlobals();
+
+  console.log(chalk.blue(`sku start`));
+
+  await Promise.all([
+    await configureProject(skuContext),
+    await watchVocabCompile(skuContext),
+  ]);
+
+  withHostile(checkHosts)(skuContext);
```
askoufis, line 28: `checkHosts`, just like `validatePeerDeps`, primarily emits warning logs, so there's no need for it to be awaited.

williamlark (approved):
Much better 💯

packages/sku/src/program/commands/start/webpack-start-handler.ts
```diff
@@ -54,19 +48,11 @@ export const webpackStartHandler = async ({
...
-  console.log(chalk.blue(`sku start`));
-
-  await watchVocabCompile(skuContext);
-
   const environment = resolveEnvironment({
     environment: environmentOption,
     skuContext,
   });
 
-  console.log();
```
askoufis, line 68: This spacing doesn't really feel necessary anymore.

packages/sku/src/utils/validatePeerDeps.ts
```diff
@@ -133,5 +133,3 @@ const validatePeerDeps = async ({ paths }: SkuContext) => {
     console.error(e);
   }
 };
-
-export default validatePeerDeps;
```
askoufis, line 137: default export bad
image

PR #1325, opened 2025-07-30
`vitest`: Fix watch mode

askoufis:
The `createVitest` API we were using is more tailored towords creating custom test-runners/micro-managing vitest for specific use cases. We don't really need that much control, so I've swapped it out for the [`startVitest`](https://vitest.dev/advanced/api/#startvitest) API. This API both creates vitest instance and starts a test run, _and_ it works with watch mode.

I've also refactored the vitest handler a little bit for clarity, and simplified the vitest config a little bit.

Finally, added a TODO about adding some tests for watch mode. I think this should be much easier thanks to #1317, but don't have the bandwidth to work on that right now.

### DIFF
--- a/.changeset/flat-dingos-drive.md
+++ b/.changeset/flat-dingos-drive.md
@@ -1,0 +1,6 @@
+---
+'@sku-lib/vitest': patch
+'sku': patch
+---
+
+`vitest`: Fix watch mode
--- a/.changeset/flat-dingos-drive.md
+++ b/.changeset/flat-dingos-drive.md
@@ -3,4 +3,10 @@
 'sku': patch
 ---
 
-`vitest`: Fix watch mode
+`vitest`: Honour watch mode when running `sku test`
+
+Previously it was not possible to run Vitest in watch mode via `sku test`. `sku test` should now behave similarly to [the `vitest` CLI][vitest cli], so `sku test`, `sku test watch` and `sku test -w/--watch` will all trigger watch mode.
+
+To run tests once and exit, you can execute `sku test run`, `sku test --run` or `sku test --no-watch` .
+
+[vitest cli]: https://vitest.dev/guide/cli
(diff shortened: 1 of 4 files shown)


packages/vitest/src/index.ts
```diff
@@ -10,26 +10,18 @@ export const runVitest = async ({
...
 
-  const vitest = await createVitest(
+  await startVitest(
     'test',
+    results.filter,
     { config: false, ...results.options },
     {
       plugins: [vanillaExtractPlugin()],
       test: {
-        watch: false,
```
askoufis, line 19: We want the `results.options` config above to handle watch mode.

packages/vitest/src/index.ts
```diff
@@ -10,26 +10,18 @@ export const runVitest = async ({
...
         environment: 'jsdom',
         globals: true,
         setupFiles,
-        include: [
-          '**/__tests__/**/*.test.{js,jsx,ts,tsx}',
-          '**/?(*.)+(spec|test).{js,jsx,ts,tsx}',
-        ],
```
askoufis, line 26: These are the default include values so this config was redundant.

.changeset/flat-dingos-drive.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sku-lib/vitest': patch
+'sku': patch
+---
+
+`vitest`: Fix watch mode
```
williamlark: Maybe something a little more descriptive here? Suggestion:


```suggestion
`vitest`: Allow watch mode flag (`-w`/`--watch`) to be passed
```

williamlark (approved):

.changeset/flat-dingos-drive.md
askoufis: You could always pass the flag in, it just wouldn't work, hence `fix`. But I agree it could be more descriptive. Also the `watch` argument works too. Updated in https://github.com/seek-oss/sku/pull/1325/commits/04b9dacdca4cee56509b81c77d02f9e5f7d39210.

PR #1327, opened 2025-07-30
`vite`:  Add `displayNamesProd` support

adenj:
Adds `displayNamesProd` configuration support to the experimental Vite bundler to achieve feature parity with webpack


.changeset/huge-ghosts-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': minor
+---
+
+`vite`: Add support for `displayNameProd` `sku.config option
```
jahredhope: Unintentional back-tick?

```suggestion
`vite`: Add support for `displayNameProd` sku.config option
```

packages/sku/src/services/vite/helpers/config/baseConfig.ts
```diff
@@ -24,7 +24,8 @@ const getBaseConfig = (skuContext: SkuContext): InlineConfig => {
   const vocabConfig = getVocabConfig(skuContext);
 
   const isProductionBuild = process.env.NODE_ENV === 'production';
-  const prodBabelPlugins = [
+
+  const prodBabelPlugins: Array<string | [string, any]> = [
```
askoufis: `object` would be _slightly_ better than using `any` here.
```suggestion
  const prodBabelPlugins: Array<string | [string, object]> = [
```

tests/browser/display-names-prod.test.ts
```diff
@@ -0,0 +1,71 @@
+import {
+  describe,
+  beforeAll,
+  afterAll,
+  it,
+  expect as globalExpected,
```
askoufis:
```suggestion
  expect as globalExpect,
```

fixtures/display-names-prod/src/App.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const Card: WithDisplayName = () => <div>This is a card</div>;
+
+const Header: WithDisplayName<{ title: string }> = ({ title }) => (
+  <h1>{title}</h1>
+);
+
+const MemoFooter = React.memo(() => <footer>Footer content</footer>);
+
+const Input: WithDisplayName = () => <input />;
```
askoufis, line 17: Braid manually sets `displayName` on [some components](https://github.com/seek-oss/braid-design-system/blob/82c6441d4636e2db3e66d8c4ed97ab953735fe69/packages/braid-design-system/src/lib/components/Box/Box.tsx#L118-L139), so it would be good to test that a manual `displayName` isn't overwritten by the babel plugin.

For example:

```ts
const PublicBox: WithDisplayName = () => <div>PublicBox</div>;
PublicBox.displayName = 'Box';
```

tests/browser/display-names-prod.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        ).toBeInTheConsole();
+      });
+
+      it('should create build output', async ({ expect }) => {
+        const distDir = fixturePath('dist');
+        const files = await dirContentsToObject(distDir);
+        expect(files).toMatchSnapshot();
+      });
+
+      it('should add displayName assignments to built JavaScript', async ({
```
askoufis: Is there any advantage to checking the JavaScript, relative to snapshotting the HTML? I'd expect the HTML to show `undefined` dislaynames if the javascript was broken.

Additionally, we should run the same tests for `sku start` to ensure displaynames aren't there.

fixtures/display-names-prod/src/App.tsx
adenj: Good shout. Added in a manually assigned displayName and the babel plugin doesn't seem to modify it

askoufis (approved):

PR #1329, opened 2025-07-31
Replace TSConfig paths with subpath imports

askoufis:
Soz dan.

Building packages containing tsconfig paths imports with `tsc` requires `ts-patch`, whereas [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) are native to Node and supported by TypeScript with `moduleResolution: node16/nodenext`.

I'm honestly still not sold on the need for these TBH. They _can_ shorten import paths if you need to traverse upwards a lot, but they can also lengthen them in some cases. It also results in two different ways of importing the same thing, instead of one. But removing them is more hassle than renaming them, so I'm happy to just move of using `ts-patch` for now.

### DIFF
--- a/packages/sku/package.json
+++ b/packages/sku/package.json
@@ -33,6 +33,9 @@
     "./webpack-plugin": "./dist/services/webpack/config/plugins/sku-webpack-plugin/index.js",
     "./package.json": "./package.json"
   },
+  "imports": {
+    "#src/*": "./src/*"
+  },
   "engines": {
     "node": ">=20.15"
   },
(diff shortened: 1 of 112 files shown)


packages/sku/package.json
```diff
@@ -33,6 +33,9 @@
     "./webpack-plugin": "./dist/services/webpack/config/plugins/sku-webpack-plugin/index.js",
     "./package.json": "./package.json"
   },
+  "imports": {
+    "#src/*": "./src/*"
```
askoufis, line 37: This could be shortened to something like `#@`, but I like the explicitness of `#src` - it makes it obviousl that you're resolving `src` from the root directory.

askoufis:
nvm this doesn't work right now. i still don't like tsconfig paths though.

PR #1330, opened 2025-07-31
Undo tsconfig paths (round 2)

askoufis:
Sorry Dan.

In libraries I don't think tsconfig paths solve any actual problems. Shortening import paths is not a real problem because of auto-importing. Instead, they complicate the build process and make it possible to import the same file in multiple different ways, instead of just 1 way.

[Subpath imports](https://nodejs.org/api/packages.html#subpath-imports) exist, and are natively supported by node and TypeScript. They aren't quite a replacement for the tsconfig paths stuff we were doing, but _not_ relying on tsconfig paths potentially enables us to use subpath imports in the future to enable some nice dev-time optimizations, such as testing against source code rather than requiring a build step.

All `@/` references have been replaced with relative imports, courtesy of https://github.com/jaredpalmer/codemods.

I've also removed [`baseUrl`](https://www.typescriptlang.org/tsconfig/#baseUrl) from all `tsconfigs`, as [it will be deprecated in TypeScript v6 and removed in v7](https://github.com/microsoft/typescript-go/issues/1431#issuecomment-3099848803) (there might be more formal documentation about this but this was the link I had at hand). It wasn't really doing anything anyway.

### DIFF
--- a/packages/sku/src/program/commands/translations/commands/validate/validate.action.ts
+++ b/packages/sku/src/program/commands/translations/commands/validate/validate.action.ts
@@ -2,8 +2,8 @@ import { validate } from '@vocab/core';
 import chalk from 'chalk';
 
 import { getResolvedVocabConfig } from '../../helpers/translation-helpers.js';
-import { configureProject } from '@/utils/configure.js';
-import type { SkuContext } from '@/context/createSkuContext.js';
+import { configureProject } from '../../../../../utils/configure.js';
+import type { SkuContext } from '../../../../../context/createSkuContext.js';
 
 const log = (message: string) => console.log(chalk.cyan(message));
 
--- a/packages/codemod/package.json
+++ b/packages/codemod/package.json
@@ -5,8 +5,7 @@
   "type": "module",
   "bin": "./bin.js",
   "scripts": {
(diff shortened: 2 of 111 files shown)


packages/sku/src/program/commands/translations/commands/validate/validate.action.ts
```diff
@@ -2,8 +2,8 @@ import { validate } from '@vocab/core';
 import chalk from 'chalk';
 
 import { getResolvedVocabConfig } from '../../helpers/translation-helpers.js';
-import { configureProject } from '@/utils/configure.js';
-import type { SkuContext } from '@/context/createSkuContext.js';
+import { configureProject } from '../../../../../utils/configure.js';
+import type { SkuContext } from '../../../../../context/createSkuContext.js';
```
williamlark, line 6: For me, seeing how deeply we are having to import some of things makes me think we could improve this structure. So, I agree with these changes. The shortening is nice, but I feel like it hides improvements we could be making. 👍

askoufis: Yeah absolutely. Have wanted to restructure things for a while but it can be quite disruptive, so best to do it at an opportune time.

PR #1332, opened 2025-08-01
`vite start`: Inline Vanilla Extract CSS to fix FOUC

askoufis:
The primary change is adding [`unstable_mode: 'inlineCssInDev'`](https://github.com/vanilla-extract-css/vanilla-extract/pull/1614) to the VE Vite plugin. The rest of the changes are some slight refactors to the middleware plugin and snapshot updates.

I tried to implement critical style extraction with [`beasties`](https://github.com/danielroe/beasties), which _did_ work, but also caused hydration errors because it was removing [this inline style tag injected by braid](https://github.com/seek-oss/braid-design-system/blob/82c6441d4636e2db3e66d8c4ed97ab953735fe69/packages/braid-design-system/src/lib/components/BraidProvider/BraidProvider.tsx#L98-L101). My efforts to stop it from doing that failed, so I removed the change. If critical CSS during dev is something we really care about, we can look further into it, but at this point in time I don't see a reason to put any effort into it.

As a side note, this change results in duplicate CSS in the DOM: the inlined style tag from the VE Vite plugin, as well as the individual stylesheets injected when the Vite client loads the VE styles (again). We _can_ remove the inlined style tag after all DOM content has loaded, and in fact I got it to work with no issues in a simple fixture. However, once again I don't think it's worth doing this as these styles are only inlined at dev and otherwise don't affect anything.

### DIFF
--- a/packages/sku/src/services/vite/plugins/middlewarePlugin.ts
+++ b/packages/sku/src/services/vite/plugins/middlewarePlugin.ts
@@ -43,8 +43,10 @@ export const middlewarePlugin = (skuContext: SkuContext): Plugin => ({
     // Code returned from here will run *after* Vite's middleware. Lazy loading middleware should happen outside (before) this return.
     return async () => {
       log('Configuring server middleware');
+
       server.middlewares.use('/', async (req, res, next) => {
         log('Handling request:', req.url);
+
         // Check that the request is for the index.html file (i.e., a valid route in our app)
         // If we don't check for index.html, the middleware will run for every request, including static assets.
         if (!req.originalUrl || req.url !== '/index.html') {
@@ -63,6 +65,7 @@ export const middlewarePlugin = (skuContext: SkuContext): Plugin => ({
             'Internal error occurred. No hostname found when handling middleware. Host header required.',
           );
         }
+
         const site =
           skuContext.sites.find((skuSite) => skuSite.host === hostname) ||
           skuContext.sites[0];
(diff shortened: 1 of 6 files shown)


packages/sku/src/services/vite/plugins/middlewarePlugin.ts
```diff
@@ -93,7 +96,7 @@ export const middlewarePlugin = (skuContext: SkuContext): Plugin => ({
         try {
           const { viteRender } = await server.ssrLoadModule(renderEntry);
 
-          const renderedHtml = await (viteRender as ViteRenderFunction)({
+          let html = await (viteRender as ViteRenderFunction)({
```
askoufis, line 99: Made these changes when experimenting with `beasties`, and decided to keep them. Having a single `html` variable that's incrementally modified makes it much easier to transform the HTML over separate steps, rather than dealing with many different named html constants while also ensuring you refer to the correct one in the next transformation.

packages/sku/src/services/vite/plugins/middlewarePlugin.ts
```diff
@@ -43,8 +43,10 @@ export const middlewarePlugin = (skuContext: SkuContext): Plugin => ({
     // Code returned from here will run *after* Vite's middleware. Lazy loading middleware should happen outside (before) this return.
     return async () => {
       log('Configuring server middleware');
+
```
askoufis, line 46: Y'all need ~Jesus~ more whitespace.

PR #1333, opened 2025-08-01
Fix sku init snapshot versioning

williamlark:
Tests for sku-init fail on Version branches because the pnpm-workspace file added to the sku-init fixture makes the sku package unavailable in the workspace, making sku-init try to fetch an unpublished version of sku from the npm registry. This changes makes sku available in the fixtures workspace and turns on linking from a workspace when available (with `linkWorkspacePackages`)


askoufis (approved):
Awesome!
linkWorkspacePackages

PR #1334, opened 2025-08-04
`bundler`: Split SkuConfig types by bundler

adenj:
Implements bundler-specific configuration types

### DIFF
--- a/packages/sku/src/context/createSkuContext.ts
+++ b/packages/sku/src/context/createSkuContext.ts
@@ -189,10 +189,7 @@ export const createSkuContext = ({
   const libraryFile = skuConfig.libraryFile!;
   const isLibrary = Boolean(skuConfig.libraryEntry);
   const polyfills = skuConfig.polyfills!;
-  const webpackDecorator =
-    skuConfig.__UNSAFE_EXPERIMENTAL__bundler === 'vite'
-      ? (config: any) => config
-      : skuConfig.dangerouslySetWebpackConfig!;
+  const webpackDecorator = skuConfig.dangerouslySetWebpackConfig!;
   const jestDecorator = skuConfig.dangerouslySetJestConfig!;
   const eslintDecorator = skuConfig.dangerouslySetESLintConfig!;
   const tsconfigDecorator = skuConfig.dangerouslySetTSConfig!;
--- a/packages/sku/src/context/createSkuContext.ts
+++ b/packages/sku/src/context/createSkuContext.ts
@@ -191,7 +191,7 @@ export const createSkuContext = ({
   const polyfills = skuConfig.polyfills!;
   const webpackDecorator =
     skuConfig.__UNSAFE_EXPERIMENTAL__bundler === 'vite'
-      ? <T>(config: T): T => config
+      ? (config: any) => config
       : skuConfig.dangerouslySetWebpackConfig!;
   const jestDecorator = skuConfig.dangerouslySetJestConfig!;
   const eslintDecorator = skuConfig.dangerouslySetESLintConfig!;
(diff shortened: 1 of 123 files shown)


packages/sku/src/utils/SkuConfigUpdater.ts
```diff
@@ -146,12 +159,12 @@ export class SkuConfigUpdater {
...
    * This method does not write the changes to the file system. Use `commitConfig` to do that.
    */
-  upsertConfig<T extends keyof SkuConfig>({
+  upsertConfig<T extends AllSkuConfigKeys>({
     property,
     value,
   }: {
     property: T;
-    value: SkuConfig[T];
+    value: AllSkuConfigValues;
```
askoufis: This isn't quite the same as the previous type. `AllSkuConfigValues` is a union of all possible values of sku config properties. `SkuConfig[T]` however was the value for the specific sku config property specified in `property`.

Clearly `SkuConfig[T]` won't work anymore, but you might need to extract a separate `JoinedSkuConfig` type that is `SkuConfigBase & WebpackSkuConfig & ViteSkuConfig`, and then do `JoinedSkuConfig[T]`.

Happy to pair on this if you need.

packages/sku/src/context/createSkuContext.ts
```diff
@@ -189,7 +189,10 @@ export const createSkuContext = ({
   const libraryFile = skuConfig.libraryFile!;
   const isLibrary = Boolean(skuConfig.libraryEntry);
   const polyfills = skuConfig.polyfills!;
-  const webpackDecorator = skuConfig.dangerouslySetWebpackConfig!;
+  const webpackDecorator =
+    skuConfig.__UNSAFE_EXPERIMENTAL__bundler === 'vite'
+      ? (config: any) => config
+      : skuConfig.dangerouslySetWebpackConfig!;
```
williamlark: I'm not a huge fan of having to check the bundler type here. We wouldn't be using this value in the vite stuff anyway, and we aren't doing any checks to enforce a different context for the vite and webpack bundle.  So, we could still set this in the defaults like the others I reckon and reference it directly here.


```suggestion
  const webpackDecorator = skuConfig.dangerouslySetWebpackConfig!;
```

packages/sku/src/context/defaultSkuConfig.ts
```diff
@@ -8,7 +8,6 @@ const defaultDecorator = <T>(a: T) => a;
 export default {
   __UNSAFE_EXPERIMENTAL__bundler: 'webpack',
   __UNSAFE_EXPERIMENTAL__testRunner: 'jest',
-  __UNSAFE_EXPERIMENTAL__cjsInteropDependencies: [],
```
williamlark, line 11: I reckon we could keep these defaults so we can refer to the values without having to check the bundler type. Perhaps we could be look in to validating the config at runtime with zod or valibot or something eventually, but until then this function could set values for both types of bundlers.

askoufis: The config is validated at runtime with `fastest-validator`, but we don't infer types from that validator like you would with zod. `SkuContext` is in this weird state of _sort of_ representing a parsed sku config, but it's mixed in with default values and "extra" properties so it's very messy.

I'd be very much in favour of using zod or something similar to properly validate the config, and split out separate types for internal use so we can finally get rid of all these exclamation marks!

packages/sku/src/context/createSkuContext.ts
askoufis: `webpackDecorator` has to be set to something because it's mapped over the webpack config. We might be able to get away with this because it should default to an identity function, but also there's a `!` on this so who knows what's real anymore.

Definitely highlights the need to refactor our config validation and types.

askoufis (approved):
I'm happy with the PR, though we should wait for @williamlark's thoughts on the outstanding conversation about `dangerouslySetWebpackConfig`.

williamlark (approved):
Noice 🥇

PR #1336, opened 2025-08-04
Emit telemetry debug logs even when `@seek/sku-telemetry` is not installed, restrict metric names to strict set

askoufis:
While working on #1335 I wanted to see the value being emitted by telemetry, but we can't install our private telemetry package in the repo. I opted to instead hijack the `noop` function in the placeholder telemetry provider so it emits debug logs.

The debug logs look like this:
image

I also make a string union of all the metric names we currently emit and updated our `TelemetryProvider` type to only accept those string types. This is 1 step towards having a better, type-safe telemetry interface.

### DIFF
--- a/tests/node/__snapshots__/custom-output-target.test.ts.snap
+++ b/tests/node/__snapshots__/custom-output-target.test.ts.snap
@@ -1,6 +1,47 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`custom-output-target > should generate an output directory with the value specified in sku.config 1`] = `
+exports[`bundler vite > custom-output-target > should generate an output directory with the value specified in sku.config 1`] = `
+{
+  "index.html": SCRIPTS: [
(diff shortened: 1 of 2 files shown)


packages/sku/src/services/telemetry/provider.ts
```diff
@@ -3,26 +3,42 @@ import { createRequire } from 'node:module';
 import { getAddCommand } from '../packageManager/packageManager.js';
 import banner from '../../utils/banners/banner.js';
 
+import _debug from 'debug';
+
 const require = createRequire(import.meta.url);
 
-function noop() {}
+const debug = _debug('sku:telemetry');
```
askoufis, line 10: This is the same debug log namespace used by `@seek/sku-telemetry`. So if you enable it, you'll either see the noop logs, or the real logs, but never both.

williamlark (commented):
Very nice 🎉 

Edit: apparently I forgot to hit approve

williamlark (approved):

PR #1337, opened 2025-08-05
Move `plop` code and dependencies into a private package

askoufis:
Plop stuff has been moved to its own private package, which will help de-clutter the repo root a little bit. Additionally:
- Renamed all files to `.js` as they were likely `.mjs` from before sku was converted to ESM
- Removed the `change-case` dependency and camel case helper as plop contains an inbuilt `camelCase` helper


package.json
```diff
@@ -3,8 +3,7 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "plop": "plop",
-    "install:sku": "pnpm install --filter sku",
```
askoufis, line 7: Not sure what this script was used for, but it's unused AFAICT.

private/plop/plopfile.js
```diff
@@ -0,0 +1,9 @@
+import { addSkuCommandGenerator } from './generators/add-sku-command/index.js';
+import { addSkuCodemodGenerator } from './generators/add-sku-codemod/index.js';
+
+/** @param {import('plop').NodePlopAPI} plop*/
```
askoufis, line 4: The `plop` CLI doesn't support `.ts` files directly, so we'd have to compile it from TypeScript, which I don't think is worth it, so I added JSDoc type annotations instead.

williamlark (approved):
Nice to see this separated 👌

PR #1339, opened 2025-08-06
`vite`: add pathAliases config for custom import paths

adenj:
Adds support for custom path aliases in Vite projects via a new `pathAliases` configuration option.


packages/sku/src/types/types.ts
```diff
@@ -520,6 +520,21 @@ export interface ViteSkuConfig {
...
+   * Note: sku automatically provides a 'src/*' alias that maps to './src/*'.
+   * This option is only needed for additional custom aliases.
+   *
+   * Example: [{ "components/*": "./src/components/*" }, { "utils/*": "./src/utils/*" }]
+   *
+   * Note: This option is only relevant when using the `vite` bundler.
+   *
+   * @default []
+   */
+  pathAliases?: Array<Record<string, string>>;
```
askoufis: Does this have to be an array of records? A `Record<string, string>` should be enough here, and would simplify some of the logic.

tests/node/path-aliases.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      expect(tsconfig.compilerOptions.paths).toBeDefined();
+    });
+  });
+
+  describe('validation', () => {
+    it('should reject pathAliases pointing to node_modules', async ({
+      expect,
+    }) => {
+      const configure = await sku('configure', ['--config=sku.config.bad.ts']);
+
```
askoufis, line 60: We should assert on the console output here too.

packages/sku/src/context/createSkuContext.ts
```diff
@@ -97,6 +97,22 @@ export const createSkuContext = ({
     process.exit(1);
   }
 
+  // Validate pathAliases destinations don't contain node_modules
+  if (skuConfig.pathAliases) {
+    for (const aliasObj of skuConfig.pathAliases) {
+      for (const [alias, destination] of Object.entries(aliasObj)) {
+        if (destination.match(/(?:^|\/)node_modules(?:\/|$)/)) {
```
askoufis: Not sure if there's a benefit to a regexp here. String `includes` should work for this use case:
```suggestion
        if (destination.includes('node_modules')) {
```

packages/sku/src/context/createSkuContext.ts
```diff
@@ -208,6 +224,28 @@ export const createSkuContext = ({
     skuConfig.skipPackageCompatibilityCompilation!;
   const externalizeNodeModules = skuConfig.externalizeNodeModules!;
 
+  // Generate TypeScript paths for Vite pathAliases
+  const tsPaths =
+    skuConfig.__UNSAFE_EXPERIMENTAL__bundler === 'vite'
+      ? (() => {
```
askoufis: Can this be moved into its own function that takes `sku.pathAliases` as an input?

packages/sku/src/types/types.ts
```diff
@@ -520,6 +520,21 @@ export interface ViteSkuConfig {
...
+   * Each alias maps a pattern to a destination path relative to the project root.
+   *
+   * Note: sku automatically provides a 'src/*' alias that maps to './src/*'.
+   * This option is only needed for additional custom aliases.
+   *
+   * Example: [{ "components/*": "./src/components/*" }, { "utils/*": "./src/utils/*" }]
+   *
+   * Note: This option is only relevant when using the `vite` bundler.
+   *
+   * @default []
```
askoufis: I think we should default this to `'src/*': './src/*'` rather than having an implicit default that's opaque to the user.

packages/sku/src/context/createSkuContext.ts
```diff
@@ -208,6 +224,28 @@ export const createSkuContext = ({
     skuConfig.skipPackageCompatibilityCompilation!;
   const externalizeNodeModules = skuConfig.externalizeNodeModules!;
 
+  // Generate TypeScript paths for Vite pathAliases
+  const tsPaths =
+    skuConfig.__UNSAFE_EXPERIMENTAL__bundler === 'vite'
```
askoufis, line 252: Was thinking, maybe this option should work for webpack too. It means that consumers can migrate to it _before_ moving to vite, and potentially stop using `dangerouslySetTSConfig`.

.changeset/fine-ducks-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': minor
+---
+
+  Add pathAliases config for custom import paths in Vite projects. Automatically provides src/* alias and enables custom aliases like @components/*
```
askoufis: We can flesh this changeset out more (or I guess combine it with the rest of them). _If_ we end up supporting it in webpack mode as well, we should give examples of migrating from `dangerouslySetTSConfig` to `pathAliases`.

packages/sku/src/types/types.ts
```diff
@@ -520,6 +520,21 @@ export interface ViteSkuConfig {
    */
   __UNSAFE_EXPERIMENTAL__cjsInteropDependencies?: string[];
 
+  /**
+   * An array of path alias mappings for module resolution.
+   * Each alias maps a pattern to a destination path relative to the project root.
```
askoufis, line 525: I think it would be worth mentioning that this affects your `tsconfig` paths. Maybe even worth linking to https://www.typescriptlang.org/docs/handbook/modules/reference.html#paths.

packages/sku/src/services/typescript/config/tsconfig.ts
```diff
@@ -42,12 +42,10 @@ export default ({ rootResolution, tsconfigDecorator }: SkuContext) => {
       target: 'es2022',
       ...(rootResolution
         ? {
-            paths: {
-              '*': ['*'],
-            },
```
askoufis, line 47: This is a bugfix so we should add a changeset for it.

packages/sku/src/context/createSkuContext.ts
adenj: I originally did use `includes()`, but was conscious that in the (admittedly very unlikely) case that someone has a custom path that does include "node_modules" string it would return a false positive. 
The chances of that are slim-to-nil so I'm happy to change it back

askoufis: Not sure I'm following. IMO any situation where the desination involves `node_modules`, regardless of whether it's actually related to package dependencies, should be an error.

fixtures/path-aliases/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@sku-fixtures/custom-output-target",
+  "private": true,
+  "type": "module",
+  "dependencies": {
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
```
williamlark: Lets bump these deps and devDeps to 19. The React 19 stuff is in so only the react-18 fixture should use that version now

packages/sku/src/context/createSkuContext.ts
```diff
@@ -226,6 +264,7 @@ export const createSkuContext = ({
     eslintDecorator,
     tsconfigDecorator,
     eslintIgnore,
+    tsPaths,
```
williamlark, line 274: Could this be called the same as the user option (`pathAliases`)? Just for the sake of not having to remember what public properties map to internal properties

tests/node/path-aliases.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    it('should always include automatic src/* alias for Vite', async ({
+      expect,
+    }) => {
+      const tsconfigPath = fixturePath('tsconfig.json');
+      const tsconfigContents = await readFile(tsconfigPath, 'utf-8');
+      const tsconfig = jsonc.parse(tsconfigContents);
+
+      expect(tsconfig.compilerOptions.paths).toHaveProperty('src/*');
+      expect(tsconfig.compilerOptions.paths['src/*']).toEqual(['./src/*']);
```
williamlark: nit: could also use `toMatchObject` or `objectContaining` to check what the object looks like in one expect

packages/sku/src/context/createSkuContext.ts
adenj: This would cause a naming collision since we are already returning `pathAliases` further down.
They also represent different purposes imo.
`tsPaths` are the processed typescript paths config that go into `tsconfig.json`, whereas `pathAliases` is the original user config that webpack is consuming

packages/sku/src/services/webpack/config/webpack.config.ts
```diff
@@ -64,6 +66,7 @@ const makeWebpackConfig = ({
   const vocabOptions = getVocabConfig(skuContext);
 
   const resolvedPolyfills = resolvePolyfills(polyfills);
+  const webpackAliases = convertPathAliasesToWebpackAliases(pathAliases);
```
askoufis: Didn't realize this would needed for webpack. I guess it sort of doubles up on `srcPaths` config too. 🤦 

Maybe we back out of the webpack stuff and stick with vite-only for now. Sorry for the back and forth 🙏.

fixtures/path-aliases/package.json
```diff
@@ -0,0 +1,20 @@
...
+  },
+  "scripts": {
+    "start": "sku start",
+    "start:vite": "sku start --experimental-bundler --config sku.config.vite.ts",
+    "build:vite": "sku build --experimental-bundler --config sku.config.vite.ts"
+  },
+  "devDependencies": {
+    "@sku-private/test-utils": "workspace:*",
+    "@types/react": "^18.2.3",
+    "@types/react-dom": "^18.2.3",
```
askoufis: These type deps also need to be updated to v19.

.changeset/fine-ducks-peel.md
```diff
@@ -0,0 +1,69 @@
...
+
+This enables imports like:
+```typescript
+import { Button } from '@components/Button';
+import { formatDate } from '@utils/dateHelpers';
+```
+
+## Migration from dangerouslySetTSConfig
+
+If you're currently using `dangerouslySetTSConfig` for TypeScript path mapping, you can migrate to the cleaner `pathAliases` approach:
```
askoufis:
```suggestion
If you're currently using `dangerouslySetTSConfig` exclusively to configure `paths`, you can migrate to the cleaner `pathAliases` approach:
```

docs/docs/configuration.md
```diff
@@ -311,6 +311,46 @@ The file name of the library. The main bundle of the library will be output to `
...
+
+Type: `Record<string, string>`
+
+Default: `{ 'src/*': './src/*' }`
+
+**Only available with Vite bundler**
+
+Custom path alias mappings for module resolution. Each alias pattern maps to a destination path relative to the project root.
+
+When using the `vite` bundler, sku provides a default `src/*` alias that maps to `./src/*`. The `pathAliases` option allows you to define additional custom aliases.
```
askoufis: We've already mentioned that this is exclusive to Vite so I think this can be removed.

```suggestion
sku provides a default `src/*` alias that maps to `./src/*`. The `pathAliases` option allows you to define additional custom aliases.
```

tests/node/path-aliases.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    it('should reject pathAliases pointing to node_modules', async ({
+      expect,
+    }) => {
+      const configure = await sku('configure', ['--config=sku.config.bad.ts']);
+
+      await waitFor(() => {
+        expect(configure.hasExit()).toMatchObject({ exitCode: 1 });
+        expect(configure.getStdallStr()).toContain(
+          'Path alias "@bad/*" cannot point to node_modules.',
+        );
```
askoufis: I think the second `expect` can go outside the `waitFor`. We _should_ only need to wait for the exit code, at which point an assert on stderr/stdout should succeed without needing to wait for anything.

packages/sku/src/context/createSkuContext.ts
```diff
@@ -20,6 +20,26 @@ const jiti = createJiti(__filename);
...
 const debug = _debug('sku:config');
 
+const generateTypeScriptPaths = (
+  pathAliases?: Record<string, string>,
+): Record<string, string[]> => {
+  const typeScriptPaths: Record<string, string[]> = {};
+
+  // Always include default src/* alias, then merge with user-provided aliases
+  const mergedAliases = {
+    ...defaultSkuConfig.pathAliases,
```
askoufis: I think we might actually not want to include the `src/*` default in `defaultSkuConfig.pathAliases`. `defaultSkuConfig` is really meant to hold default values for `undefined` fields, not be merged into user-provided values.

So rather than specifying `{ 'src/*': './src/*' }` as the default (both in `defaultSkuConfig` and in the docs), we should instead just say "you always get this mapping no matter what", and merge it in manually.

tests/node/path-aliases.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    }) => {
+      const configure = await sku('configure', ['--config=sku.config.bad.ts']);
+
+      await waitFor(() => {
+        expect(configure.hasExit()).toMatchObject({ exitCode: 1 });
+      });
+
+      expect(configure.getStdallStr()).toContain(
+        'Path alias "@bad/*" cannot point to node_modules.',
+      );
```
williamlark: You can use the `getBy` commands here too.


```suggestion
      expect(
        configure.getByText(
          'Path alias "@bad/*" cannot point to node_modules.',
        ),
      ).toBeInTheConsole();
```

williamlark (approved):
Looks good to me. Nice one 💯

PR #1340, opened 2025-08-11
Test changes

williamlark:
Making some changes to try to reduce flakiness on the snapshots.

* I've changed how diffs work a little. Rather than outputting a seperate "POST HYDRATE DIFFS" section, the diffs will display inline. This makes it a lot easier to tell where the diffs are actually present in the dom and makes reviewing changes easier to read. This makes most of the snapshots a little smaller as well. 
* Disabled caching in puppeteer. This was causing some flakiness and also making a false positive appear in the ssr-hello-world test. The explicit publicPath makes the urls fail on the custom-port test, but puppeteer wasn't raising warning because of the cache it seems.
* I've isolated the braid tests. I removed this at one point, but I'm adding it back since they are too inconstant when run with everything else. I'm also resetting the browser object and deleting the cache just in case (now I don't trust anything 🙈)
* Turned on incognito mode for puppeteer. I don't know if this will help, but turning it on just in case some caching is playing with the snapshots.

### DIFF
--- a/packages/sku/src/services/webpack/config/webpack.config.ts
+++ b/packages/sku/src/services/webpack/config/webpack.config.ts
@@ -18,7 +18,7 @@ import { cwd } from '../../../utils/cwd.js';
 import { getVocabConfig } from '../../vocab/config/vocab.js';
 import getStatsConfig from './statsConfig.js';
 import getSourceMapSetting from './sourceMaps.js';
-import getCacheSettings from './cache.js';
+import getCacheSettings, { disableCacheOverride } from './cache.js';
 import modules from './resolveModules.js';
 import targets from '../../../config/targets.json' with { type: 'json' };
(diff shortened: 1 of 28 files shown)


packages/sku/src/services/webpack/config/webpack.config.ts
```diff
@@ -195,7 +195,7 @@ const makeWebpackConfig = ({
                       loader: require.resolve('babel-loader'),
                       options: {
                         babelrc: false,
-                        cacheDirectory: true,
+                        cacheDirectory: !disableCacheOverride,
```
williamlark: This touches the webpack stuff but I don't know if it deserves a release. The variable is only used internally so I don't think it needs one.

adenj (approved):

askoufis (approved):

PR #1342, opened 2025-08-11
Centralising string serialisation in test

williamlark:
A lot of our snapshots share the same requirements for serialisation, so I've centralised it into its own sanitiser. App snapshots and directory snapshots will now share the same serialisation.

(I don't know whether it's `serialisation` or `serialization`... We use an `s` in AU, but the vite types use a `z`. I just stuck with `z` to fit with the vite stuff.)

I've also added a vite plugin in the fixtures to normalise the hashes (removes them) matching what we do for webpack. 

I also deleted a fixture we don't use anywhere. I think it was meant to be a testing some vite stuff, but we don't use it. Happy to keep it if anyone objects 🧹


askoufis:
Briefly tested modifying `assetFileNames` and `chunkFileNames`, but couldn't get them to work for some reason.

private/puppeteer/sanitizeFilesSerializer.ts
```diff
@@ -0,0 +1,18 @@
+import type { SnapshotSerializer } from 'vitest';
+import { sanitizeString } from './sanitizeString.ts';
+
+/** Formats CSS strings */
```
askoufis:
```suggestion
/** Sanitizes file contents */
```

PR #1343, opened 2025-08-12
Move"rootResolution" to the webpack config types

williamlark:
The new `pathAliases` field deprecated `rootResolution` in the vite config, so it is now only a webpack option. Moving it to the webpack types to show this.

I've also added a changeset for the type changes that happened in #1339. There is a changeset for the new field in that PR, but I thought it would be nice to have one for the config separation since it's a QOL improvement for the user.

Fixed a typo too.

### DIFF
--- a/packages/sku/src/types/types.ts
+++ b/packages/sku/src/types/types.ts
@@ -357,19 +357,6 @@ export interface SkuConfigBase {
   renderEntry?: string;
 
   /**
-   * Enables root resolution.
-   *
-   * By default, sku allows importing from the root of the project e.g. `import something from 'src/modules/something'`.
-   *
-   * Unfortunately, these kinds of imports only work for apps.
-   * In packages, the imports will work locally, but fail when consumed from `node_modules`.
-   *
-   * @default true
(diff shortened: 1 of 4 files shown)


packages/sku/src/types/types.ts
```diff
@@ -505,6 +492,19 @@ export interface WebpackSkuConfig {
...
+   *
+   * By default, sku allows importing from the root of the project e.g. `import something from 'src/modules/something'`.
+   *
+   * Unfortunately, these kinds of imports only work for apps.
+   * In packages, the imports will work locally, but fail when consumed from `node_modules`.
+   *
+   * @default true
+   * @link https://seek-oss.github.io/sku/#/./docs/configuration?id=rootresolution
+   */
+  rootResolution?: boolean;
```
adenj, line 507: I think we should call this out as a `webpack`-exclusive property in the docs

williamlark: Good call. I've updated the docs to describe which config options are for webpack or vite only, and decided to add the experimental stuff in as well since we will have users starting to use it soon (in beta).

PR #1344, opened 2025-08-12
`vite`: Add bundle analyser generation

adenj:
Adds bundle analysis functionality to Vite builds using `rollup-plugin-visualizer`, providing feature parity with the existing `webpack-bundle-analyzer` integration.

 - Add `rollup-plugin-visualizer` dependency
 - Vite mode builds generate `/report/` directory
 - Uses project name from `package.json` as report title

### DIFF
--- a/tests/browser/report-generation.test.ts
+++ b/tests/browser/report-generation.test.ts
@@ -3,10 +3,10 @@ import {
   bundlers,
   type BundlerValues,
   cleanup,
-  skipCleanup,
   scopeToFixture,
 } from '@sku-private/testing-library';
 import fs from 'node:fs';
+import { rm } from 'node:fs/promises';
 
 const timeout = 50_000;
 
@@ -23,9 +23,7 @@ describe('report-generation', () => {
       beforeAll(async () => {
         // Clean up any existing report folder
         const reportDir = fixturePath('report');
-        if (fs.existsSync(reportDir)) {
-          fs.rmSync(reportDir, { recursive: true, force: true });
-        }
+        rm(reportDir, { recursive: true, force: true });
 
         const build = await sku('build', args[bundler]);
         globalExpect(
@@ -35,14 +33,12 @@ describe('report-generation', () => {
         return cleanup;
       });
 
-      it('should generate bundle analysis report', async ({ expect, task }) => {
-        skipCleanup(task.id);
+      it('should generate bundle analysis report', async ({ expect }) => {
         const reportHtml = fixturePath('report/client.html');
         expect(fs.existsSync(reportHtml)).toBe(true);
 
         // Should be a substantial file (bundle analysis reports are typically large)
         const stats = fs.statSync(reportHtml);
-        console.log({ stats });
         expect(stats.size).toBeGreaterThan(1000); // At least 1KB
       });
     });


.changeset/huge-owls-wonder.md
```diff
@@ -0,0 +1,6 @@
+---
+'sku': minor
+---
+
+Add bundle analysis support for Vite builds using [`rollup-plugin-visualizer`](https://www.npmjs.com/package/rollup-plugin-visualizer)
+Bundle analysis reports are now generated for Vite builds in the `/report/` directory
```
williamlark: nit: adding a scope makes it a bit easier to scan what APIs have changed. I also feel like the `rollup-plugin-visualizer` is an implementation details consumers don't really need to know about 🤷 (they can look at the changes if their curious)

```suggestion
`build(vite)`: Generate bundle analysis reports

Bundle analysis reports are now generated for Vite builds in the `/report/` directory. This matches the behaviour of the Webpack bundler.
```

packages/sku/src/services/vite/helpers/config/createConfig.ts
```diff
@@ -36,8 +37,17 @@ export const createServerBuildConfig = (skuContext: SkuContext) => {
...
+  const isProductionBuild = process.env.NODE_ENV === 'production';
+  const isIntegration = process.env.SKU_INTEGRATION === 'true';
+  const bundleAnalyzer =
+    isProductionBuild || !isIntegration
+      ? [bundleAnalyzerPlugin({ name: 'client' })]
+      : [];
+
   return createSkuViteConfig(
     {
+      plugins: [...bundleAnalyzer],
```
williamlark: nit: falsey plugins will be ignored, so you could do it like this too


```suggestion
  const bundleAnalyzer =
    isProductionBuild || !isIntegration
      ? bundleAnalyzerPlugin({ name: 'client' })
      : null;

return createSkuViteConfig(
{
      plugins: [bundleAnalyzer],
```

packages/sku/src/services/vite/helpers/config/createConfig.ts
```diff
@@ -36,8 +37,17 @@ export const createServerBuildConfig = (skuContext: SkuContext) => {
 
 export const createClientBuildConfig = (skuContext: SkuContext) => {
   const outDir = createOutDir(skuContext.paths.target);
+
+  const isProductionBuild = process.env.NODE_ENV === 'production';
+  const isIntegration = process.env.SKU_INTEGRATION === 'true';
```
williamlark: Curious, What is `SKU_INTEGRATION` used for?

adenj: That was a placeholder I popped in to mimic the webpack bundle analyzer setup while I was testing and forgot to remove 🤦  It isn't needed here

tests/browser/report-generation.test.ts
```diff
@@ -0,0 +1,62 @@
...
+} from '@sku-private/testing-library';
+import fs from 'node:fs';
+import path from 'node:path';
+
+const timeout = 50_000;
+
+vi.setConfig({
+  hookTimeout: timeout + 1000,
+  testTimeout: timeout + 1000,
+});
```
williamlark: Does this test fail with the default timeouts? You only need to set this if you need the test to override the defaults

tests/browser/report-generation.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      };
+
+      beforeAll(async () => {
+        const build = await sku('build', args[bundler]);
+        globalExpect(
+          await build.findByText('Sku build complete', {}, { timeout }),
+        ).toBeInTheConsole();
+
+        return cleanup;
+      });
```
williamlark: You may want to delete the report folder beforeAll to make sure it's not asserting on previously generated files

tests/browser/report-generation.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      it('should generate bundle analysis report', async ({ expect, task }) => {
+        skipCleanup(task.id);
+        const reportPath = fixturePath('report');
+        const clientReportPath = path.join(reportPath, 'client.html');
+
+        expect(fs.existsSync(reportPath)).toBe(true);
+        expect(fs.existsSync(clientReportPath)).toBe(true);
```
williamlark: Checking that a file exists inside a folder inherently checks that the folder exists. _Probably_ don't need the folder check (unless it gives a better error when it fails)

```suggestion
        const reportHtml = fixturePath('report/client.html');
        
        expect(fs.existsSync(reportHtml)).toBe(true);
```

tests/browser/report-generation.test.ts
```diff
@@ -0,0 +1,62 @@
...
+        // Verify the report is a valid HTML file with reasonable content
+        const reportContent = fs.readFileSync(clientReportPath, 'utf-8');
+        const stats = fs.statSync(clientReportPath);
+
+        expect(reportContent).toContain('<!DOCTYPE html>');
+        expect(reportContent).toContain('<html');
+        expect(reportContent).toContain('</html>');
+        expect(reportContent).toMatch(
+          /(bundle|Bundle|chart|Chart|visual|Visual)/i,
+        );
```
williamlark: Maybe these are overkill? We already check for a html file existing so I think we should trust the content that's generated since it's a plugin doing the generating. "does the report/client.html exist?" and "is it a big file?" feels like enough to gauge it's working correctly 🤷

tests/browser/report-generation.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        const build = await sku('build', args[bundler]);
+        globalExpect(
+          await build.findByText('Sku build complete', {}, { timeout }),
+        ).toBeInTheConsole();
+
+        return cleanup;
+      });
+
+      it('should generate bundle analysis report', async ({ expect, task }) => {
+        skipCleanup(task.id);
```
williamlark: Sorry, missed this. You only need to skip cleanup if you're keeping a server running. skipping cleanup keeps the process running, which we don't care about for build.

```suggestion
      });

      it('should generate bundle analysis report', async ({ expect, task }) => {
```

tests/browser/report-generation.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        const reportHtml = fixturePath('report/client.html');
+        expect(fs.existsSync(reportHtml)).toBe(true);
+
+        // Should be a substantial file (bundle analysis reports are typically large)
+        const stats = fs.statSync(reportHtml);
+        console.log({ stats });
```
williamlark:
```suggestion
```

tests/browser/report-generation.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        vite: ['--config', 'sku.config.vite.ts', '--experimental-bundler'],
+        webpack: [],
+      };
+
+      beforeAll(async () => {
+        // Clean up any existing report folder
+        const reportDir = fixturePath('report');
+        if (fs.existsSync(reportDir)) {
+          fs.rmSync(reportDir, { recursive: true, force: true });
+        }
```
williamlark: using `await fs.rm(...)` won't fail if the file doesn't exits so you could skip checking existence 

```suggestion
        import fs from 'node:fs/promises';
        // ...
        await fs.rm(reportDir, { recursive: true, force: true });
```

williamlark (approved):
Sorry, few small things but approving 💯

PR #1346, opened 2025-08-15
Vite v7

williamlark:
Bumping deps to Vite v7 and moving some of the shared deps to the pnpm-workspace catalog. 

Some minor test changes as well.

### DIFF
--- a/tests/browser/storybook-config.test.ts
+++ b/tests/browser/storybook-config.test.ts
@@ -25,7 +25,7 @@ import {
 const storybookStartedRegex =
   /Storybook \d+\.\d+\.\d+ for react-webpack5 started/;
 
-const timeout = 100_000;
+const timeout = 150_000;
 
 vi.setConfig({
   hookTimeout: timeout + 1000,


pnpm-workspace.yaml
```diff
@@ -8,8 +8,6 @@ packages:
 configDependencies:
   pnpm-plugin-sku: 0.0.1+sha512-9tebopr+ruP0nprOotkf8IeMASkXzYdsZuJgOHXhULQLDmg0MkDsPw/zX8P0LcyrImjqQQSXfRvoaWYNXgtrkw==
 
-hoistWorkspacePackages: false
```
williamlark, line 11: I wasn't sure why we turned this off, and now that the workspace deps are being injected I don't think this is doing anything useful 🤷

askoufis: Without it I think it was masking a missing dependency issue that was caused by a workspace package being hoisted (`@sku-lib/vite` I think). As long as `injectWorkspacePackages` still results in that package not being hoisted, I'm happy.

askoufis (approved):

PR #1352, opened 2025-08-19
`sku configure`: Don't emit `.npmrc` warning on CI

askoufis:
`.npmrc` files may be generated on CI, causing this warning to be displayed when really it's not valid. The rest of the PNPM config warnings are fine to emit on CI.

### DIFF
--- a/packages/sku/src/services/packageManager/pnpmConfig.test.ts
+++ b/packages/sku/src/services/packageManager/pnpmConfig.test.ts
@@ -6,12 +6,11 @@ import { validatePnpmConfig } from './pnpmConfig.js';
 
 describe('validatePnpmConfig', () => {
   describe.for([true, false])('When isCI: $0', (mockedIsCI) => {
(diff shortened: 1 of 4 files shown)


packages/sku/src/services/packageManager/pnpmConfig.test.ts
```diff
@@ -1,10 +1,16 @@
...
+import * as ci from '../../utils/isCI.js';
 
 import { validatePnpmConfig } from './pnpmConfig.js';
 
 describe('validatePnpmConfig', () => {
-  describe.for`
+  describe('when not in CI', () => {
+    beforeEach(() => {
+      vi.spyOn(ci, 'default', 'get').mockReturnValue(false);
+    });
```
askoufis: This is the recommended way of mocking exported values https://main.vitest.dev/guide/mocking.html#mock-exported-variables.

PR #1353, opened 2025-08-19
`start (vite)`: Fix sporadic hydration issues

williamlark:
The Vanilla Extract plugin provides a `inlineCssInDev` that extract styles and places them in the head of the document. This works most of the time, but due to the lifecycle of vite, it won't always be able to fetch all of the styles before the first load. This causes sporadic hydration errors.

I found an alternative solution using [hi-ogawa/vite-plugins/ssr-css](https://github.com/hi-ogawa/vite-plugins/blob/main/packages/ssr-css/README.md). This PR forks that code and adds a sort to make sure Braid reset styles are included before everything else.

I've also removed Braid from the flakey tests array since the race condition seemed to have been the cause of the flakeyness. If we see it flake again though we can add it back to the flakey list.


packages/sku/src/services/vite/plugins/ssrCss/collect.ts
```diff
@@ -0,0 +1,75 @@
...
+// https://github.com/remix-run/remix/blob/1a8a5216106bd8c3073cc3e5e5399a32c981db74/packages/remix-dev/vite/styles.ts
+// https://github.com/vikejs/vike/blob/f9a91f3c47cab9c2871526ef714cc0f87a41fda0/vike/node/runtime/renderPage/getPageAssets/retrieveAssetsDev.ts
+// https://github.com/sveltejs/kit/blob/998edb26d431e4ee4d3b4dc792a86960a85a5b45/packages/kit/src/exports/vite/dev/index.js#L186-L219
+// https://github.com/withastro/astro/blob/6aaeec5034cabf6a83e1949ec1ca8f50e7978cc1/packages/astro/src/vite-plugin-astro-server/css.ts
+
+/**
+ * We need to make sure that the Braid reset styles are loaded first.
+ * This will catch any `reset` files, which should generally be the first styles to load anyway.
+ */
+function sortResetFirst(a: string, b: string) {
```
askoufis: While this addresses the specific case of Braid's reset, there's likely other order-dependent VE styles within braid and/or applications that this won't fix.

Looking at the implementation of `collectStyleUrls` though, I don't see why styles shouldn't already be ordered correctly. I'll dig into this.

williamlark: Perhaps, but I couldn't recreate anymore issues from testing. It's also only an issue on initial load, and only in dev, so I didn't think it was worth optimising for unless we catch it or it is reported.

askoufis: Ok, so it's _correct_, but only as far as the traversal implementation is concerned. The issue is that because the traversal uses `Promise.all`, not only is the order potentially inconsistent, but it more closely mimics a breadth-first search, so naturally it will hit the shallowest entries first, which is typically local styles such as `App.css.ts`. If it was depth first, it would hit braid's `reset.css.ts` first and all would be good in the world.

This _could_ be worth discussing upstream, but at least we can easily convert it to depth-first ourselves. I can't think of a reason to not mimic import order (depth first) during this traversal, other than that it's maybe a bit slow?

```ts
async function collectStyleUrls(
  server: ViteDevServer,
  entries: string[],
): Promise<string[]> {
  const visited = new Set<string>();

  async function traverse(url: string) {
    const [, id] = await server.moduleGraph.resolveUrl(url);
    if (visited.has(id)) {
      return;
    }
    visited.add(id);
    const mod = server.moduleGraph.getModuleById(id);
    if (!mod) {
      return;
    }
    // Sequential traversal for depth-first behavior
    for (const childMod of mod.importedModules) {
      await traverse(childMod.url);
    }
  }

  // ensure vite's import analysis is ready _only_ for top entries to not go too aggresive
  await Promise.all(entries.map((e) => server.transformRequest(e)));

  // traverse sequentially for depth-first
  for (const url of entries) {
    await traverse(url);
  }

  // filter
  return [...visited].filter((url) => url.match(CSS_LANGS_RE));
}
```

askoufis:
> > I've also removed Braid from the flakey tests
> 
> Well, [I guess I won't then...](https://github.com/seek-oss/sku/actions/runs/17063778804/job/48376083425?pr=1353)

The flakiness _might_ still be caused by some module invalidation stuff that I haven't had time to look into.

PR #1354, opened 2025-08-20
docs: Rewrite Vite migration docs

askoufis:
Also:
- Added missing documentation for `__UNSAFE_EXPERIMENTAL__testRunner`
- Adjusted the `devServerMiddleware` API for Vite so we only pass in the Connect server instance rather than the entire Vite dev server


docs/docs/vite.md
```diff
@@ -0,0 +1,251 @@
...
+[Vite]: https://vite.dev/
+[`#sku-support`]: https://seek.enterprise.slack.com/archives/CDL5VP5NU
+
+## Limitations
+
+Vite support is currently only available for [static applications (SSG)][SSG].
+Support for [server-rendered applications (SSR)][SSR] is planned for the future.
+
+Practically, this means that currently only [`sku start`] and [`sku build`] are supported with Vite.
+[`sku serve`] is also available as it is bundler agnostic.
```
williamlark: I don't think we need to mention SSR stuff will be in the future. 

```suggestion
Vite support is currently only available for [static applications (SSG)][SSG]. 
This means that only [`sku start`], [`sku build`] and [`sku serve`] are supported.
```

docs/docs/vite.md
```diff
@@ -0,0 +1,251 @@
...
+  // or use a path
+  server.use('/api', (req, res, next) => {
+    // your middleware logic
+    next();
+  });
+}
+```
+
+?> Currently only JavaScript middleware is supported.
+Support for TypeScript middleware will be added in the near future.
```
williamlark: I don't want to make promises in our docs haha Great idea though. worth a card for supporting ts middleware if we don't have one already.

```suggestion
?> Currently only JavaScript middleware is supported.
```

docs/docs/vite.md
williamlark: It be great to add some more of the limitations that people will encounter in this document upfront here. Even a dot point is enough here before going deeper in the specifics later. For example:

```
- Vite applications must be written in ESM. (See #Migrating to ESM)
- Vite applications will most likely need to use the experimental Vitest test runner. (See #Migrating to Vitest)
```

docs/docs/vite.md
```diff
@@ -0,0 +1,251 @@
...
+[libraries]: ./docs/libraries.md
+
+## Prerequisites
+
+!> Before making any changes to your application, please ensure you have read this document in its entirety.
+
+### Migrating to ESM
+
+Vite applications must be written in [ESM (ECMAScript Modules)][ESM].
+**_Before swapping to Vite_**, you must ensure the Node.js package containing your application declares itself as an ES module.
```
williamlark:
```suggestion
**_Before swapping to Vite_**, ensure the your application declares itself as an ES module.
```

docs/docs/vite.md
```diff
@@ -0,0 +1,251 @@
...
+
+However, Vite can resolve these imports for you, so it is only necessary to include file extensions in import paths within non-application code.
+
+?> By default, `sku` configures `allowImportingTsExtensions: true` in your `tsconfig.json` file.
+In situations where an explicit file extension is required, such as in a Node.js script, this setting allows you to import TypeScript files with a `.ts` extension instead of a `.js` extension, which can be a source of confusion for those new to ESM codebases.
+
+[ESM]: https://nodejs.org/api/esm.html
+[CommonJS]: https://nodejs.org/api/modules.html
+[explicit file extensions]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#:~:text=you%20cannot%20omit%20the%20file%20extension%20or%20the%20index.js%20file%20name.%20
+
```
williamlark, line 118: Is it worth mentioning the `__UNSAFE_EXPERIMENTAL__cjsInteropDependencies` here? People may run in to issues with cjs interop issues for their deps and not know how to fix it (or think about looking for  ESM alternatives)

docs/docs/vite.md
```diff
@@ -0,0 +1,251 @@
...
+[Vitest] is a testing framework that supports ESM out-of-the-box, integrates with the Vite ecosystem and has a similar API to Jest.
+These features make it a great replacement for Jest in `sku` applications, especially given [Jest's current limitations with ESM].
+
+?> Due to these limitations, it's likely that you'll need to migrate to Vitest at the same time as migrating to ESM.
+
+Migrating to Vitest should be fairly straightforward.
+Before making any changes, it is highly recommend to read [the Vitest documentation](https://vitest.dev/guide/) to familiarise yourself with the API and features, as well as to better understand the differences between Jest and Vitest.
+
+To configure `sku` to use Vitest instead of Jest when running `sku test`, configure `__UNSAFE_EXPERIMENTAL__testRunner` in your `sku` config:
```
williamlark: They will also need to install the `@sku-lib/vitest` dependancy

docs/docs/vite.md
askoufis: Yeah good point, would be nice to front load that info, though IMO the `Limitations` section doesn't feel like the right place. I'm gonna put them at the top of `Prerequisites` for now.

docs/docs/vite.md
askoufis: Yeah good point 😅. I think we've tried to tackle this before with `esbuild-register` but ran into issues. When writing this I was thinking that `jiti` would make it much easier to achieve. I'll add a card for it.

docs/docs/vite.md
askoufis: I knew I was forgetting something.

williamlark (approved):

PR #1355, opened 2025-08-20
`docs`: Fix incorrect links in CLI docs

askoufis:
Identified in [slack](https://seekchat.slack.com/archives/CDL5VP5NU/p1755670303584389).

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -186,7 +186,7 @@ sku configure
 Translation-specific commands.
 These commands are only useful if your application is configured to be a [multi-language application].
 
-[multi-language application]: ./docs/multi-language-applications.md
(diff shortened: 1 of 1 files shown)


docs/docs/cli.md
```diff
@@ -228,6 +226,8 @@ See the documentation on [phrase-specific features] for more information.
 sku translations pull
 ```
 
+[phrase-specific features]: ./docs/multi-language.md#phrase-specific-features
+
```
williamlark, line 230: Why did this move up here instead of staying at the bottom of the page?

askoufis: It was both deduplicated and moved closer to where it was used. It's not actually used in the last section, but rather the 2 sections above, so I moved it to the bottom of the last section it's used in.

PR #1356, opened 2025-08-20
`start(vite)`: Fix sporadic hydration issues (again)

williamlark:
This PR fixes deps being missing in the `noExternal` array. This was due to `analyseDependency` throwing errors and bailing early depending on the order the dependency analysis promises were resolved. We now catch and log the errors to make sure it doesn't bail early.

The fix that users will see is that there should now no longer be any hydration errors when running `sku start` with the vite bundler. I've removed braid from the flakey tests array since that _should_ now work consistently.


packages/sku/src/services/vite/plugins/setSsrNoExternal.ts
```diff
@@ -8,23 +8,22 @@ import type { SkuContext } from '../../../context/createSkuContext.js';
...
-    config: async (config) => {
+    async config() {
       const depGraph = await extractDependencyGraph(process.cwd());
+
       const ssrExternals = getSsrExternalsForCompiledDependency(
         '@vanilla-extract/css',
         depGraph,
       );
 
-      const mergedConfig = mergeConfig(config, {
```
askoufis, line 18: Nice

PR #1358, opened 2025-08-21
Flatten file structure, co-locate some config files

askoufis:
Trying to flatten `sku`'s folder structure a bit as it's a bit too nested. Various tooling config files (both public and private) are scattered throughout the repo. This PR consolidates some of them into `src/config`, and renames `index.ts` files to `${tool}Config.ts`. Also thought about just `${tool}.ts` since the file is already in a directory named `config`. Happy to swap if we think that's better.


williamlark (approved):
I have some differing opinions on file naming but nothing I feel strongly enough to block this PR. Good improvement 👍

adenj (approved):

williamlark (approved):

PR #1359, opened 2025-08-21
Add `fixture` script for running fixture scripts

askoufis:
Cobbled together a little script that lists all fixtures, and then lists all scripts for the selected fixture, and then runs that script. Saves having to `cd` directly into directories and memorize script names or needing to use `pnpm --filter`.

https://github.com/user-attachments/assets/33624111-d217-4b6b-8a34-2c5652e6b496

`@clack/prompts` is neat. Will followup replacing our `prompts` dev dep with it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,11 +78,17 @@ pnpm run test:update
 > [!TIP]
 > The test suite needs to pass for your changes to be accepted, so it's worth running this locally during development and right before committing.
 
-You can also run any `sku` CLI command against any of the app fixtures.
+You can also run commands directly against any of the app fixtures.
 This can be a faster way to iterate on a feature than rather than running the test suite every time you make a change.
 
-1. `cd` into the fixture you want to test. E.g. `cd fixtures/styling`.
-1. Run your sku command. E.g. `pnpm run sku build`.
+There are 2 ways to run commands against a fixture:
+
+1. Run `pnpm fixture`, select the fixture you want and the `package.json` script to run
+2. `cd` into the fixture and run a command directly. E.g. `cd fixtures/styling && pnpm sku start`.
+
+> [!TIP]
+> `pnpm fixture` lists scripts in a fixture's `package.json` file.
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -78,11 +78,17 @@ pnpm run test:update
...
-You can also run any `sku` CLI command against any of the app fixtures.
+You can also run commands directly against any of the app fixtures.
 This can be a faster way to iterate on a feature than rather than running the test suite every time you make a change.
 
-1. `cd` into the fixture you want to test. E.g. `cd fixtures/styling`.
-1. Run your sku command. E.g. `pnpm run sku build`.
+There are 2 ways to run commands against a fixture:
+
+1. Run `pnpm fixture`, select the fixture you want and the `package.json` script to run
+2. `cd` into the fixture and run a command directly. E.g. `cd fixtures/styling && pnpm sku start`.
```
williamlark: Would be great to add the pnpm filter way of doing it too in case people don't know about it (I use it most of the time instead of changing directory)

```
3. pnpm --filter ...some example
```

williamlark (approved):
This is great! I've been wanting to add something like this for some time. Cool library!

This could probably be extended to support fuzzy matching of fixture commands too, right? This is great for one-off commands, but I use git history to quickly rerun commands when testing something a lot and having to type out the entire pnpm filter the first time is a little annoying. Would be great if it was able to fuzy match the fixture names and run the command directly on a match, or prompt with matching fixtures if there are multiple. (Not needed for this PR. I'm keen to play with this).

E.g., 
```bash
$ pnpm fixture braid start:vite
> Found @sku-fixtures/braid-design-system
> Running `start:vite` command
...

// or
$ pnpm fixture routes build
> Multiple fixtures found. Please select which one to use.
> {selectable list}
> Running command `build` on fixture {selected option}
```

CONTRIBUTING.md
askoufis: Yep, can definitely add this.

PR #1360, opened 2025-08-22
Fuzzy fixture search

williamlark:
Building on #1359, added the ability to run fixture commands with a fuzzy directory name search. This can replace `pnpm --filter=...` commands for fixtures and reduces having to change directories to run scripts. This also allows you to quickly rerun the command with terminal history instead of having to select from the fixture list every time.

All arguments are optional. The CLI will try to fuzzy match a fixture directory if possible, otherwise provides a selectable list. If a script is provided it will try running it, otherwise will display a selectable list.

* Renamed the package since it has an isolated purpose now.
* Turned it into a commander app.
* Moved some of the reused functions into seperate files.

### Examples

```
pnpm --filter @sku-fixtures/braid-design-system start:vite

// can now be run as 

pnpm fixure braid start:vite
```

https://github.com/user-attachments/assets/41a10cd5-ed31-4d83-93c4-328028b2c79f


private/fixture-cli/src/selectFixture.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (input && fixtureMatches.length === 0) {
+    log.error(
+      `No fixtures matching '${styleText(['bold', 'red'], input)}' found`,
+    );
+    process.exit(1);
+  }
+
+  if (fixtureMatches.length === 1) {
+    log.info(
+      `Found matching fixtures ${styleText(['bold', 'blue'], fixtureMatches[0])}`,
```
askoufis:
```suggestion
      `Found matching fixture ${styleText(['bold', 'blue'], fixtureMatches[0])}`,
```

askoufis (approved):
Slick 👌

PR #1362, opened 2025-08-26
Improve `postinstall` script

askoufis:
Various changes:
- Improve error handling. Previously thrown errors would be caught by the surrounding `try/catch`, so _any_ error would log `package.json does not exist`, which isn't great. Now we simply log errors and exit the script.
- Run multiple promises concurrently
- Replace `chalk` with native `styleText`

### DIFF
--- a/packages/sku/scripts/postinstall.js
+++ b/packages/sku/scripts/postinstall.js
@@ -3,15 +3,26 @@
 import { readFile } from 'node:fs/promises';
 import { join } from 'node:path';
 import debug from 'debug';
-import chalk from 'chalk';
+import { styleText } from 'node:util';
+
+/** @param {string} text */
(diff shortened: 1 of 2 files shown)


packages/sku/scripts/postinstall.js
```diff
@@ -3,15 +3,26 @@
 import { readFile } from 'node:fs/promises';
 import { join } from 'node:path';
 import debug from 'debug';
-import chalk from 'chalk';
+import { styleText } from 'node:util';
```
askoufis, line 6: Aiming to slowly replace `chalk` usage with `stlyeText` over time. It's technically an unstable API given our current node target, but it's stable in current LTS versions and there aren't any API changes that affect our usage of it.

PR #1365, opened 2025-08-29
`create`: Add `create` package

adenj:
Introduces a new `create` package for scaffolding new sku projects replacing the previous `sku init` functionality.

**Changes**
- **New CLI package**: `@sku/create` with `create-sku` binary for scaffolding projects
- **Template support**: Both `webpack` and `vite` template options accessible via interactive CLI or `--template [vite|webpack]` flag


pnpm-workspace.yaml
```diff
@@ -25,30 +47,5 @@ preferWorkspacePackages: true
 publicHoistPattern:
   - '@babel/*'
 
-# This is needed so that the React 18 fixture doesn't clobber React 19 in every other fixture. We can remove this once we've dropped React 18 support.
```
askoufis, line 28: Did `pnpm` update this config? I think it tends to delete comments which is annoying. Looks like it didn't actually updated anything so it would be great revert this change.

packages/create/templates/base/src/render.tsx
```diff
@@ -0,0 +1,44 @@
...
+import type { Render } from 'sku';
+
+import { App } from './App/App';
+import type { ClientContext } from './types';
+
+interface RenderContext {
+  appHtml: string;
+}
+
+const skuRender: Render<RenderContext> = {
```
askoufis, line 11: We should use this opportunity to uplift our template a bit. IMO this should be an inline default export with `satisfies`:

```suggestion
export default {
```

packages/create/templates/base/src/render.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+        ${headTags}
+      </head>
+      <body>
+        <div id="app">${app.appHtml}</div>
+        ${bodyTags}
+      </body>
+    </html>
+  `,
+};
```
askoufis, line 42:
```suggestion
} satisfies Render<RenderContext>;
```

packages/create/templates/base/src/render.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export default skuRender;
```
askoufis, line 44:
```suggestion
```

packages/sku/src/program/commands/init/init.action.ts
```diff
@@ -1,273 +1,14 @@
...
-
-  const logLevel = verbose ? 'verbose' : 'regular';
-
-  await install({ deps, logLevel });
-  await install({
-    deps: devDeps,
-    type: 'dev',
-    logLevel,
-    exact: false,
+  return createProject(projectName, {
```
askoufis, line 10: I think this should be `await`ed. We should set up a lint for that.

packages/sku/src/program/commands/init/init.action.ts
```diff
@@ -1,273 +1,14 @@
...
-  }
-
-  await mkdir(projectName, { recursive: true });
-
-  if (!isEmptyDir(initDir)) {
-    console.log(`The directory ${chalk.green(projectName)} is not empty.`);
-    process.exit(1);
-  }
+  // Delegate to the create package for backward compatibility
+  const { createProject } = await import('@sku-lib/create');
```
askoufis, line 8: Does this need to be a dynamic import or can it be statically imported at the top of the file? IIRC `.action.ts` files are already dynamically imported to avoid importing unnecessary deps, so another layer of dynamic import shouldn't be necessary.

packages/create/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "devDependencies": {
+    "@types/node": "catalog:",
+    "@types/prompts": "^2.4.9",
+    "@types/cross-spawn": "^6.0.3",
+    "@types/debug": "^4.1.12",
+    "@types/semver": "^7.5.8",
+    "typescript": "~5.8.0",
+    "vitest": "catalog:",
+    "@sku-lib/vitest": "workspace:*",
+    "fs-fixture": "^1.1.0"
```
askoufis, line 60: This is an old version. The rest of the repo is on `^2.6.0`.

packages/create/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "type": "module",
+  "bin": {
+    "create-sku": "./dist/cli.js"
+  },
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    }
+  },
```
askoufis, line 14: Should add a `package.json` entry here like our other packages:
```suggestion
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "default": "./dist/index.js"
    },
    "./package.json": "./package.json"
  },
```
This also enables you to easily import package info in the CLI entrypoint.

packages/create/src/cli.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { Command } from 'commander';
+import { createProject } from './actions/createProject.js';
+
+const program = new Command();
+
+program
+  .name('@sku-lib/create')
+  .description('Create new sku projects')
+  .version('1.0.0')
```
askoufis, line 11: If we add a `package.json` export entry we can then import it to get package info. E.g.:

```suggestion
import { createProject } from './actions/createProject.js';
import packageJson from 'sku/package.json' with { type: 'json' };

const { name, description, version } = packageJson;
const program = new Command();

program
  .name(name)
  .description(description)
  .version(version)
```

packages/create/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@sku-lib/create",
+  "version": "1.0.0",
```
askoufis, line 3: I'd rather this start at `0.0.0` (to be released as `0.1.0`) so we can more safely iterate on its API if necessary.

packages/create/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "../../tsconfig.json",
```
askoufis, line 2: This should extend from `@tsconfig/node20/tsconfig.json` like the rest of the packages.
```suggestion
  "extends": "@tsconfig/node20/tsconfig.json",
```

packages/create/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+  "compilerOptions": {
+    "outDir": "dist",
+    "rootDir": "src",
+    "module": "ESNext",
+    "moduleResolution": "Node",
+    "allowImportingTsExtensions": false,
+    "noEmit": false,
+    "declaration": true,
+    "declarationMap": true,
+    "sourceMap": true
```
askoufis, line 12: We usually don't set these as they aren't that useful for our use case.

packages/create/tsconfig.build.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.json",
```
askoufis, line 2: I think this needs a `compilerOptions.outDir` otherwise all the built files will end up alongside the source files.

packages/create/src/utils/prettier.ts
```diff
@@ -0,0 +1,68 @@
...
+  const prettierIgnorePath = join(projectPath, '.prettierignore');
+  const ignoreExists = await fileExists(prettierIgnorePath);
+
+  const prettierArgs = [
+    'prettier',
+    '--write',
+    '--cache',
+    ...(ignoreExists ? ['--ignore-path', '.prettierignore'] : []),
+    '.',
+  ];
```
askoufis, line 27: I don't think this approach will work in all situations. It happens to work in the sku repo because we have `prettier` (and `eslint`) as root-level dependencies, but in a new app (especially in a PNPM) app, `prettier` won't be available.

packages/create/src/templates/templateProcessor.ts
```diff
@@ -0,0 +1,133 @@
...
+  lintScript: string;
+  formatScript: string;
+  buildScript: string;
+}
+
+function generateTemplateData(appName: string): TemplateData {
+  return {
+    appName,
+    gettingStartedDocs: dedent`
+      First of all, make sure you've installed [pnpm](${getPackageManagerInstallPage()}).
```
askoufis, line 58:
```suggestion
      First of all, make sure you've installed [${packageManager}](${getPackageManagerInstallPage()}).
```

packages/create/src/utils/isEmptyDir.test.ts
```diff
@@ -0,0 +1,57 @@
+import { describe, it, vi } from 'vitest';
+
+const mockReaddirSync = vi.fn();
+
+vi.mock('node:fs', () => ({
+  readdirSync: mockReaddirSync,
+}));
+
+const { isEmptyDir } = await import('./isEmptyDir.js');
```
askoufis, line 9: This shouldn't need to be a dynamic import right?

askoufis (commented):
I started giving some granular feedback, but stopped before getting through all the changes because not only is this hard to review but it needs some rethinking IMO. Some broad issues I have are:
- Utilities from sku being copied and/or re-implemented in the new package. E.g. `packageManager` is almost the same, yet uses more switch cases rather than being stricter on type-safety instead. `banner` is quite different to the existing implementation, `isEmptyDir` to a lesser extent.
- Re-implementing `prettier` and `eslint` execution. I understand that currently we format initialized code using `sku`, so I can see why you've implemented this, however this could've been circumvented entirely but just running `sku format` in the project after initialization. Additionally, the current implementation won't work in a real repo because `prettier` and `eslint` aren't available as top-level dependencies. It works currently because the sku monorepo has these top-level dependencies.

This is why I would've preferred the PR to be smaller and implemented incrementally, because these issues around sharing and refactoring utilities would've been surfaced earlier, _and_ it would've made this much easier to review.

I'm going to be a pain and ask that this be split out into at least the following PRs:
- Create new _private_ package with just a CLI entrypoint that contains a stubbed action that doesn't do anything
- Create a new package for shared utilities
- Identify and move shared utilities to the new package and add a changeset for a `sku` release for this
- Add tests for the utilities if necessary
- Implement the functionality for the new CLI

PR #1366, opened 2025-08-29
`create`: Add minimal sku-create CLI package

adenj:
Adds minimal `@sku/create` package with working CLI binary and interactive template selection


packages/create/src/cli.ts
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env node
+
+import { Command } from 'commander';
+import inquirer from 'inquirer';
```
askoufis: Can we stick to `prompts` for now given we already use it elsewhere?

packages/create/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@sku/create",
```
askoufis:
```suggestion
  "name": "@sku-lib/create",
```

packages/create/package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "files": [
+    "dist"
+  ],
+  "engines": {
+    "node": ">=18"
+  },
+  "scripts": {
+    "build": "tsc --project tsconfig.build.json",
+    "test": "vitest"
```
askoufis: Tests will be handles by the top-level `test` script so this shouldn't be necessary.

packages/create/src/cli.ts
```diff
@@ -0,0 +1,35 @@
...
+        await inquirer.prompt([
+          {
+            type: 'list',
+            name: 'template',
+            message: 'Which template would you like to use?',
+            choices: ['webpack', 'vite'],
+            default: 'webpack',
+          },
+        ])
+      ).template;
```
askoufis: This is a bit hard to read. IMO it would read better with an `if` statement:

```suggestion
    let selectedTemplate = options.template
    if (!template) {
      const { template } = await inquirer.prompt([
        {
          type: 'list',
          name: 'template',
          message: 'Which template would you like to use?',
          choices: ['webpack', 'vite'],
          default: 'webpack',
        },
      ]);
      selectedTemplate = template;
    };
```

packages/create/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "scripts": {
+    "build": "tsc --project tsconfig.build.json"
+  },
+  "dependencies": {
+    "commander": "^12.1.0",
+    "prompts": "^2.4.2"
+  },
+  "devDependencies": {
+    "typescript": "^5.6.2",
+    "vitest": "^2.1.1",
```
askoufis: Shouldn't need this dep either.
```suggestion
```

packages/create/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "private": true,
+  "bin": {
+    "create-sku": "./dist/cli.js"
+  },
+  "files": [
+    "dist"
+  ],
+  "engines": {
+    "node": ">=18"
+  },
```
askoufis: This needs to align with `sku`'s config:
https://github.com/seek-oss/sku/blob/ac06eef4c8061d6b252339093a0d146c82844d4d/packages/sku/package.json#L36-L38

packages/create/src/cli.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { Command } from 'commander';
+import prompts from 'prompts';
+
+const program = new Command();
+
+program
+  .name('create-sku')
+  .description('Create a new sku project')
+  .version('0.0.0')
```
askoufis: Left this [feedback](https://github.com/seek-oss/sku/pull/1365#discussion_r2308942692) on the other PR, but if we add a `./package.json` export entry, you can import this stuff from `@sku-lib/create/package.json` instead of duplicating it.

packages/create/src/cli.ts
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
+
+import { Command } from 'commander';
+import prompts from 'prompts';
+
+import packageJson from '../package.json' with { type: 'json' };
```
askoufis: Adding a `package.json` export would allow this to be path agnostic (i.e. not relative), e.g. https://github.com/seek-oss/sku/blob/ac06eef4c8061d6b252339093a0d146c82844d4d/packages/sku/package.json#L34

```suggestion
import packageJson from '@sku-lib/create/package.json' with { type: 'json' };
```

askoufis (approved):

PR #1369, opened 2025-09-01
docs: Add alternatives to CJS interop config

askoufis:
`__UNSAFE_EXPERIMENTAL__cjsInteropDependencies` should ideally be a last resort, so I've added some potential solutions that should be tried _before_ reaching for that config.


williamlark (approved):
🧔‍♂️ 🪳 _last resort_ 🎵

PR #1370, opened 2025-09-01
`start`: Fix 404 messaging

williamlark:
When going to a non-existent route, sku would show a 404 messaging `Cannot GET {route_path}`. When adding Vite support, this message broke and instead sku would throw a `getLanguageFromRoute` error when encountering a 404. This PR fixes restores the 404 messaging.

### DIFF
--- a/.changeset/bitter-nails-carry.md
+++ b/.changeset/bitter-nails-carry.md
@@ -1,0 +1,5 @@
+---
+'sku': patch
+---
+
+`start`: Shows 404 messaging instead of an error when routes are empty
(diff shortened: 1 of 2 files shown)


.changeset/bitter-nails-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+`start`: Shows 404 messaging instead of an error when routes are empty
```
jahredhope: I know 404 is universally understood to mean Not Found (even by non engineers), but I personally prefer to use the status text. If not, perhaps both?

```suggestion
`start`: Shows Not Found messaging instead of an error when routes are empty
```

williamlark: I've added them both in 👌

askoufis (approved):

PR #1371, opened 2025-09-01
Make use of dev server as proxy for SSR

jahredhope:
Makes use of sku's dev server as an entrypoint for SSR content, allowing it to act as a single entrypoint similar to how most production-scale web applications would work.

Further information in [Release Notes](.changeset/itchy-taxes-cut.md) and [Docs](docs/docs/server-rendering.md).

With this new behaviour we no longer need to [pin webpack-dev-server](https://github.com/seek-oss/sku/pull/1293/files).

When making this change I considered either:
- Opt-in
- Opt-out
- Always-on

I've chosen always-on, as this:
- allows us to unpin webpack-dev-server
- avoids the situation where we create unused configuration options, or unintentional deviation
- simplifies the code considerably

If it becomes necessary to return to the previous behaviour we can ship that as a patch once a use-case is identified.
See example with opt-in/out behaviour: https://github.com/seek-oss/sku/tree/dev-server-as-opt-in

### DIFF
--- a/packages/sku/src/program/commands/start-ssr/webpack-start-ssr-handler.ts
+++ b/packages/sku/src/program/commands/start-ssr/webpack-start-ssr-handler.ts
@@ -174,16 +174,15 @@ export const webpackStartSsrHandler = async ({
         port: clientPort,
       },
     },
+    setupMiddlewares: (middlewares) => [
+      ...middlewares,
+      {
+        name: 'send-to-ssr',
+        middleware: proxy(serverHost),
+      },
+    ],
     setupExitSignals: true,
   };
-
-  devServerConfig.setupMiddlewares = (middlewares) => [
-    ...middlewares,
-    {
-      name: 'send-to-ssr',
-      middleware: proxy(serverHost),
-    },
-  ];
 
   if (httpsDevServer) {
     const pems = await getCertificate('.ssl', hosts);


packages/sku/src/program/commands/start-ssr/webpack-start-ssr-handler.ts
```diff
@@ -178,6 +177,14 @@ export const webpackStartSsrHandler = async ({
...
   };
 
+  devServerConfig.setupMiddlewares = (middlewares) => [
+    ...middlewares,
+    {
+      name: 'send-to-ssr',
+      middleware: proxy(serverHost),
+    },
+  ];
+
```
askoufis: Likely leftover from before making the config always-on, but can this be moved into the `devServerConfig` definition instead of mutating it?

jahredhope: It was yes, thanks

.changeset/itchy-taxes-cut.md
```diff
@@ -0,0 +1,19 @@
...
+SSR server now accessible via webpack dev server
+
+When running `sku start` users would previously access the SSR server directly.
+They would then load static assets from the webpack dev server on a separate host.
+
+`sku start` will now proxy requests from the webpack dev server to the consumer's app. Allowing the dev server to act as a single entrypoint.
+
+When starting the dev server consumers may notice some changes:
+
+- A new port used when opening the browser window
```
williamlark, line 14: Can you clarify why there is a change in port here? My wording here is bad but a suggestion:


```suggestion
- The SSR server will start on `serverPort`, the browser will open at `port`. This logic was previously reversed.
```

williamlark (commented):
It might be worth updating the `port` docs description. It's current description is `The port the app is hosted on when running sku start`, however, now it is the port the app is accessible at in the browser, regardless of CSR or SSR (which is a nice change).

PR #1374, opened 2025-09-02
Flatten webpack plugins folder structure

askoufis:
Flattened the folder structure of our webpack plugins and replaced some default exports with named exports. ~I left `SkuWebpackPlugin`'s export alone as it's public API so didn't want to break it just yet~ Exposed a named `SkuWebpackPlugin` and deprecated the default export.

### DIFF
--- a/packages/sku/src/services/webpack/config/plugins/skuWebpackPlugin.ts
+++ b/packages/sku/src/services/webpack/config/plugins/skuWebpackPlugin.ts
@@ -19,7 +19,7 @@ import {
 import targets from '../../../../config/targets.json' with { type: 'json' };
 import { rootResolutionFileExtensions } from '../../../../config/fileResolutionExtensions.js';
 
-class SkuWebpackPlugin implements WebpackPluginInstance {
+export class SkuWebpackPlugin implements WebpackPluginInstance {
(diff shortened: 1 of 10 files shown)


packages/sku/src/services/webpack/config/plugins/skuWebpackPlugin.ts
```diff
@@ -10,13 +10,14 @@ import {
...
+import {
+  validateOptions,
   type SkuWebpackPluginOptions,
 } from './validateOptions.js';
-import targets from '../../../../../config/targets.json' with { type: 'json' };
-import { rootResolutionFileExtensions } from '../../../../../config/fileResolutionExtensions.js';
+import targets from '../../../../config/targets.json' with { type: 'json' };
+import { rootResolutionFileExtensions } from '../../../../config/fileResolutionExtensions.js';
 
 class SkuWebpackPlugin implements WebpackPluginInstance {
```
williamlark: Can this be named _and_ the default? Then people could start moving to named exports without breaking anything currently.

askoufis: Yeah good point

williamlark (approved):

PR #1381, opened 2025-09-03
`test (vitest)`: pathAlias now works in test files

williamlark:
TL;DR
* Added `pathAlias`, `compilePackages` and `cjsInteropDependencies` support for the vitest test runner
* Updated the vite migration docs a little to remove some alerts. Some of them were a little distracting so I've inlined and bolded a few + some other small additions. 

### Description

The `pathAlias` property would only work when setting the bundler to vite, however moving to vitest can be done independently to the vite migration. This PR makes path aliases work when using the vitest test runner without vite.

This PR also adds support for `compilePackages` and `__UNSAFE_EXPERIMENTAL__cjsInteropDependencies` to the vitest, which is needed to fix some errors related to cjs modules. This logic already lives in the vite config code so had to be copied over.

This change also highlights that the `@sku-lib/vitest` package doesn't have access to the whole `SkuContext` object since that is internal to sku. We may consider adding some of those internal types to a private package in the monorepo that can be bundled with the libraries (we don't want these internal types to be public). Perhaps in `@sku-lib/utils` or `@sku-lib/core` (doesn't exist yet), however since it was only two fields for now I've left that to fix another time.


docs/docs/vite.md
```diff
@@ -76,6 +76,16 @@ You may have non-application code such as Node.js scripts or configuration files
 If these files contain [CommonJS (CJS)][CommonJS] syntax and you do not wish to convert them to ESM, you can keep them as CommonJS by using the `.cjs` or `.cts` file extensions.
 However, **it is highly recommended to convert all code to ESM if possible**.
 
+#### Finding ESM code changes
+
+After changing the repo to `type: module`, the [eslint-cjs-to-esm](https://github.com/azu/eslint-cjs-to-esm) package can be used to check for potential ESM code changes needed.
```
askoufis:
```suggestion
After changing the repo to `type: module`, the [eslint-cjs-to-esm](https://github.com/azu/eslint-cjs-to-esm) package can be used to check for potential ESM code changes needed:
```

docs/docs/vite.md
```diff
@@ -120,13 +130,18 @@ In situations where an explicit file extension is required, such as in a Node.js
...
 These features make it a great replacement for Jest in `sku` applications, especially given [Jest's current limitations with ESM].
-
-?> Due to these limitations, it's likely that you'll need to migrate to Vitest at the same time as migrating to ESM.
+**Due to these limitations, it's likely that you'll need to migrate to Vitest at the same time as migrating to ESM.**
 
 Migrating to Vitest should be fairly straightforward.
 Before making any changes, it is highly recommend to read [the Vitest documentation](https://vitest.dev/guide/) to familiarise yourself with the API and features, as well as to better understand the differences between Jest and Vitest.
 
-To configure `sku` to use Vitest instead of Jest when running `sku test`, configure [`__UNSAFE_EXPERIMENTAL__testRunner`][test runner] in your `sku` config:
+To enable Vitest in `sku`, first install the required dependencies.
```
askoufis:
```suggestion
To enable Vitest in `sku`, first install the required dependencies:
```

docs/docs/vite.md
```diff
@@ -138,16 +153,20 @@ export default {
...
-Additionally, install the `@sku-lib/vitest` dependency:
-
-```bash
-pnpm add -D @sku-lib/vitest
-```
+`sku` will now invoke [the `vitest` CLI][Vitest CLI] instead of the `jest` CLI when running `sku test`.
 
-With this change, `sku` will now invoke [the `vitest` CLI][Vitest CLI] instead the `jest` CLI when running `sku test`.
+**A key difference** between the `jest` and `vitest` CLIs is that `vitest` defaults to watch mode when running tests locally.
+To run test without watch mode you can use `sku test --run`.
```
askoufis:
```suggestion
To run test without watch mode you can use `sku test --run`:
```

williamlark:
Asking for a re-review Adam since I've updated the snapshots and added a little bit of code to allow waiting for text in the DOM. For some reason the snapshot was sometimes taken too early before hydration, so I've added a property in the `getAppSnapshot` function that will wait for some text to appear before returning the html. This seems to make the braid snapshot a little more consistent. 

Some of the paths were updated too to what they are meant to be (the testing lib is meant to look in to node_modules of the fixture). Not sure why it wasn't being picked up in these tests before.

tests/browser/braid-design-system.test.ts
```diff
@@ -32,20 +31,9 @@ function getLocalUrl(site: string, port: number) {
   return `http://${host}:${port}`;
 }
 
-const { sku, exec, fixturePath } = scopeToFixture('braid-design-system');
-
-const removeCache = async () => {
```
williamlark, line 37: Removed these clean up scripts that were added before finding the braid css hydration issues. I don't think these are necessary now

PR #1265, opened 2025-05-26
separate vite loadable from the `sku` package.

DanDroryAu:
-- DRAFT --

PR #1382, opened 2025-09-04
Delete codemod unit tests, replace `sku/vite/loadable` references

askoufis:
While trying to make a new codemod I discovered that the existing tests weren't running because we weren't awaiting the test function. So that's fixed. There's probably other improvements to be made to these tests but I just want to get them working for the moment.

Also, since https://github.com/seek-oss/sku/pull/1265 `sku/vite/loadable` hasn't actually pointed to anything as the APIs were moved to `sku-vite/loadable` and now `@sku-lib/vite/loadable`. I've updated references to point to the new package name.

Ideally we'd set up a lint rule to enforce promises to be awaited, but adding typescript eslint rules to `sku`'s eslint config is tricky because they're configured with a custom parser in `eslint-config-seek`, which makes them hard (impossible?) to extend on our end.

**EDIT**: As discussed below, the codemod unit tests and e2e tests were effectively testing the same thing. The e2e tests are more valuable so the unit tests have been deleted.


private/plop/generators/add-sku-codemod/templates/test.hbs
```diff
@@ -1,3 +1,4 @@
+import { describe, it } from 'vitest';
```
askoufis: This is necessary now that we don't use globals.

packages/codemod/src/transform/worker.ts
```diff
@@ -2,17 +2,21 @@ import { workerData, parentPort } from 'node:worker_threads';
...
+import type { JobWorkerData } from './runner.js';
+import type { Transform } from '../utils/types.js';
 
-const { transformerPath, jobs, options } = workerData;
+const { transformerPath, jobs, options } = workerData as JobWorkerData;
 
-const fileTransformer = (await import(transformerPath)).transform;
+const { transform } = (await import(transformerPath)) as {
+  transform: Transform;
+};
```
askoufis, line 12: Casting isn't ideal but it's the simplest way to at least get some type information into the worker script.

packages/codemod/src/codemods/transform-vite-loadable/__tests__/transform-vite-loadable.test.ts
williamlark, line 1: Is this file testing the same stuff as `sku-codemods.test.ts`. It looks like it is, and that one passes since it's output correctly mentions `@sku-lib/vite/loadable`. Could we potentially delete this?

packages/codemod/src/utils/test-utils.ts
```diff
@@ -17,21 +18,25 @@ type TestProps = {
...
 
 const runInlineTest = async ({
   transform,
   input,
   expectedOutput,
 }: InlineTestProps) => {
-  const output = await transform(input.source);
+  if (!isValidTransform(transform)) {
+    throw new Error('Invalid transform function provided');
+  }
```
williamlark: Why is this check needed when the `transform` parameter is already typed? Could this fail at runtime?

packages/codemod/src/codemods/transform-vite-loadable/__tests__/transform-vite-loadable.test.ts
askoufis: `sku-codemods` is I guess more of an e2e test that tests the CLI behaviour, rather than testing the transform directly. I think it's still valuable as it's the only way to test dry run behaviour, for example.

packages/codemod/src/utils/test-utils.ts
askoufis: Good point, this should be done right after the `import` of the transform file rather than here where it already has a type.

packages/codemod/src/codemods/transform-vite-loadable/__tests__/transform-vite-loadable.test.ts
williamlark: It does still check for the transform behaviour with the fs-fixture files though. Isn't en e2e test enough here? The test util code to get these `transform-vite-loadable.test.ts` working is doing a lot of things that the e2e test doesn't need to care about. 

(to be clear, i was suggesting we could delete this `transform-vite-loadable.test.` file, not the `sku-codemod` one)

askoufis: Oh right. I think that's sensible. I'm running into flakiness with the codemod unit tests anyway for some reason. The CLI tests seem more robust so let's keep those.

askoufis: Deleted in https://github.com/seek-oss/sku/pull/1382/commits/ac745806e6b46ef34c21c9acaef2f1c78def8cd4

PR #1383, opened 2025-09-04
`create`: Add standalone create-sku CLI package

adenj:
Introduces a new @sku-lib/create package that provides a modern, standalone CLI for scaffolding new sku projects. This package replaces the functionality of`sku init` with a dedicated tool optimised for project creation.

**CLI package includes**
- **Interactive CLI:** Template selection with prompts when no `--template` flag provided
- **Project scaffolding**: Full project creation from directory setup to formatted output
- **Template system:** Support for webpack (default) and vite (experimental) templates
- **Package manager detection:** Automatically detects and uses `npm`, `yarn`, or `pnpm`

**Generated projects include**
- `sku.config.ts` with bundler-specific configuration
- `package.json` with appropriate dependencies and scripts
- TypeScript setup with proper configs (tsconfig.json)
- ESLint (eslint.config.mjs) and Prettier configuration
- `.gitignore`, `.prettierignore`, and `README.md` files
- Sample React app with pre-installed dependencies (`react`, `braid-design-system`, `vanilla-extract`)


packages/create/package.json
```diff
@@ -2,7 +2,6 @@
   "name": "@sku-lib/create",
   "version": "0.0.0",
   "description": "Create sku projects",
-  "private": true,
```
adenj, line 5: Removing to avoid changsets from erroring. See https://github.com/seek-oss/sku/pull/1378

packages/create/src/actions/createProject.ts
```diff
@@ -0,0 +1,83 @@
...
+  banner('info', `${projectName} created`, [
+    'Get started by running:',
+    nextSteps,
+  ]);
+};
+
+const resolveProjectPath = (projectName: string) => {
+  if (projectName === '.') {
+    return cwd();
+  }
```
adenj: Allows users to create a new sku project in their current directory

packages/create/templates/vite/sku.config.ts
```diff
@@ -0,0 +1,11 @@
+import type { SkuConfig } from 'sku';
+
+const skuConfig: SkuConfig = {
+  clientEntry: 'src/client.tsx',
+  renderEntry: 'src/render.tsx',
+  environments: ['development', 'production'],
+  publicPath: '/path/to/public/assets/', // <-- Required for sku build output
+  __UNSAFE_EXPERIMENTAL__bundler: 'vite',
```
askoufis, line 9: We should enable Vitest in this fixture to IMO. No reason to stick with Jest.

packages/create/src/cli.ts
```diff
@@ -33,9 +35,23 @@ program
       selectedTemplate = template;
     }
 
-    console.log('🚧 create-sku is not yet implemented');
-    console.log(`Project name: ${projectName}`);
-    console.log(`Template: ${selectedTemplate}`);
+    if (!selectedTemplate) {
+      console.log('❌ No template selected. Exiting.');
+      process.exit(1);
+    }
```
adenj, line 82: Users should only encounter this if they kill the template selection prompt

.changeset/yellow-lions-enjoy.md
```diff
@@ -0,0 +1,45 @@
+---
+'@sku-lib/create': major
```
askoufis: We should keep this 0-versioned for now by only publishing a minor so we can break in minors just in case we need to make quick breaking API changes.

```suggestion
'@sku-lib/create': minor
```

packages/create/templates/webpack/sku.config.ts
```diff
@@ -0,0 +1,10 @@
+import type { SkuConfig } from 'sku';
+
+const skuConfig: SkuConfig = {
+  clientEntry: 'src/client.tsx',
+  renderEntry: 'src/render.tsx',
+  environments: ['development', 'production'],
+  publicPath: '/path/to/public/assets/', // <-- Required for sku build output
+};
+
+export default skuConfig;
```
askoufis: Good opportunity to change both templates to use `satisfies`, which aligns with how we write sku configs in the docs:
```suggestion
export default {
  clientEntry: 'src/client.tsx',
  renderEntry: 'src/render.tsx',
  environments: ['development', 'production'],
  publicPath: '/path/to/public/assets/', // <-- Required for sku build output
} satisfies SkuConfig;
```

pnpm-workspace.yaml
```diff
@@ -25,30 +47,5 @@ preferWorkspacePackages: true
 publicHoistPattern:
   - '@babel/*'
 
-# This is needed so that the React 18 fixture doesn't clobber React 19 in every other fixture. We can remove this once we've dropped React 18 support.
```
askoufis: We should keep these comments as they have some valuable context.

packages/create/src/services/format.ts
```diff
@@ -0,0 +1,32 @@
...
+        console.log('⚠️ Formatting completed with warnings');
+      } else {
+        console.warn(`⚠️ Formatting failed with exit code ${code}`);
+      }
+      resolve();
+    });
+
+    child.on('error', (error) => {
+      console.warn(`⚠️ Failed to run format command: ${error.message}`);
+      resolve();
```
adenj, line 29: Since formatting isnt a critical part of this process, we resolve on warnings/errors so the project still gets generated and installation runs

packages/create/src/validation/packageName.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (!isValidPackageName) {
+    throw new Error(
+      `Could not create a project called "${appName}" because of npm naming restrictions. Please see https://docs.npmjs.com/cli/configuring-npm/package-json for package name rules.`,
+    );
+  }
+
+  if (reservedNames.indexOf(appName) >= 0) {
+    throw new Error(
+      `We cannot create a project called ${appName} because a dependency with the same name exists. Due to the way npm works, the following names are not allowed:\n\n${reservedNames.map((depName) => `  ${depName}`).join('\n')}\n\nPlease choose a different project name.`,
```
askoufis: > Due to the way npm works

I get that you didn't write this but this is sooo vague and unhelpful 😂. Might be a good opportunity to refine this message and align it to the previous one which is better:

```suggestion
      `Could not create a project called ${appName} because an NPM package with the same name already exists. The following names are not allowed:\n\n${reservedNames.map((depName) => `  ${depName}`).join('\n')}\n\nPlease choose a different project name.`,
```

packages/create/src/validation/packageName.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const reservedNames = [
+  'react',
+  'react-dom',
+  'sku',
+  'braid-design-system',
+].sort();
+
+export const validatePackageName = (targetPath: string): void => {
+  const appName = basename(targetPath);
```
askoufis: Deriving the `appName` from the `targetPath` seems outside the scope of this function. IMO it should take the `appName` as a parameter directly.

packages/create/src/actions/createProject.ts
askoufis: I like the idea, but reading the code it's a bit confusing that the `projectName` is the input to this function. Lines 25 and 26 almost feel like they're operating on each other's input.

E.g. `projectName` should be derived from `targetPath` - if it's `.` then `projectName` is the directory name.

tests/node/sku-create.test.ts
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * When snapshot testing YAML files, we don't care about the specific versions.
+ * This function strips version numbers from YAML content.
+ */
+function stripYamlVersions(yamlContent: string): string {
+  // Replace version patterns like "0.0.1+sha512-..." with "VERSION_IGNORED"
+  return yamlContent.replace(
+    /:\s*[\d.]+(?:\+sha\d+-[a-f0-9]+)?.*/g,
+    ': VERSION_IGNORED',
+  );
```
askoufis, line 141: This didn't seem to replace anything in the snapshot strangely, but it does in the `sku-init` test:
https://github.com/seek-oss/sku/blob/eec389b9f9ebc831d920f4667e9d95af73a9a0c3/tests/node/__snapshots__/sku-init.test.ts.snap#L342-L347

packages/create/src/services/install.ts
```diff
@@ -0,0 +1,54 @@
...
+      deps,
+      type,
+      logLevel: 'regular',
+      exact: false,
+    });
+    const [command, ...args] = addCommand.split(' ');
+
+    const child = spawn(command, args, {
+      cwd: projectPath,
+      stdio: ['inherit', 'inherit', 'inherit'],
```
askoufis: `stdio: 'inherit'` is shorthand for this. See https://nodejs.org/api/child_process.html#optionsstdio.
```suggestion
      stdio: 'inherit',
```

packages/create/src/generators/packageJson.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  beforeEach(async () => {
+    tempDir = join(tmpdir(), `create-test-${Date.now()}`);
+    await mkdir(tempDir, { recursive: true });
+  });
+
+  afterEach(async () => {
+    try {
+      await rm(tempDir, { recursive: true });
+    } catch {}
+  });
```
askoufis: We should use `fs-fixture` here like our other tests as it handles temp dir creation and cleanup for you https://github.com/privatenumber/fs-fixture?tab=readme-ov-file#using-keyword-explicit-resource-management

packages/create/src/generators/packageJson.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    const packageJsonContent = await readFile(
+      join(tempDir, 'package.json'),
+      'utf8',
+    );
+    const packageJson = JSON.parse(packageJsonContent);
+
+    expect(packageJson.name).toBe('my-vite-app');
+    expect(packageJson.type).toBe('module');
+    expect(packageJson.scripts.start).toContain('--experimental-bundler');
+    expect(packageJson.scripts.build).toContain('--experimental-bundler');
```
askoufis: IMO an inline snapshot would be better here (and in the test above) because it also catch accidental changes to template fields not checked here _and_ it means you don't need to cherry-pick specific fields, you can just test all of them.

packages/create/src/generators/templates.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    expect(result.projectName).toBe('my-app');
+    expect(result.appName).toBe('my-app');
+    expect(result.startScript).toBe('pnpm start');
+    expect(result.buildScript).toBe('pnpm build');
+    expect(result.testScript).toBe('pnpm test');
+    expect(result.formatScript).toBe('pnpm format');
+    expect(result.lintScript).toBe('pnpm lint');
+    expect(result.gettingStartedDocs).toContain('pnpm install');
+    expect(result.gettingStartedDocs).toContain('https://pnpm.io/installation');
```
askoufis: Again, inline snapshots are simpler and better here.

packages/create/src/actions/createProject.ts
```diff
@@ -0,0 +1,83 @@
...
+  createProjectDirectory(targetPath);
+
+  await generatePackageJson(targetPath, { projectName, template });
+  await generateTemplateFiles(targetPath, { projectName, template });
+  await installDependencies(targetPath);
+  await formatProject(targetPath);
+
+  const nextSteps = [
+    `${styleText('cyan', 'cd')} ${projectName}`,
+    `${styleText('cyan', getRunCommand('start'))}`,
```
askoufis: Shouldn't all of `cd ${projectName}` be styled here? If so then the backticks shouldn't be necessary around the `styleText`:

```suggestion
    styleText('cyan', `cd ${projectName}`)},
    styleText('cyan', getRunCommand('start')),
```

packages/create/src/actions/createProject.ts
```diff
@@ -0,0 +1,83 @@
...
+  if (!existsSync(targetPath)) {
+    return;
+  }
+
+  const stats = statSync(targetPath);
+  if (!stats.isDirectory()) {
+    throw new Error(`${targetPath} exists but is not a directory`);
+  }
+
+  const isEmpty = isEmptyDir(targetPath);
```
askoufis, line 66: `isEmptyDir` already handles the case above where the path is not a directory.

packages/create/src/actions/createProject.ts
```diff
@@ -0,0 +1,83 @@
...
+    return cwd();
+  }
+
+  return resolve(cwd(), projectName);
+};
+
+const validateTargetDirectory = (targetPath: string) => {
+  if (!existsSync(targetPath)) {
+    return;
+  }
```
askoufis, line 64: `exists` is an existing utility that we should move to the shared utils package. Can do it in a separate PR though.

.changeset/yellow-lions-enjoy.md
```diff
@@ -0,0 +1,45 @@
...
+---
+
+Add new standalone create-sku CLI package
```
askoufis:
```suggestion
Release new standalone `sku` template creation CLI
```

.changeset/yellow-lions-enjoy.md
```diff
@@ -0,0 +1,45 @@
...
+
+Introduces a new `@sku-lib/create` package that provides a standalone CLI for scaffolding new sku projects. This package replaces the functionality of `sku init` with a dedicated tool optimized for project creation. The existing `sku init` command is expected to be deprecated in a future release.
```
askoufis: I'd say we deprecate `sku init` and guide users towards the create package immediately after this PR. Don't see a reason not to do that sooner rather than later.

```suggestion
Introduces a new `@sku-lib/create` package that provides a standalone CLI for scaffolding new sku projects. This package replaces the functionality of `sku init` with a dedicated tool optimized for project creation. The existing `sku init` command is now deprecated and will be removed in a future release.
```

packages/create/src/cli.ts
```diff
@@ -33,9 +35,23 @@ program
...
+
```
askoufis, line 83: I think we need a warning about the experimental nature of the Vite template here.

.changeset/yellow-lions-enjoy.md
```diff
@@ -0,0 +1,45 @@
...
+- Comprehensive test coverage with unit and integration tests
+
+**Generated Projects Include:**
+- `sku.config.ts` - Bundler configuration
+- `package.json`
+- TypeScript setup with proper configs
+- ESLint and Prettier configuration
+- `.gitignore` and `README.md` files
+- Starter React app
+- `braid-design-system` and `vanilla-extract` pre-installed
```
askoufis: IMO there's not that much value in including this info in user-facing release notes as they don't really provide much value to users. Perhaps the generated content list would be better place in the README?

pnpm-workspace.yaml
adenj: Yeah I think pnpm might have done this? I've restored it to its original state

tests/node/sku-create.test.ts
adenj: I'm not sure I understand, it looks to have replaced versions in the snapshot
https://github.com/seek-oss/sku/pull/1383/files#diff-91b3ccc12ff4cfd9efebc5ff87af6330c49b733df4b532a39a93fd8e18983cd4R86-R89

.changeset/yellow-lions-enjoy.md
adenj: Mm yeah I got carried away trying to write a descriptive changeset. Looking back on it, probably not that useful, even in the readme

tests/node/sku-create.test.ts
askoufis: That's the `package.json` snapshot, not the `pnpm-workspace.yaml` snapshot. The latter is the one that is different for some reason.

packages/create/src/generators/templates.ts
```diff
@@ -0,0 +1,140 @@
+import { join, dirname, relative } from 'node:path';
+import { writeFile, mkdir } from 'node:fs/promises';
+import { fdir as Fdir } from 'fdir';
+import { Eta } from 'eta';
+import { fileURLToPath } from 'node:url';
+import dedent from 'dedent';
```
askoufis, line 6: Ran [a snapshot](https://github.com/seek-oss/sku/actions/runs/17536063998) locally and got an error because `dedent` isn't declared as a dependency.

packages/create/src/generators/packageJson.test.ts
```diff
@@ -0,0 +1,95 @@
...
+          "build": "sku build",
+          "test": "sku test",
+          "format": "sku format",
+          "lint": "sku lint"
+        }
+      }
+      "
+    `);
+
+    await fixture.rm();
```
askoufis, line 34: If you use [explicit resource management](https://github.com/privatenumber/fs-fixture?tab=readme-ov-file#using-keyword-explicit-resource-management) then these cleanup calls can be removed.

packages/create/src/cli.ts
```diff
@@ -25,17 +27,48 @@ program
...
+        message:
+          '⚠️  Vite bundler in sku is currently experimental and not yet suitable for production use. Continue?',
+        initial: false,
+      });
+
+      if (!confirmVite) {
+        console.log(
+          '❌ Cancelled. Use `webpack` template for a stable production-ready experience.',
+        );
+        process.exit(0);
```
askoufis: I feel like anything other than a success should be exit code 1 to signify some kind of failure.

packages/create/src/cli.ts
```diff
@@ -25,17 +27,48 @@ program
...
 
-    console.log('🚧 create-sku is not yet implemented');
-    console.log(`Project name: ${projectName}`);
-    console.log(`Template: ${selectedTemplate}`);
+    if (selectedTemplate === 'vite') {
+      const { confirmVite } = await prompts({
+        type: 'confirm',
+        name: 'confirmVite',
+        message:
+          '⚠️  Vite bundler in sku is currently experimental and not yet suitable for production use. Continue?',
```
askoufis:
```suggestion
          '⚠️ Vite support in sku is currently experimental and not yet suitable for production use. Continue?',
```

packages/create/src/cli.ts
```diff
@@ -15,7 +17,7 @@ program
   .version(version)
   .argument('[project-name]', 'Name of the project to create')
   .option('-t, --template <template>', 'Template to use (webpack, vite)')
-  .action(async (projectName: string = '.', options: { template?: string }) => {
+  .action(async (targetDir: string = '.', options: { template?: string }) => {
```
askoufis: Defaulting immediately to the current directory could be a bit confusing as there's no indication that this is happening until creation has already started.

I feel like this project name could be another prompt just like the template. That way if a name isn't provided to the CLI you can indicate that either a directory name or `.` is valid.

tests/node/sku-create.test.ts
adenj: Turns out I wasn't installing `pnpm-plugin-sku` as part of the set up, have rectified that

packages/create/src/generators/packageJson.test.ts
adenj: I did try and use `using` originally, but ran into issues and decided to look into it after we release the package

packages/create/package.json
askoufis, line 1: The `template` files aren't present in the built package. `templates` needs to be added to `package.json#files`. Tests aren't failing because the tests use source code.

pnpm-workspace.yaml
```diff
@@ -38,7 +38,7 @@ catalog:
   '@types/react-dom': '^19.1.6'
   '@vanilla-extract/vite-plugin': ^5.1.1
   '@vocab/vite': ^0.2.3
-  braid-design-system: '^33.11.0'
+  braid-design-system: '33.11.0'
```
askoufis: We intentionally keep this unlocked so lockfile updates can get new versions. I think if you unpin this now it _should_ stay resolved to `33.11.0` though.

askoufis (approved):
Huge effort getting this over the line 🏅

PR #1387, opened 2025-09-08
`test (vitest)`: Disable vitest globals

williamlark:
Turning off Vitest globals to match the default Vitest settings. I've also updated the vitest migration docs so that the specific changes users might need to make (or at least be aware of) are in their own section. I've also highlighted why this change was made in the docs.

Marking it as a major version bump since this would break peoples tests. Though, it's still technically experimental so I can be convinced otherwise. Semantically this would be a major release though since it's breaking.

### Summary of changes

* Vitest globals turned off
* Docs updated
* Vitest is now a peer dep since projects will need to install it locally to add the imports.


.changeset/deep-cooks-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sku-lib/vitest': major
```
askoufis: `minor` would still be a semantic break since the package is still 0-versioned.

packages/vitest/package.json
```diff
@@ -34,12 +34,16 @@
   },
   "homepage": "https://github.com/seek-oss/sku#readme",
   "devDependencies": {
-    "typescript-transform-paths": "^3.5.2"
+    "typescript-transform-paths": "^3.5.2",
+    "vitest": "catalog:"
   },
   "dependencies": {
     "@vanilla-extract/vite-plugin": "catalog:",
     "vitest": "catalog:",
```
askoufis: Since we're making this a peer dependency, I don't think we need it as a regular dep anymore.

williamlark: Ah yes, forgot to remove it

docs/docs/vite.md
```diff
@@ -155,7 +155,19 @@ export default {
 
 `sku` will now invoke [the `vitest` CLI][Vitest CLI] instead of the `jest` CLI when running `sku test`.
 
-**A key difference** between the `jest` and `vitest` CLIs is that `vitest` defaults to watch mode when running tests locally.
+#### Key differences between Vitest and Jest
+
+Vitest has strong compatibility with the Jest API, however it still has differences that may affect your tests. Differences that affect sku projects are listed here. The full list is documented in the Vitest documentation: [Migrating from Jest to Vitest].
```
askoufis: Can we keep line breaks after each sentence please? I'd like to enforce this via linting but haven't found anything that works. See https://sembr.org/.

docs/docs/vite.md
```diff
@@ -168,18 +180,26 @@ To run test without watch mode you can use `sku test --run`:
 }
 ```
 
-Your tests are unlikely to work immediately as there are some differences between Jest and Vitest that may need addressing.
-The [Jest -> Vitest][codemod] can fix some of these differences automatically:
+**Globals disabled**
 
-```sh
-pnpm dlx codemod jest/vitest
+Jest has their globals API enabled by default. Vitest does not. Sku adopts the Vitest defaults, meaning you will need to explicitly import test functions (`expect`, `beforeEach`, etc.). This change results in a cleaner global namespace, better type safety, clearer dependencies and is more consistency with modern javascript practices.
```
askoufis:
```suggestion
Jest enables global APIs such as `it`, `describe`, `beforeAll`, etc., by default. Vitest does not. Sku adopts the Vitest defaults, meaning you will need to explicitly import test functions (`expect`, `beforeEach`, etc.). This change results in a cleaner global namespace, better type safety, clearer dependencies and is more consistent with modern JavaScript practices.
```

docs/docs/vite.md
```diff
@@ -168,18 +180,26 @@ To run test without watch mode you can use `sku test --run`:
...
-This aligns with the Jest API and allows you to write tests without needing to import these functions.
+```diff
+import '@testing-library/jest-dom';
+
++ import { cleanup } from '@testing-library/react';
++ import { afterEach } from 'vitest';
+
++ afterEach(() => {
++  cleanup();
++ });
```
askoufis: I like how concise this is but happy to leave it as is.
```suggestion
+ afterEach(cleanup);
```

williamlark: Happy to do that. I have no preference -- I just copied the snippet from the testing-library docs

askoufis (commented):
Now that #1383 is merged we should add `vitest` as a peer dependency to the vite template.

askoufis:
Taking over from Will to get this over the line:

- Added `vitest` and `@sku-lib/vitest` deps to the vite template (note that we don't test this template currently, but it works)
- Made separate test files for `jest` and `vitest` because the vitest doesn't use globals
- Fixed a bug causing `sku test --run` to hang when using vitest

PR #1388, opened 2025-09-08
Add `jest-to-vitest` codemod and other codemod tweaks

askoufis:
Added a `jest-to-vitest` codemod that handles more cases than the previously recommended [`jest/vitest` codemod](https://app.codemod.com/registry/jest/vitest). Specifically it handles:
- Renaming `jest` to `vi`
- Importing globals such as `expect, it, describe, beforeAll`
- Updating `jest.requireActual` to `await vi.importActual`
- Wrapping functions with `async` that call `await vi.importActual`

And it's much faster, taking ~2s to parse and transform ~2k files, compared to ~20s for the other codemod (though that on does default to using only 4 threads).

I've also tweaked the codemod CLI to no longer accept a glob pattern. It now only accepts a single path to a directory. This was done so we can better control which files are parsed, as the previous implementation would pick up _all_ files, potentially resulting in the codemod parser treating non-JS/TS files as JS.

I also fixed a bug that was resulting in an incorrect changed files count to be reported. Finally, the `sku-codemod` tests have been refactored to enable testing of multiple different codemods.

### DIFF
--- a/.changeset/tender-parrots-refuse.md
+++ b/.changeset/tender-parrots-refuse.md
@@ -1,0 +1,5 @@
+---
+'@sku-lib/codemod': minor
+---
+
+WIP: Add `jest-to-vitest` codemod
--- a/packages/codemod/src/transform/runner.ts
+++ b/packages/codemod/src/transform/runner.ts
@@ -58,7 +58,7 @@ export const getPathFromPrompt = async (): Promise<string> => {
 
 export const runTransform = async (
   transform: string,
-  path: string,
+  _path: string,
   options: Options,
 ): Promise<void> => {
   if (options.dry) {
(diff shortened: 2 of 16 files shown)


packages/codemod/src/transform/runner.ts
```diff
@@ -141,8 +141,10 @@ export const runTransform = async (
   process.exit(0);
 };
 
-const getAllFiles = (paths: string[]) =>
-  glob([...paths, '!**/node_modules', '!**/dist'], {
+const getAllFiles = (directory: string) =>
+  glob(join(directory, '**/*.?(c|m){js,ts}?(x)'), {
+    ignore: ['**/node_modules', '**/dist', ''],
```
adenj: Is the empty ignore string a typo?

.changeset/tender-parrots-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sku-lib/codemod': minor
+---
+
+WIP: Add `jest-to-vitest` codemod
```
adenj: Is this still a WIP?

packages/codemod/src/transform/runner.ts
askoufis: lol yes

.changeset/tender-parrots-refuse.md
askoufis: Not anymore!

PR #1396, opened 2025-09-15
`codemod`: Add support for converting jest mocked typecasts to `vi.mocked` function calls

askoufis:
Patterns such as `const mockedFoo = foo as jest.Mock;` and `const mockedFoo = foo as jest.MockedFunction<typeof foo>;` seem to be common enough patterns that a codemod would be useful to convert them to use `vi.mocked()`.

### DIFF
--- a/tests/node/sku-codemods.test.ts
+++ b/tests/node/sku-codemods.test.ts
@@ -112,6 +112,13 @@ const testCases: TestCase[] = [
 
       const standalone = jest.requireActual('standalone');
 
+      const mockedFoo = foo as jest.Mock;
+      const mockedFoo = foo as jest.Mock<any>;
+      const mockedFoo = foo as jest.Mock<any, any, any>;
+      const mockedFoo = foo as jest.MockedFunction;
+      const mockedFoo = foo as jest.MockedFunction<typeof foo>;
+      const mockedFoo = foo as jest.MockedFunction<typeof foo> & { otherProperty: any };
+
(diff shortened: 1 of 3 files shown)


tests/node/sku-codemods.test.ts
```diff
@@ -163,6 +170,13 @@ const testCases: TestCase[] = [
 
       const standalone = await vi.importActual('standalone');
 
+      const mockedFoo = vi.mocked(foo);
+      const mockedFoo = vi.mocked(foo);
+      const mockedFoo = vi.mocked(foo);
+      const mockedFoo = vi.mocked(foo);
+      const mockedFoo = vi.mocked(foo);
+      const mockedFoo = vi.mocked(foo);
```
askoufis, line 178: This is a bit of a blunt hammer approach that doesn't attempt to reconstruct a valid type from the typecast, instead relying purely on the result of `vi.mocked`.

The intention isn't to end up with type-safe code, but rather to make consumers re-evaluate their usage of typecasting so they hopefully land on a better solution (i.e not typecasting).

williamlark: I do prefer this since ideally people shouldn't have to cast. Hopefully this change doesn't confuse the consumer too much but I prefer the blunt hammer approach here. Let's see how it goes

askoufis: We can always make this behaviour opt-out if it ruffles too many feathers.

PR #1404, opened 2025-09-16
`codemod`: Correctly replace `jest.fn` with `vi.fn` inside mock factories

askoufis:
An alternative solution would be to commit edits and re-parse the result between some steps of the codemod, but that's less efficient so I opted against it.

### DIFF
--- a/.changeset/tough-wings-throw.md
+++ b/.changeset/tough-wings-throw.md
@@ -1,0 +1,5 @@
+---
+'@sku-lib/codemod': patch
+---
+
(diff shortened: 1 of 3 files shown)


williamlark (commented):
Is it line 27 ` not: { pattern: 'jest.requireActual' }` that's causing it to skip over the `vi.fn()` calls within the the mocks entirely? If so could you _not_ exclude it and then change line 71 to `pattern: 'vi.requireActual',` to fix up the incorrect edits? Saves another explicit findAll and replace at the sake of making some extra edits

PR #1405, opened 2025-09-16
Add warnings for unnecessary polyfills in `start`/`start-ssr` commands

adenj:
# 🏴‍☠️ 🦜 POLY(fill) WANT A CRACKER? 🦜 🏴‍☠️ 
Our browser support policy has been evolving, but the polyfills teams use don't always keep up, resulting in larger bundle sizes than what is necessary. 
We've added warnings during `sku start` and `sku start-ssr` to help teams identify unnecessary polyfills that are no longer needed for modern browsers.

Example output:
Screenshot 2025-09-16 at 3 09 02 pm

### DIFF
--- a/.changeset/deep-ends-vanish.md
+++ b/.changeset/deep-ends-vanish.md
@@ -2,4 +2,6 @@
 'sku': minor
 ---
 
-Add warnings for unnecessary polyfills during development. `sku start` and `sku start-ssr`, along with `sku build` and `sku build-ssr` now warn about polyfills that are no longer needed for modern browsers to help reduce bundle sizes.
+Emit warnings when unnecessary polyfills are detected
+
+To help reduce bundle size, start and build commands now warn about polyfills that are no longer necessary in modern browsers.
--- a/.changeset/deep-ends-vanish.md
+++ b/.changeset/deep-ends-vanish.md
@@ -2,4 +2,4 @@
 'sku': minor
 ---
 
-Add warnings for unnecessary polyfills during development. `sku start` and `sku start-ssr` now warn about polyfills that are no longer needed for modern browsers to help reduce bundle sizes.
+Add warnings for unnecessary polyfills during development. `sku start` and `sku start-ssr`, along with `sku build` and `sku build-ssr` now warn about polyfills that are no longer needed for modern browsers to help reduce bundle sizes.
--- a/.changeset/deep-ends-vanish.md
+++ b/.changeset/deep-ends-vanish.md
@@ -1,0 +1,5 @@
+---
+'sku': minor
+---
+
+Add warnings for unnecessary polyfills during development. `sku start` and `sku start-ssr` now warn about polyfills that are no longer needed for modern browsers to help reduce bundle sizes.
--- a/packages/sku/src/program/commands/build-ssr/build-ssr.action.ts
+++ b/packages/sku/src/program/commands/build-ssr/build-ssr.action.ts
@@ -16,7 +16,7 @@ import {
   configureProject,
   validatePeerDeps,
 } from '../../../utils/configure.js';
-import { displayPolyfillWarnings } from '../../../utils/polyfillWarnings.js';
+import { validatePolyfills } from '../../../utils/polyfillWarnings.js';
 import type { StatsChoices } from '../../options/stats/stats.option.js';
 import type { SkuContext } from '../../../context/createSkuContext.js';
 
@@ -37,9 +37,8 @@ export const buildSsrAction = async ({
   const { port, cspEnabled } = skuContext;
   await configureProject(skuContext);
   validatePeerDeps(skuContext);
+  validatePolyfills(skuContext.polyfills);
 
-  // Check for unnecessary polyfills and display warnings
-  displayPolyfillWarnings(skuContext.polyfills);
   try {
     await runVocabCompile(skuContext);
     const [clientConfig, serverConfig] = makeWebpackConfig({
--- a/packages/sku/src/program/commands/build-ssr/build-ssr.action.ts
+++ b/packages/sku/src/program/commands/build-ssr/build-ssr.action.ts
@@ -16,6 +16,7 @@ import {
   configureProject,
   validatePeerDeps,
 } from '../../../utils/configure.js';
+import { displayPolyfillWarnings } from '../../../utils/polyfillWarnings.js';
 import type { StatsChoices } from '../../options/stats/stats.option.js';
 import type { SkuContext } from '../../../context/createSkuContext.js';
 
@@ -36,6 +37,9 @@ export const buildSsrAction = async ({
   const { port, cspEnabled } = skuContext;
   await configureProject(skuContext);
   validatePeerDeps(skuContext);
+
+  // Check for unnecessary polyfills and display warnings
+  displayPolyfillWarnings(skuContext.polyfills);
   try {
     await runVocabCompile(skuContext);
     const [clientConfig, serverConfig] = makeWebpackConfig({
--- a/packages/sku/src/program/commands/start/start.action.ts
+++ b/packages/sku/src/program/commands/start/start.action.ts
@@ -7,6 +7,8 @@ import {
 } from '../../../utils/configure.js';
 import { watchVocabCompile } from '../../../services/vocab/runVocab.js';
 import { checkHosts, withHostile } from '../../../context/hosts.js';
+import { detectUnnecessaryPolyfills } from '../../../utils/polyfillDetector.js';
+import { displayPolyfillWarnings } from '../../../utils/polyfillWarnings.js';
 import chalk from 'chalk';
 
 export const startAction = async (
@@ -30,6 +32,10 @@ export const startAction = async (
 
   withHostile(checkHosts)(skuContext);
   validatePeerDeps(skuContext);
+
+  // Check for unnecessary polyfills and display warnings
+  const detectedPolyfills = detectUnnecessaryPolyfills(skuContext.polyfills);
+  displayPolyfillWarnings(detectedPolyfills);
 
   if (skuContext.bundler === 'vite') {
     const { viteStartHandler } = await import('./vite-start-handler.js');
--- a/packages/sku/src/program/commands/start/start.action.ts
+++ b/packages/sku/src/program/commands/start/start.action.ts
@@ -7,7 +7,6 @@ import {
 } from '../../../utils/configure.js';
 import { watchVocabCompile } from '../../../services/vocab/runVocab.js';
 import { checkHosts, withHostile } from '../../../context/hosts.js';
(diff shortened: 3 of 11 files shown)


packages/sku/src/utils/polyfillRegistry.ts
```diff
@@ -0,0 +1,42 @@
...
+  },
+  'whatwg-fetch': {
+    reason: 'Fetch API is natively supported in modern browsers',
+  },
+  'isomorphic-fetch': {
+    reason: 'Fetch API is natively supported in modern browsers',
+  },
+  'intersection-observer': {
+    reason: 'IntersectionObserver is natively supported in modern browsers',
+  },
```
adenj: I'm positive this list isnt 100% correct/exhaustive, so please highlight anything that doesn't look right

packages/sku/src/utils/polyfillDetector.ts
```diff
@@ -0,0 +1,30 @@
...
+      return [];
+    }
+
+    return [
+      {
+        polyfillName,
+        ...registryEntry,
+      },
+    ];
+  });
```
williamlark: `nit`: Performance isn't an issue here so I recommend filtering and mapping separately. It's a little simpler and easier to  comprehend compared to returning empty arrays and flattening them out

```suggestion
export const detectUnnecessaryPolyfills = (
  polyfills: string[]
): DetectedPolyfill[] =>
  polyfills
    .filter((polyfillName) => polyfillName in POLYFILL_REGISTRY)
    .map((polyfillName) => ({
      polyfillName,
      ...POLYFILL_REGISTRY[polyfillName]!,
    }));
```

packages/sku/src/program/commands/start/start.action.ts
```diff
@@ -31,6 +33,10 @@ export const startAction = async (
   withHostile(checkHosts)(skuContext);
   validatePeerDeps(skuContext);
 
+  // Check for unnecessary polyfills and display warnings
+  const detectedPolyfills = detectUnnecessaryPolyfills(skuContext.polyfills);
+  displayPolyfillWarnings(detectedPolyfills);
```
williamlark: Nice that these are in seperate functions, though I reckon you could call `detectUnnecessaryPolyfills` inside `displayPolyfillWarnings`. We don't have any use cases to use them independently at the moment and just calling `displayPolyfillWarnings()` in the places we want to warn is simpler.

packages/sku/src/utils/polyfillWarnings.ts
williamlark, line 1: Do you think it's worth having a fixture to test the logging? The testing library we have should make it pretty easy to check the output. The `polyfillDetector.test.ts` file is doing a good job of testing the logic but it could be worth having a fixture to make sure a simple case works e2e

packages/sku/src/utils/polyfillRegistry.ts
williamlark: Nothing on the top of my head. Maybe see if there is anything you missed in [search](https://github.com/search?q=org%3ASEEK-Jobs+%22polyfills%3A%22+NOT+is%3Aarchived&type=code)?

packages/sku/src/utils/polyfillRegistry.ts
```diff
@@ -0,0 +1,42 @@
...
+  docsUrl?: string;
+}
+
+export type PolyfillRegistry = Record<string, PolyfillRegistryEntry>;
+
+/**
+ * Registry of polyfills that are unnecessary for sku's supported browsers
+ */
+export const POLYFILL_REGISTRY: PolyfillRegistry = {
+  'core-js': {
```
williamlark, line 12: Looks `core-js` can be a prefix as well. Maybe need to support that or add some entries in manually. [Example](https://github.com/SEEK-Jobs/adv-jp-express-ui/blob/4489f11142f717adc27634b7b98954bad7b9725b/sku.config.ts#L221)

williamlark:
The warning looks nice! 👨‍🎨 Do you think start and start-ssr is enough. Do we need it on build too?

packages/sku/src/utils/polyfillRegistry.ts
```diff
@@ -0,0 +1,42 @@
...
+    reason:
+      'Most ES2015-ES2017 features are natively supported in modern browsers',
+    docsUrl: 'https://github.com/zloirock/core-js#usage',
+  },
+  '@babel/polyfill': {
+    reason: 'Deprecated package, replaced by core-js',
```
askoufis: This message might result in someone replacing `@babel/polyfill` with `core-js`, which is also something we don't recommend. Maybe we should just say it's deprecated and nothing else?

packages/sku/src/utils/polyfillRegistry.ts
```diff
@@ -0,0 +1,42 @@
...
+    docsUrl: 'https://babeljs.io/docs/en/babel-polyfill',
+  },
+  'regenerator-runtime': {
+    reason: 'async/await is natively supported in modern browsers',
```
askoufis: I'd prefer the reason be a bit more precise, rather than simply saying "it's modern". E.g we should mention should mention "[Baseline](https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility#baseline_badges) widely available" or "Baseline 2022", as these are more precise and align with common web platform communication. For example:

```suggestion
    reason: 'async/await is well established and works across many devices and browser versions.',
```

(This wording was taken directly from the "Baseline widely available" messaging, but it doesn't have to be an exact copy)

packages/sku/src/utils/polyfillRegistry.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export type PolyfillRegistry = Record<string, PolyfillRegistryEntry>;
+
+/**
+ * Registry of polyfills that are unnecessary for sku's supported browsers
+ */
+export const POLYFILL_REGISTRY: PolyfillRegistry = {
+  'core-js': {
+    reason:
+      'Most ES2015-ES2017 features are natively supported in modern browsers',
```
askoufis: `core-js` contains polyfills for a lot more than just ES2015-ES2017 features. We might need to either be more specific about which polyfills we target, or maybe put a blanket message that says "`core-js` polyfills lots of behaviour that you may or may not use. Consider either removing it or using targeted polyfill for specific APIs".

packages/sku/src/utils/polyfillWarnings.ts
askoufis: I think this makes sense. We already have a polyfills fixture, so adding one of the unnecessary polyfills to its sku config would be an easy way to test it.

adenj: Yeah good shout, I was using that fixture locally. I've added a `config.bad` config file and corresponding start command to the fixture

.changeset/deep-ends-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
```
adenj, line 1: I feel like this changeset file name would go kinda hard as a band name

packages/sku/src/program/commands/build-ssr/build-ssr.action.ts
```diff
@@ -36,6 +37,9 @@ export const buildSsrAction = async ({
   const { port, cspEnabled } = skuContext;
   await configureProject(skuContext);
   validatePeerDeps(skuContext);
+
+  // Check for unnecessary polyfills and display warnings
+  displayPolyfillWarnings(skuContext.polyfills);
```
askoufis: This comment seems redundant. Maybe renaming the function to `validatePolyfills` to match `validatePeerDeps` (which also scans and warns about stuff) would be clearer?

packages/sku/src/utils/polyfillRegistry.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type PolyfillRegistry = Record<string, PolyfillRegistryEntry>;
+
+/**
+ * Registry of polyfills that are unnecessary for sku's supported browsers
+ */
+export const POLYFILL_REGISTRY: PolyfillRegistry = {
+  'core-js': {
+    reason:
+      'Polyfills many features that may be unnecessary. Consider removing or using targeted polyfills for specific APIs',
```
askoufis:
```suggestion
      'Polyfills many features that may be unnecessary. Consider removing or using targeted polyfills for specific APIs.',
```

packages/sku/src/utils/polyfillRegistry.ts
```diff
@@ -0,0 +1,47 @@
...
+    reason: 'Deprecated package',
+    docsUrl: 'https://babeljs.io/docs/en/babel-polyfill',
+  },
+  'babel/polyfill': {
+    reason: 'Deprecated package',
+    docsUrl: 'https://babeljs.io/docs/en/babel-polyfill',
+  },
+  'regenerator-runtime': {
+    reason:
+      'async/await is well established and works across many devices and browser versions',
```
askoufis: Can we use this similar messaging for the promise, fetch and intersection observer polyfills too? In fact maybe we should just copy the message from MDN. I like that it gives a date because it really drives home how old and useless these polyfills are.

```suggestion
      'async/await is well established and works across many devices and browser versions. It has been available across browsers since April 2017.',
```

[Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) were baseline in July 2015, fetch and [async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) were baseline in April 2017, and [intersection observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) was baseline in March 2019.

.changeset/deep-ends-vanish.md
```diff
@@ -0,0 +1,5 @@
...
+'sku': minor
+---
+
+Add warnings for unnecessary polyfills during development. `sku start` and `sku start-ssr` now warn about polyfills that are no longer needed for modern browsers to help reduce bundle sizes.
```
askoufis: `sku build` and `build-ssr` are now included too.

.changeset/deep-ends-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': minor
+---
+
+Add warnings for unnecessary polyfills during development. `sku start` and `sku start-ssr`, along with `sku build` and `sku build-ssr` now warn about polyfills that are no longer needed for modern browsers to help reduce bundle sizes.
```
askoufis: Just a bit of a reword.

```suggestion
Emit warnings when unnecessary polyfills are detected

To help reduce bundle size, start and build commands now warn about polyfills that are no longer necessary in modern browsers.
```

askoufis (approved):
Approved but have a suggestion for the changeset.

Would also be worthwhile adding a metric for the number of unnecessary polyfills, so we can (ideally) see it decrease over time. Happy to chat about the implementation if you're unfamiliar with how sku's metrics work.

PR #1414, opened 2025-09-23
Worker logs are now debug only

williamlark:
The sku codemod would output a log whenever a worker is started. This isn't useful information to most people so I've moved it to a debug log instead. 

#### Before
CleanShot 2025-09-23 at 14 00 19

#### After
CleanShot 2025-09-23 at 14 00 34


pnpm-lock.yaml
```diff
@@ -18070,6 +18093,157 @@ snapshots:
       - webpack-plugin-serve
       - yaml
 
+  sku@file:packages/sku(8e85ba77538caef8ce19d3b6ffc65c92):
```
askoufis: This is sus.

williamlark: `pnpm dedupe` seems to fix it

PR #1416, opened 2025-09-24
Extend unnecessary polyfill detection to include dependencies

adenj:
# 🦜 POLY(fill) WANT A CRACKER? 2: ELECTRIC BOOGALOO 🦜
 Builds on the unnecessary polyfill detection introduced in #1405 by expanding detection beyond `sku.config.ts` to also scan `package.json` dependencies and devDependencies. 

Warnings will include the source of the polyfill (config or deps), and action guidance
E.g.
Before
```
  ❌ raf
     requestAnimationFrame is well established and works across many devices and browser versions. It has been available across browsers since July 2015.
```

  After
```
  ❌ raf (found in dependencies)
     requestAnimationFrame is well established and works across many devices and browser versions. It has been available across browsers since July 2015.
     Action: Remove from dependencies in package.json
```

### DIFF
--- a/packages/sku/src/utils/detectUnnecessaryPolyfillsFromDependencies.ts
+++ b/packages/sku/src/utils/detectUnnecessaryPolyfillsFromDependencies.ts
@@ -1,0 +1,56 @@
+import { readFileSync, existsSync } from 'node:fs';
+import { getPathFromCwd } from '@sku-lib/utils';
+import { POLYFILL_REGISTRY } from './polyfillRegistry.js';
+import type { DetectedPolyfillWithSource } from './polyfillDetector.js';
+
+/**
+ * Detects unnecessary polyfills from package.json dependencies
+ * by matching package names against the registry.
+ */
+export const detectUnnecessaryPolyfillsFromDependencies =
+  (): DetectedPolyfillWithSource[] => {
+    const packageJsonPath = getPathFromCwd('package.json');
+
+    if (!existsSync(packageJsonPath)) {
+      return [];
+    }
+
+    try {
+      const packageJsonContent = readFileSync(packageJsonPath, 'utf-8');
+      const packageJson = JSON.parse(packageJsonContent);
+
+      const dependencies = packageJson.dependencies || {};
+      const devDependencies = packageJson.devDependencies || {};
+
+      const detectedPolyfills: DetectedPolyfillWithSource[] = [];
+
+      // Check dependencies
+      for (const depName of Object.keys(dependencies)) {
+        if (depName in POLYFILL_REGISTRY || depName.startsWith('core-js')) {
+          detectedPolyfills.push({
+            polyfillName: depName,
+            detectionSource: 'dependency',
+            dependencyType: 'dependencies',
+            ...(POLYFILL_REGISTRY[depName] || POLYFILL_REGISTRY['core-js']!),
+          });
+        }
+      }
+
+      // Check devDependencies
+      for (const depName of Object.keys(devDependencies)) {
+        if (depName in POLYFILL_REGISTRY || depName.startsWith('core-js')) {
+          detectedPolyfills.push({
+            polyfillName: depName,
+            detectionSource: 'dependency',
+            dependencyType: 'devDependencies',
+            ...(POLYFILL_REGISTRY[depName] || POLYFILL_REGISTRY['core-js']!),
+          });
+        }
+      }
+
+      return detectedPolyfills;
+    } catch {
+      // If we can't read or parse package.json, return empty array
+      return [];
+    }
+  };
--- a/packages/sku/src/utils/detectUnnecessaryPolyfillsFromDependencies.ts
+++ b/packages/sku/src/utils/detectUnnecessaryPolyfillsFromDependencies.ts
@@ -1,7 +1,9 @@
 import { readFileSync, existsSync } from 'node:fs';
 import { getPathFromCwd } from '@sku-lib/utils';
-import { POLYFILL_REGISTRY } from './polyfillRegistry.js';
-import type { DetectedPolyfillWithSource } from './polyfillDetector.js';
+import {
+  type DetectedPolyfillWithSource,
+  getDeprecatedPolyfill,
+} from './polyfillDetector.js';
 
 /**
  * Detects unnecessary polyfills from package.json dependencies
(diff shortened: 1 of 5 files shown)


packages/sku/src/utils/detectUnnecessaryPolyfillsFromDependencies.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Detects unnecessary polyfills from package.json dependencies
+ * by matching package names against the registry.
+ */
+export const detectUnnecessaryPolyfillsFromDependencies =
+  (): DetectedPolyfillWithSource[] => {
+    const packageJsonPath = getPathFromCwd('package.json');
+
+    if (!existsSync(packageJsonPath)) {
+      return [];
+    }
```
williamlark: Would this ever be possible? I feel like this should throw an error if no package.json is detected since that doesn't seem like an expected case, unless this can trigger on sku init somehow

packages/sku/src/utils/detectUnnecessaryPolyfillsFromDependencies.ts
```diff
@@ -0,0 +1,56 @@
...
+      const packageJson = JSON.parse(packageJsonContent);
+
+      const dependencies = packageJson.dependencies || {};
+      const devDependencies = packageJson.devDependencies || {};
+
+      const detectedPolyfills: DetectedPolyfillWithSource[] = [];
+
+      // Check dependencies
+      for (const depName of Object.keys(dependencies)) {
+        if (depName in POLYFILL_REGISTRY || depName.startsWith('core-js')) {
```
williamlark: Some shared logic here from the config file. Could move this out


```suggestion
        if (isDeprecatedPolyfill(depName)) {
```

packages/sku/src/utils/detectUnnecessaryPolyfillsFromDependencies.ts
```diff
@@ -0,0 +1,56 @@
...
+          detectedPolyfills.push({
+            polyfillName: depName,
+            detectionSource: 'dependency',
+            dependencyType: 'dependencies',
+            ...(POLYFILL_REGISTRY[depName] || POLYFILL_REGISTRY['core-js']!),
```
williamlark: Same here, reused logic could be moved out.

Come to think of it, you could probably just have `getDeprecatedPolyfill`. It either returns the pollyfill or null.

```suggestion
            ...(getDeprecatedPolyfill(depName)),
```

packages/sku/src/utils/detectUnnecessaryPolyfillsFromDependencies.ts
```diff
@@ -0,0 +1,56 @@
...
+          });
+        }
+      }
+
+      // Check devDependencies
+      for (const depName of Object.keys(devDependencies)) {
```
williamlark: These loops are similar. Could you combine them into a nested loop? e.g.,

```ts
for (const [dependencyType, deps] of Object.entries({ dependencies, devDependencies })) {
  for (const depName of Object.keys(deps)) {
    const polyfill = getDeprecatedPolyfill(depName);
    if (!polyfill) {
      continue;
    }

    detectedPolyfills.push({
      polyfillName: depName,
      detectionSource: "dependency",
      dependencyType,
      ...polyfill,
    });
  }
}

```

packages/sku/src/utils/detectUnnecessaryPolyfillsFromDependencies.ts
```diff
@@ -0,0 +1,56 @@
...
+            dependencyType: 'devDependencies',
+            ...(POLYFILL_REGISTRY[depName] || POLYFILL_REGISTRY['core-js']!),
+          });
+        }
+      }
+
+      return detectedPolyfills;
+    } catch {
+      // If we can't read or parse package.json, return empty array
+      return [];
```
williamlark: Hiding an error here might lead to missing issues. I'd prefer this to throw or at least have a debug log.

packages/sku/src/utils/polyfillDetector.test.ts
williamlark, line 1: Do you have any package.json polyfill test you could add?

packages/sku/src/utils/detectUnnecessaryPolyfillsFromDependencies.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    if (!existsSync(packageJsonPath)) {
+      throw new Error(
+        `package.json not found at ${packageJsonPath} - this is required for polyfill detection`,
+      );
+    }
+
+    try {
+      const packageJsonContent = readFileSync(packageJsonPath, 'utf-8');
+      const packageJson = JSON.parse(packageJsonContent);
```
williamlark, line 18: It's generally not recommended to check for file existence before reading the file because of race conditions. It's not a big deal here, but readFile will throw an error anyway, and you can check the error code (ENOENT I think) if you want to display a specific error message. 


```suggestion
    const packageJsonPath = getPathFromCwd('package.json');

    try {
      const packageJsonContent = readFileSync(packageJsonPath, 'utf-8');
      const packageJson = JSON.parse(packageJsonContent);
```

packages/sku/src/utils/polyfillDetector.test.ts
```diff
@@ -1,7 +1,21 @@
...
 import { detectUnnecessaryPolyfills } from './polyfillDetector.js';
 
+vi.mock('node:fs', () => ({
+  existsSync: vi.fn(),
+  readFileSync: vi.fn(),
+}));
+
 describe('polyfillDetector', () => {
+  beforeEach(async () => {
+    const { existsSync, readFileSync } = await import('node:fs');
```
williamlark: Does the mock not work if this is imported up the top of the file? 🤔 I don't _think_ it needs a dynamic import

packages/sku/src/utils/polyfillDetector.test.ts
```diff
@@ -67,22 +84,63 @@ describe('polyfillDetector', () => {
...
+      };
+
+      const { existsSync, readFileSync } = await import('node:fs');
+      vi.mocked(existsSync).mockReturnValue(true);
+      vi.mocked(readFileSync).mockReturnValue(JSON.stringify(mockPackageJson));
+
+      const result = detectUnnecessaryPolyfills([]);
+
+      expect(result).toHaveLength(2);
+      expect(
```
williamlark: Could use an inline snapshot here instead like the other test 🖼️

williamlark (approved):
Nice work. Some minor comments but looks good

packages/sku/src/utils/polyfillDetector.test.ts
adenj: Ah yeah I was messing around with the mocking and forgot to clean this up

PR #1417, opened 2025-09-24
Prevent async being appended to intentionally synchronous functions

williamlark:
When running the codemod it would previously add `async` to functions that where intentionally written as synchronous. The ast rules were matching functions _inside_ jest.mock implementations, which is not needed. 

This change removes the `jest.mock` pattern match and instead matches the functions that have `jest.requireActual`  inside it directly. `jest.requireActual` will be converted into `vi.importActual` which is an async function, so the enclosing function will need to be async regardless of whether it's in a `jest.mock` or not. 

This also catches instances where the mock is defined outside of a `jest.mock` call. Not common, but an edge case that this catches now.

### DIFF
--- a/packages/sku/src/utils/detectUnnecessaryPolyfillsFromConfig.ts
+++ b/packages/sku/src/utils/detectUnnecessaryPolyfillsFromConfig.ts
@@ -1,0 +1,19 @@
+import {
+  type DetectedPolyfillWithSource,
+  getDeprecatedPolyfill,
+} from './polyfillDetector.js';
+
+/**
+ * Detects unnecessary polyfills from the configured polyfills array
+ * by matching package names against the registry.
+ */
+export const detectUnnecessaryPolyfillsFromConfig = (
+  polyfills: string[],
+): DetectedPolyfillWithSource[] =>
+  polyfills
+    .filter((polyfillName) => getDeprecatedPolyfill(polyfillName))
+    .map((polyfillName) => ({
+      polyfillName,
+      detectionSource: 'config' as const,
+      ...getDeprecatedPolyfill(polyfillName)!,
+    }));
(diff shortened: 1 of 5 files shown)


packages/codemod/src/codemods/jest-to-vitest.ts
```diff
@@ -13,8 +13,8 @@ const jestGlobals = [
 const serializeImports = (imports: Set<string>) =>
   Array.from(imports).sort().join(', ');
 
-export const transform = (source: string) => {
-  const ast = parse(Lang.Tsx, source);
+export const transform = async (source: string) => {
+  const ast = await parseAsync(Lang.Tsx, source);
```
williamlark, line 17: According to the docs `parseAsync` is preferred. [[link](https://ast-grep.github.io/guide/api-usage/performance-tip.html#prefer-parseasync-over-parse)]

askoufis: I saw that too, however I'm not sure if there's any benefit to our use case since we're running these transforms in their own worker threads already.

I did a quick benchmark of the codemod in a large internal repo and the async version was actually slower, but that's comparing ~750ms to ~800ms, so it's really not a big deal.

tests/node/sku-codemods.test.ts
```diff
@@ -252,6 +276,56 @@ const testCases: TestCase[] = [
...
+        },
+        qux: async () => {
+          return 'qux';
+        },
+        wert: function () {
+          return 'wert';
+        },
+        pulp: async function () {
+          return 'pulp';
+        },
```
askoufis, line 300: I think this should work, but can we also test a function that calls `jest.requireActual` (even though this doesn't really make any sense)?
E.g.
```suggestion
        },
        blorp: () => {
          return jest.requireActual();
        }
```

williamlark: I've finally got this working. It seems that there is an issue with making edits on matches within another match. I've had to match on `formal_parameters` instead so that the edit works on nested matches.

packages/codemod/src/codemods/jest-to-vitest.ts
williamlark: As discussed we'll keep this since it might be useful for when we look in to the [findInFiles](https://ast-grep.github.io/guide/api-usage/performance-tip.html#prefer-findinfiles-when-possible) API

jahredhope (approved):

PR #1418, opened 2025-09-24
Update commander to latest

williamlark:
Upgrading commander to the latest version. This will need a major release since extra, unsupported arguments passed to a command will now throw an error.

**Note:** We will wait to merge this until we are ready for a major release


.changeset/ten-hotels-notice.md
```diff
@@ -0,0 +1,23 @@
+---
+'@sku-lib/codemod': major
+'@sku-lib/create': major
+'sku': major
+---
```
williamlark, line 3: I as considering breaking this up into seperate changesets. Thoughts?

If we're going out with a major for sku I don't mind doing a major for create and codemod.

.changeset/ten-hotels-notice.md
```diff
@@ -0,0 +1,23 @@
...
+
+Updating commander.js to the latest version.
```
askoufis: I'm honestly not a fan of continuous-tense messages like this. I don't think it makes sense in the context of a changelog/commit as it implies "something is still happening", when really something has already happened.

I much prefer [imperative mood](https://cbea.ms/git-commit/#imperative), e.g. `Update commander.js to the latest version`.

.changeset/ten-hotels-notice.md
askoufis: Separate changesets could be good, but at the same time I doubt anyone is reading the changelogs for the create and codemod packages since they're never used as dependencies.

Releasing a major for create and codemod would make them 1.0.0, which is probably fine, but I'd prefer not rushing out a 1.0 release just for the sake of it, and rather use it to signal API stability. But again, for packages that people never really depend on it's likely fine. Your call.

askoufis (commented):
The main command I'm concerned about is `test` because that one passes through arguments and flags, so as long as that one still works then I'm happy.

williamlark:
I've separated the changesets for the three packages. codemod and create are ready for a stable release so I think going out with a 1.0.0 is fine.

.changeset/solid-socks-ask.md
```diff
@@ -0,0 +1,11 @@
+---
+'@sku-lib/codemod': major
+'@sku-lib/create': major
+---
+
+Update commander.js to the latest version.
```
jahredhope: Should the release note be more focused on the impact rather than the change?

```suggestion
Throw error when given unsupported arguments.
This is now the default behaviour of the updated commander.js version.
```

.changeset/ten-hotels-notice.md
```diff
@@ -0,0 +1,20 @@
+---
+'sku': major
+---
+
+Update commander.js to the latest version.
```
jahredhope: Same as above suggestion re: impact focused

tests/node/cli-arguments.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { scopeToFixture } from '@sku-private/testing-library';
+
+const { sku } = scopeToFixture('cli-arguments');
+
+describe('cli-arguments', () => {
+  it('should throw an error when an unknown option is provided', async () => {
+    const process = await sku('build', ['--foo']);
+
+    expect(
```
jahredhope, line 11: Should this also check for a non-zero status code?

PR #1419, opened 2025-09-24
Bumping minimum node version to 22.19.0

williamlark:
Bumping minimum node version to 22.19.0. This will require a major release.

Also bumping the pnpm version of the monorepo to 10.17.1. This doesn't affect consumers.

**Note:** Do not merge until we are ready for a major release


.changeset/chatty-hotels-cover.md
```diff
@@ -0,0 +1,10 @@
+---
+'@sku-lib/codemod': major
+'@sku-lib/create': major
+'@sku-lib/vitest': major
+'@sku-lib/utils': major
+'@sku-lib/vite': major
+'sku': major
+---
+
+Minimum node version bumped to 22.20.0
```
askoufis: This should have stronger messaging. See https://github.com/seek-oss/sku/releases/tag/sku%4014.0.0#:~:text=minimum%20supported%20node.js.

askoufis (commented):
Re: test failures.

We've hit this issue before when trying to move to a version of node that browserslist doesn't know about yet. [This code](https://github.com/seek-oss/sku/blob/b938dd0899b25d7ea03a24982d185c6eed47f279/postinstall.js#L10-L18) keeps `targets.json` up to date, but [babel preset env](https://github.com/seek-oss/sku/blob/b938dd0899b25d7ea03a24982d185c6eed47f279/packages/sku/src/config/babel.ts#L93-L99) complains about it.

jahredhope (approved):

PR #1420, opened 2025-09-25
Error on missing config file and update default entry extensions to .tsx

adenj:
Includes two breaking changes to improve config handling and modernise defaults:

**Stricter config file validation**
sku now errors and exists when a config file specified with `--config` flag cannot be found (no longer silently falling back to a default config)

**Update default entry file extensions**
Default entry files now use `.tsx` extensions rather than `.js`


.changeset/short-hats-fetch.md
```diff
@@ -0,0 +1,14 @@
...
+**Breaking changes to configuration handling:**
+
+1. **Custom config file validation**: `sku` no longer falls back to default sku config files when the config file specified with the `--config` flag cannot be found. It will now throw an error and exit the program instead. This ensures users are aware that the configuration file is either missing or incorrectly specified, rather than silently falling back to a default configuration that may not be appropriate for their use case. If you encounter this error, ensure that the `--config` flag points to a valid configuration file.
+
+2. **Default entry file extensions**: Default values for entry files now use `.tsx` extensions instead of `.js`:
+   - `clientEntry`: `'src/client.js'` → `'src/client.tsx'`
+   - `renderEntry`: `'src/render.js'` → `'src/render.tsx'`  
+   - `serverEntry`: `'src/server.js'` → `'src/server.tsx'`
+
+   Existing projects with `.js` entry files will need to either rename their files to `.tsx` or explicitly specify the `.js` paths in their sku config.
```
askoufis: Would be great to include a sku config diff for this change, just to make it super obvious what changes are required.

docs/docs/configuration.md
```diff
@@ -26,6 +26,17 @@ If you need to specify a different config file you can do so with the `--config`
...
+> When using the `--config` parameter, the specified file must exist. Sku will exit with an error if the file cannot be found.
+> Config files can use either TypeScript or JavaScript.
+
+When **no** `--config` parameter is provided, sku will automatically look for config files in this order:
+
+1. `sku.config.ts`
+2. `sku.config.js`
+3. `sku.config.mjs`
+
+If none of these files exist, sku will use its built-in default configuration.
```
askoufis, line 38: This is the feature we're removing right?
```suggestion
```

adenj: No, it will still fallback to the default config when the no `--config` flag is provided and and none of the above config files can be found.

What's being removed is the behaviour of silently falling back to the default config when the user provides an invalid config file path

.changeset/short-hats-fetch.md
```diff
@@ -0,0 +1,24 @@
+---
+'sku': major
+---
+
+**Breaking changes to configuration handling:**
+
+1. **Custom config file validation**: `sku` no longer falls back to default sku config files when the config file specified with the `--config` flag cannot be found. It will now throw an error and exit the program instead. This ensures users are aware that the configuration file is either missing or incorrectly specified, rather than silently falling back to a default configuration that may not be appropriate for their use case. If you encounter this error, ensure that the `--config` flag points to a valid configuration file.
+
+2. **Default entry file extensions**: Default values for entry files now use `.tsx` extensions instead of `.js`:
```
williamlark: `nit`: Sorry, but could these be two seperate changesets? 😅  It'll just look a bit nicer in the changeset in this format with two seperate files 🙏 :

```
Throw on non-existing `--config` file

**BREAKING CHANGE**:

`sku` no longer falls back to default sku config files when the config file specified with the `--config` flag cannot be found. It will now throw an error and exit the program instead. This ensures users are aware that the configuration file is either missing or incorrectly specified, rather than silently falling back to a default configuration that may not be appropriate for their use case. If you encounter this error, ensure that the `--config` flag points to a valid configuration file.
```

packages/sku/src/context/configPath.ts
williamlark, line 1: Would be great to have a fixture that tests this. Could be a basically empty fixture with a package.json and just test running sku with the flag.

fixtures/ssr-hello-world/sku-build.config.ts
```diff
@@ -2,6 +2,8 @@ import { makeStableHashes } from '@sku-private/test-utils';
 import type { SkuConfig } from 'sku';
 
 export default {
+  clientEntry: 'src/client.js',
```
jahredhope: Would it be worth updating the fixture to use the default values? I feel like this fixture should be a simple "hello-world" behaviour, so in most cases it should just use the default values.

jahredhope: If we want to test this perhaps there should be a seperate "js" focused fixture to explicitly cover JS use-cases?

But I don't think that's necessary... it's not a common enough scenario.

williamlark: Yea I can swap it to do the default behaviour. Happy to write some tests to test both the ts and js behaviour

williamlark: I've swapped the braid and ssr fixture to use ts files now and removed these specific client/server entry points

PR #1421, opened 2025-09-25
Removing sku-init in favour of `@sku-lib/create`

williamlark:
As part of the next major release we would like to remove more references to sku init. This PR removes all `sku init` command actions and replaces them with an error to the user prompting them to use `@sku-lib/create`.

I've added a few more tests for `create` to test vite and webpack, and that passing flags works. I've also added some test to check for the sku-init error message.


packages/sku/src/program/commands/init/init.command.ts
```diff
@@ -1,19 +1,16 @@
 import { Command } from 'commander';
-import { packageManagerOption } from '../../options/packageManager/packageManager.option.js';
-import { setPackageManager } from '@sku-lib/utils';
 
+/**
+ * This command is deprecated.
+ */
```
askoufis, line 6: So we have a reminder to eventually delete this entirely, we should mark it as deprecated.

```suggestion
/**
 * @deprecated Will be removed entirely in a future major version. Replaced by `@sku-lib/create`.
 */
```

packages/sku/src/program/commands/init/init.command.ts
```diff
@@ -1,19 +1,16 @@
...
-  .addOption(packageManagerOption)
-  .on('option:package-manager', (packageManager) => {
-    setPackageManager(packageManager);
-  })
-  .action(async (projectName, options) => {
-    const { initAction } = await import('./init.action.js');
-    await initAction(projectName, options);
-  });
+  .allowExcessArguments(true)
+  .allowUnknownOption(true)
```
askoufis, line 10: Are these set to `true` so that any invocation of `sku init` will hit the `error` below, instead of argument validation erroring first?

pnpm-workspace.yaml
```diff
@@ -25,30 +47,5 @@ preferWorkspacePackages: true
 publicHoistPattern:
   - '@babel/*'
 
-# This is needed so that the React 18 fixture doesn't clobber React 19 in every other fixture. We can remove this once we've dropped React 18 support.
```
askoufis, line 28: I find it so frustrating that `pnpm` obliterates comments when it updates the workspace file for you :(

private/testing-library/render-cli.ts
```diff
@@ -17,14 +17,10 @@ type SkuCommand =
   | 'test';
 
 export const scopeToFixture = (fixtureFolder: string) => {
-  const appDir = path.dirname(
-    require.resolve(
-      path.join(process.cwd(), `fixtures/${fixtureFolder}/package.json`),
-    ),
+  const fixturePath = makeFixturePathResolver(
```
askoufis, line 20: suggestion: Rename this function to make it a bit clearer as to what it does:
```suggestion
  const resolvePathFromFixtureTo = makeFixturePathResolver(
```

.changeset/proud-ducks-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': major
+---
+
+`sku init` will now exit with a message prompting to use `@sku-lib/create` instead
```
askoufis: I think this should be a bit of a stronger message that signals 2 things:
- `sku init` no longer does anything, alongside it being deprecated
- It will be removed in a future major version

packages/sku/src/program/commands/init/init.command.ts
williamlark: Yep, that's right.

pnpm-workspace.yaml
williamlark: Right! How silly 🤦 I've added the comments back in

private/testing-library/render-cli.ts
williamlark: I might do this separately since it'll touch quite a few tests. Agree it could be clearer

pnpm-lock.yaml
```diff
@@ -201,7 +201,7 @@ importers:
     dependencies:
       braid-design-system:
         specifier: 'catalog:'
-        version: 33.11.0(@types/react@19.1.9)(babel-plugin-macros@3.1.0)(react-dom@19.1.1(react@19.1.1))(react@19.1.1)
+        version: 33.11.5(@types/react@19.1.9)(babel-plugin-macros@3.1.0)(react-dom@19.1.1(react@19.1.1))(react@19.1.1)
```
williamlark, line 204: Braids version bump has made the snapshot hashes need to be updated

tests/browser/__snapshots__/braid-design-system.test.ts.snap
williamlark, line 1: This and sku-webpack-plugin test have been updated since the braid version has been bumped in the pnpm-lockfile. Only the braid css hashes have changed.

I could do this in another PR but it's only two tests snapshot so I don't think it matters much 🤷

PR #1548, opened 2026-04-14
`vite`: Improve error messaging during `sku build`

askoufis:
Errors during pre-rendering are very terse and contain no stack trace. This PR addresses this by adding inline sourcemaps to the `render` bundle and enabling sourcemaps on the pre-render workers.

I used a custom error to distinguish pre-render errors from other errors because logging an error with a message and a cause needs to be handled differently than logging a regular error. Without this special handling the worker's stack trace is also logged, which isn't useful information to the user.

### DIFF
--- a/packages/sku/src/services/vite/helpers/prerender/prerenderWorker.ts
+++ b/packages/sku/src/services/vite/helpers/prerender/prerenderWorker.ts
@@ -51,7 +51,6 @@ await Promise.all(
       routeName,
       language,
     }: JobWorkerData) => {
-      // TODO: It would be nice to type this file properly
       const loadableCollector = createCollector({
         manifest,
         base: publicPath,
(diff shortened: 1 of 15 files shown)


tests/node/vite-render-error.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    await waitFor(() => {
+      expect(build.hasExit()).toMatchObject({ exitCode: 1 });
+    });
+
+    const stderr = build.stderrArr.map((item) => item.contents).join('');
+
+    expect(stderr).toMatchInlineSnapshot(`
+      "Error rendering HTML for route /Error: Home page error
+          at HomePage ({cwd}/fixtures/vite-render-error/src/HomePage.tsx:2:9)
```
williamlark, line 20: The route here doesn't look right. `/Error`?

Oh Is it missing a space between the path (`/`) and the `Error: ...` perhaps?

williamlark: Maybe add quotes around the path to seperate it better?  🤷 
`Error rendering HTML for route "/". Error: Home page error`

packages/sku/src/services/vite/helpers/prerender/prerenderWorker.ts
```diff
@@ -66,28 +65,35 @@ await Promise.all(
...
+          html = cspHandler.handleHtml(html);
+        }
+      } catch (e) {
+        throw new Error(`Error rendering HTML for route ${route}`, {
+          cause: e,
+        });
       }
 
-      await ensureTargetDirectory(filePath.split('/').slice(0, -1).join('/'));
+      await ensureTargetDirectory(path.dirname(filePath));
```
williamlark, line 92: This should be fine, but just noting these handle paths that end in a forward slash differently when I mess with it in runjs. I don't think that will ever be an issue here though? I prefer the update

tests/node/vite-render-error.test.ts
askoufis: Oh, it's because I'm joining the lines with an empty string instead of a newline 🤦. Will add the quotes too 👍.

packages/sku/src/services/vite/helpers/prerender/prerenderWorker.ts
askoufis: That's not what I'm seeing.
image

Though I think the path will always end with `.html` because it's run through this function first:
https://github.com/seek-oss/sku/blob/eadd0d3a0ac2154d63c8eb52670b84eeccfefc50/packages/sku/src/services/vite/helpers/prerender/prerenderConcurrently.ts#L39-L52

williamlark: The first value would return `foo/bar` in the original code instead, so slightly different to what path.dirname does. But yea I don't think it'll be an issue since it should always end in a file prefix. Just wanted to note it just in case

askoufis: Oh right, you meant the original code. It's always passed to `mkdir` anyway, which I think doesn't care about trailing slashes.

williamlark (approved):

PR #1551, opened 2026-04-15
`vite`: Handle error when importing sku render entrypoint

askoufis:
I noticed that there was no stack trace when an error occurs while importing the render entrypoint. This happens _before_ route rendering, so the error handling in #1548 doesn't handle it.

This PR adds error handling for this case.

### DIFF
--- a/.changeset/shiny-peas-clean.md
+++ b/.changeset/shiny-peas-clean.md
@@ -4,4 +4,4 @@
 
 `vite`: Improve error messaging during `sku build`
 
-The path of the first route that fails to render will now be logged, along with a stack trace to aid debugging. Errors that occur before rendering will also now be logged with a stack trace.
+Pre-render errors and failures in the first errored route will now be logged with stack traces for easier debugging.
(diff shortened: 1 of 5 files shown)


tests/node/vite-render-error.test.ts
```diff
@@ -4,15 +4,28 @@ import { scopeToFixture, waitFor } from '@sku-private/testing-library';
 
 const { sku } = scopeToFixture('vite-render-error');
 
+const filterNodeInternalStackFrames = (stack: string) =>
```
askoufis, line 7: Node internal stack traces aren't useful and are flaky on CI (see [here](https://github.com/seek-oss/sku/actions/runs/24431721830/job/71377401386#step:11:340)), so I've filtered them out.

.changeset/shiny-peas-clean.md
```diff
@@ -4,4 +4,4 @@
 
 `vite`: Improve error messaging during `sku build`
 
-The path of the first route that fails to render will now be logged, along with a stack trace to aid debugging
+The path of the first route that fails to render will now be logged, along with a stack trace to aid debugging. Errors that occur before rendering will also now be logged with a stack trace.
```
williamlark: Could combine the sentences 
```suggestion
Pre-render errors and failures in the first errored route will now be logged with stack traces for easier debugging.
```

williamlark (approved):

PR #1558, opened 2026-04-20
Fix vanilla-extract dep scan for aliased `css.ts` imports

williamlark:
Importing vanilla-extract css files in ts files can be done by including or excluding the `ts` extension.

```ts
// both of these import src/styles.css.ts without errors
import * as styles from '#src/styles.css';
import * as styles from '#src/styles.css.ts';
```

However, when including the ts extensions, the import gets caught by the vanilla-extract css filter that the fixViteVanillaExtractDepScanPlugin uses. If a path alias is configured, the built-in resolve to this file fails since it doesn't resolve the alias.

To fix this, we can use the rolldown resolve with `this.resolve`. This uses all the configured rolldown plugins, including the ts alias resolution. This PR adds that in.

### DIFF
--- a/fixtures/path-aliases/src/App.tsx
+++ b/fixtures/path-aliases/src/App.tsx
@@ -1,9 +1,17 @@
 import { Button } from '@components/Button';
 import { add } from 'src/utils/add';
 
+// This does not get picked up by the vanilla-extract css filter since it doesn't have a ts/js extension (e.g., "foo.css.ts")
+import * as stylesCss from '#styles/root.css';
+
+// This _does_ get picked up by the vanilla-extract css filter so we need to make sure that it resolves correctly.
+// The internal resolve (`this.resolve`) is used in that plugin to make sure it resolves aliases.
+import * as stylesCssTs from '#styles/root.css.ts';
+
 export default () => (
   <div>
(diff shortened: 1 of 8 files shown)


fixtures/path-aliases/src/App.tsx
```diff
@@ -1,9 +1,17 @@
 import { Button } from '@components/Button';
 import { add } from 'src/utils/add';
 
+// This does not get picked up by the vanilla-extract css filter since it doesn't have a ts/js extension (e.g., "foo.css.ts")
+import * as stylesCss from '#styles/root.css';
```
askoufis, line 5: To be clear, it _will_ get picked up now that we use `this.resolve` right?

williamlark: It still won't. The vanilla-extract css filter catches `css.(ext)?`, which doesn't match this (because there isn't a period at the end)

askoufis: Oh right, but that doesn't matter for non-source code because it _should_ be bundled. Maybe we should eventually enforce full file extensions for all imports?

askoufis (approved):

PR #1561, opened 2026-04-22
Fix opening on fallback port

williamlark:
Pass the port allocated at start through skuContext into the Vite dev plugins so the opened URL uses the fallback port when the default is taken.

### DIFF
--- a/.changeset/loud-insects-prove.md
+++ b/.changeset/loud-insects-prove.md
@@ -1,0 +1,5 @@
+---
+'sku': patch
+---
+
+Fix the browser not opening to the right url when a fallback port it used
--- a/.changeset/loud-insects-prove.md
+++ b/.changeset/loud-insects-prove.md
@@ -2,4 +2,4 @@
 'sku': patch
 ---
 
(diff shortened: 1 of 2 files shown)


packages/sku/src/services/vite/index.ts
```diff
@@ -31,26 +31,36 @@ export const viteService = {
...
     environment: string;
   }) => {
-    const server = await createServer(createConfig(skuContext, environment));
-
     const availablePort = await allocatePort({
       port: skuContext.port.client,
       strictPort: skuContext.port.strictPort,
     });
 
+    const skuContextOverride = {
```
williamlark, line 39: Open to better names for this

askoufis (approved):

.changeset/loud-insects-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+Fix the browser not opening to the right url when a fallback port it used
```
askoufis:
```suggestion
`vite`: Fix the browser not opening to the right url when a fallback port it used
```

PR #1563, opened 2026-04-23
Speed up compile package

williamlark:
The compilePackage search walks all of `node_modules` for `@seek` packages. It used to run synchronously on first use (then cached), which could take ~3s and block sku startup, even for commands that never need compile packages (e.g. configure).

This PR splits that into two helpers, each called only when needed:

* Async search - faster in practice (~3s to under 1s) and used wherever async is natural (e.g., Vite plugins).
* Sync search - kept for the Jest preset and sku Webpack plugin (and subsequently the Storybook plugin). Webpack doesn’t fit a simple async-plugin refactor without a larger hook rework. Since Webpack is on the way out, a dedicated sync function avoids that and is easy to delete later.

Vite users should see quicker script startup; Webpack users are unchanged or slightly better on paths that don’t touch compile packages.


packages/sku/src/config/jest/preset.ts
```diff
@@ -7,7 +7,8 @@ import type { Config } from 'jest';
 const { paths, rootResolution, jestDecorator } = getSkuContext();
 
 const slash = '[/\\\\]'; // Cross-platform path delimiter regex
-const compilePackagesRegex = paths.compilePackages
+const compilePackagesRegex = paths
+  .compilePackagesSync()
```
askoufis: Modern jest uses `import` for modules, so this preset file (compiled to .mjs) should be `import`ed, which means we can use a top-level-await with the async version of this API.

See https://github.com/jestjs/jest/blob/c24dcd42a7192d87793fa33821eeff9459bf2c3b/packages/jest-config/src/normalize.ts#L196
and https://github.com/jestjs/jest/blob/bacb7de30d053cd87181294b0c8a8576632a8b02/packages/jest-util/src/requireOrImportModule.ts#L45-L74

packages/sku/src/context/defaultCompilePackages.ts
```diff
@@ -11,96 +12,136 @@ import {
...
-        .crawl(pnpmVirtualStoreRelativePath)
-        .sync();
-
-      results.push(...pnpmVirtualStoreResults);
-    }
+  const pnpmVirtualStorePath = getPnpmVirtualStorePath();
+  if (!pnpmVirtualStorePath) {
+    cachedResult = emptyResult;
+    return cachedResult;
+  }
```
askoufis: `getPnpmVirtualStorePath` is conflating two separate bits of logic:
- Is there a `rootDir`
- Getting the pnpm virtual store path

Without looking at `getPnpmVirtualStorePath`, you'd assume this logic would always return `undefined` for `yarn` or `npm`, but it doesn't. I think the root dir check should be split out to make this clearer.

askoufis: Addressed in https://github.com/seek-oss/sku/pull/1563/commits/357a101d6d1183ffc548271a34ed347b061d3ed4

packages/sku/src/config/jest/preset.ts
askoufis: Addressed in https://github.com/seek-oss/sku/pull/1563/commits/f3ef47016febd734b10160533f716a1431089067

.changeset/fiery-eyes-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+Lazily detect compile packages
```
jahredhope, line 5: Is there value in this being two different release notes? Feels like they could be merged. That way the only release note can start with "Improve performance" and only mention Lazy loading as detail.

askoufis: Was about to make a comment on this. I split them up literally because I couldn't word them together nicely in one sentence (definitely a skill issue). IDM them being separate dot points in the release notes because I think they both have value - perf is speed, and lazy computation means it's only computed where needed instead of before every sku command.

jahredhope (approved):

askoufis:
This seems to have sped up the test suite a bit (on average, there's still some variability) 🏎️

PR #1568, opened 2026-05-01
Refactor codemods for Jest to Vitest with focused test suites

SpiralLogic:
Splits the monolithic Jest→Vitest codemod into a **phased pipeline** of focused sub-codemods (imports, hooks, mocks, generics, timeouts, etc.), adds **shared TSX/parser helpers** and package wiring, and **restructures tests** into per-codemod Vitest fixture suites with a simpler `runCodemodTests` flow (no label; no single full-pipeline jest-to-vitest test).

### Latest update 🧪
- `codemod`: phased pipeline with exported steps; shared TSX parser helpers; plop codemod output path flattening


packages/codemod/src/utils/constants.ts
```diff
@@ -10,9 +10,52 @@ export const CODEMODS = [
...
   },
+  {
+    description:
+      'jest-to-vitest: convert `as jest.Mock`/`MockedFunction` casts to `vi.mocked()`',
+    value: 'jest-to-vitest-mock-types',
+  },
+  {
+    description: 'jest-to-vitest: replace `jest.<method>` with `vi.<method>`',
+    value: 'jest-to-vitest-jest-members',
+  },
```
askoufis, line 24: `members` is a bit of a strange name for this, and I think `methods` makes more sense given the description.
```suggestion
  {
    description: 'jest-to-vitest: replace `jest.<method>` with `vi.<method>`',
    value: 'jest-to-vitest-jest-methods',
  },
```

packages/codemod/src/codemods/_shared/index.ts
askoufis, line 1: Not a fan of `_shared` or `index.ts`. The globals should also probably be scoped to specifically the `jest-to-vitest` codemod. `serializeImports` and `parseTsx` are more generic than that, so maybe they should be in their own file. I'll let you sort it out.

packages/codemod/src/codemods/jest-to-vitest.ts
```diff
@@ -1,403 +1,39 @@
...
+ */
+export const transform: Transform = async (source) => {
+  let current = source;
+  for (const step of steps) {
+    const next = await step(current);
+    if (next !== null) {
+      current = next;
+    }
+  }
+  return current === source ? null : current;
```
askoufis, line 38: I don't think it was possible for the codemod to return `null` before - it always has to return source code, either modified or unmodified. Is there any reason not to return the original code instead as an early exit? Wouldn't that simplify the transformation pipeline (i.e. just apply all codemods in order)?

packages/codemod/src/utils/constants.ts
SpiralLogic: yeah the naming is a bit conflated. From an AST perspective, they are members, however from the code I moved it was originally called methods, I like methods more as well , so will update

packages/codemod/src/codemods/jest-to-vitest.ts
SpiralLogic: In the logic doing the main transform, a null (well, any falsey value) will skip writing to the file:
https://github.com/seek-oss/sku/blob/9273963804a543d3586654658bc06382bb8220de/packages/codemod/src/transform/worker.ts#L21-L23

By returning null here, I am preventing the extra write of the file to disk when nothing has changed and in the case of the dry run you avoid showing a whole file has changed, when really it hasn't

askoufis: Oh, I should've looked at the type 🤦. Makes sense 👍

askoufis (approved):

PR #1569, opened 2026-05-04
Refactor codemods for Vitest with enhanced interactive prompts and tests

SpiralLogic:
### Latest update :sparkles:
- Split **Jest → Vitest** into focused sub-codemods and **chain** transforms per worker job.
- **Clack** flow: intro, codemod menu, **`path`** target directory, optional **dry-run confirm** when not using `-d`.
- **Dry run** prints a **unified diff** per would-change file and skips the spinner so output stays readable; **`--print`** unchanged for live runs.
- **`interactive-selection`** module + colocated **`interactive-selection.test.ts`** (typed `vi.importActual` for `@clack/prompts`); internal-only `assertClackSubmittedString`.
- **Runner/worker** refactors: `resolveCodemodModule`, diff `createTwoFilesPatch` options aligned with **diff** v8 typings.
- **Changeset** for the codemod release; **constants** / **lockfile** updates.


packages/codemod/src/transform/worker.ts
```diff
@@ -5,44 +5,56 @@ import picocolors from 'picocolors';
...
       const diff = createTwoFilesPatch(
         'source',
         'output',
         source,
-        output,
+        current,
         undefined,
         undefined,
-        { ignoreNewlineAtEof: true, context: 3 },
+        { context: 3 },
```
SpiralLogic, line 49: typescript didn't like `ignoreNewlineAtEof` for some reason I hadn't yet worked out

.changeset/cute-suits-exist.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sku-lib/codemod': minor
+---
+
+- Split jest-to-vitest codemods into individual modules for better organization.
+- Transition interactive menu implementation to Clack for improved user interaction.
```
askoufis: The first point here is implementation detail, so it shouldn't be mentioned in the release notes.

Also, we tend to avoid dot points in changesets. The typical structure is a short description of the change, with optional extra paragraphs to provide more details/usage examples/migration guides. I think the gist of this change would be:

```
Enable interactive selection of individual `jest-to-vitest` codemods. Previously all `jest-to-vitest` codemods had to be applied - now individual codemods can be applied as necessary.
```

packages/codemod/src/transform/worker.ts
askoufis: This might've been an old property that is no longer supported. We used it elsewhere until it was changed in a PR a while ago. See https://github.com/seek-oss/sku/commit/22b01536a8d1768ec50dd6934a4d99a62f7b9584#diff-695fbdd48e71c249ff10b519874fc198b8f27e662cd32cf14d1cd78f5c178284L25.

packages/codemod/src/transform/runner.ts
```diff
@@ -1,112 +1,123 @@
...
   options: Options,
 ): Promise<void> => {
-  if (options.dry) {
+  let transformerPaths: string[];
+  let showClackOutro = false;
+
+  if (transform) {
+    const codemod = CODEMODS.find((c) => c.value === transform);
+    if (!codemod) {
+      console.error('Invalid transform choice, pick one of:');
```
askoufis: Now that we're using `clack` we should use its `log` functions instead of `console` so it keeps the output inside the vertical bar/step tracker thing. I think it looks nicer this way, though not sure if it's worth getting the diff output to follow that as well.

Image

packages/codemod/src/transform/interactive-selection.ts
```diff
@@ -0,0 +1,157 @@
...
+        value: 'steps' as const,
+        label: 'Choose specific steps',
+        hint: 'Pick substeps; import synthesis runs last when needed',
+      },
+    ],
+    initialValue: 'full',
+  });
+
+  if (isCancel(pipelineMode)) {
+    exitCancel();
```
askoufis, line 105: Could use the inbuilt [`cancel` function](https://bomb.sh/docs/clack/packages/prompts/#cancel) here, or call that inside `exitCancel`.

SpiralLogic: as i understand it, cancel is what causes `pipelineMode` to return true when passed to `isCancel`. so

```
const something = cancel();
if (isCancel(something)) { exitCancel(); }
```
which in this case would always be true. putting cancel inside `exitCancel` would be too late as we are already exiting due to a cancel.

askoufis: Ohh, so it doesn't actually do the cancelling (`process.exit`), it's just for the CLI state machine or something.

packages/codemod/src/transform/runner.ts
SpiralLogic: I didn't even notice clack log until you mention it. in all the examples they seem to use console.log. I've converted it to use clack log and stream for the diff

tests/node/codemods/interactive-selection.test.ts
```diff
@@ -0,0 +1,56 @@
...
+} from '../../../packages/codemod/src/transform/interactive-selection.js';
+
+describe('getInteractiveRootCodemodSlugs', () => {
+  it('includes orchestrator and family slugs but not granular jest substeps', () => {
+    const slugs = getInteractiveRootCodemodSlugs();
+    expect(slugs).toContain('jest-to-vitest');
+    expect(slugs).toContain('transform-vite-loadable');
+    expect(slugs).toContain('svg-import-query-param');
+    expect(slugs).not.toContain('jest-to-vitest-hooks');
+    expect(slugs).not.toContain('jest-to-vitest-imports');
```
askoufis: Feel like the assertions in this file should use `toStrictEqual` to assert on the entire array. That way there's no need for negative assertions, length assertions or assertions on each array element.

tests/node/codemods/interactive-cli.test.ts
```diff
@@ -0,0 +1,135 @@
...
+};
+
+afterEach(async () => {
+  await Promise.all(
+    tmpDirs.splice(0).map((d) => fs.rm(d, { recursive: true, force: true })),
+  );
+});
+
+describe('interactive CLI', () => {
+  it('routes a non-jest codemod', async () => {
```
askoufis:
```suggestion
  it('runs a non-jest codemod', async () => {
```

askoufis (approved):
Just some minor comments, otherwise LGTM.

PR #1570, opened 2026-05-11
Update supported browsers

askoufis:
- `browserslist-config-seek` now has correct types, so the type override has been removed
- Added a browserslist converter test for our browserslist config to make sure it can handle any changes we make in the future

### DIFF
--- a/packages/sku/package.json
+++ b/packages/sku/package.json
@@ -118,7 +118,7 @@
     "babel-plugin-transform-remove-imports": "^1.7.0",
     "babel-plugin-unassert": "^3.1.0",
     "browserslist": "^4.26.2",
-    "browserslist-config-seek": "3.3.0",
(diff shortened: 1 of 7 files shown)


packages/sku/package.json
```diff
@@ -118,7 +118,7 @@
     "babel-plugin-transform-remove-imports": "^1.7.0",
     "babel-plugin-unassert": "^3.1.0",
     "browserslist": "^4.26.2",
-    "browserslist-config-seek": "3.3.0",
+    "browserslist-config-seek": "~3.4.1",
```
askoufis, line 121: I figured it made more sense to keep this dep flexible on the _patch_ version rather than fully pinned, just in case we want to roll out a bug fix without having to do a sku release.

jahredhope (approved):

PR #1574, opened 2026-05-18
Initilize `pnpm` apps with a `pnpm-workspace.yaml` file, expose pnpm config

askoufis:
Some initial work at hardening sku apps, specifically new apps use PNPM. When the create CLI is run, if PNPM is detected the `packageManager` field will be added to `package.json`, and a `pnpm-workspace.yaml` file will be initialized in the app.

The latter change necessitated exposing the pnpm config used in the plugin for programmatic use. Naturally I took the opportunity to update the config to use more of PNPM's security-related settings. [`pnpm-plugin-skuba`](https://github.com/seek-oss/skuba/blob/c22976f28d15bc26cb64b9a5b6b7dd17fa55c0a8/packages/pnpm-plugin-skuba/pnpmfile.cjs) was used as a reference, but settings were tweaked to better suit `sku`.

Additionally, `pnpm-plugin-sku` is now authored in TypeScript and built with `tsdown`. This was for 2 reasons:
- It made it easy to expose the raw config via an entrypoint
- PNPM V11 supports `.mjs` pnpm config files, so it made sense to build from TypeScript source and provide two root-level files that reference the `dist` output (PNPM config files must be at the package root)

> [!NOTE]
> While it seems redundant to both initialize a pnpm workspace config and configure a config dependency that references the same config, ongoing issues with renovate and pnpm config dependencies means this is a necessary evil for now.


packages/pnpm-plugin/src/config.ts
```diff
@@ -0,0 +1,28 @@
...
+    'sku',
+  ],
+  packageManagerStrictVersion: true,
+  publicHoistPattern: ['eslint', 'prettier'],
+  strictDepBuilds: false,
+  trustPolicy: 'off',
+  trustPolicyExclude: ['semver@6.3.1'],
+} satisfies Partial<Config>;
+// `satisfies` ensures types are correct while still emitting a narrow type with no reference to
+// `@pnpm/config`
```
askoufis, line 28: This was a happy little accident that means we don't need to depend on or bundle `@pnpm/config`, which is [a relatively large dep](https://npmx.dev/package/@pnpm/config).

tests/node/sku-create.test.ts
```diff
@@ -195,7 +196,7 @@ function replaceDependencyVersions(packageJson: Record<string, any>) {
 function stripYamlVersions(yamlContent: string): string {
   // Replace version patterns like "0.0.1+sha512-..." with "VERSION_IGNORED"
   return yamlContent.replace(
-    /:\s*[\d.]+(?:\+sha\d+-[a-f0-9]+)?.*/g,
+    /(?<!minimumReleaseAge):\s*[\d.]+(?:\+sha\d+-[a-f0-9]+)?.*/g,
```
askoufis, line 208: This regexp was picking up the `minimumReleaseAge: 4320` field, so it's now explicitly excluded.

packages/create/src/generators/pnpmWorkspace.ts
```diff
@@ -0,0 +1,21 @@
...
+    await import('pnpm-plugin-sku/config');
+
+  const doc = new Document(pnpmWorkspaceYamlConfig);
+
+  // @ts-expect-error `getIn` returns type `unknown`
+  doc.getIn(['trustPolicyExclude', 0], true).comment =
+    ' dependency of eslint-plugin-react';
+  // @ts-expect-error `get` returns type `unknown`
+  doc.get('minimumReleaseAge', true).comment = ' 3 days';
+
```
askoufis, line 16: This adds comments into the YAML AST that we can't encode in the object exposed at `pnpm-plugin-sku/config`. Hopefully this code should only exist until we can purely rely on pnpm config dependencies.

packages/create/src/generators/packageJson.ts
```diff
@@ -1,35 +1,61 @@
...
+      : {}),
   };
 
   const packageJsonPath = join(targetPath, 'package.json');
   const packageJsonContent = `${JSON.stringify(packageJson, null, 2)}\n`;
 
   await writeFile(packageJsonPath, packageJsonContent, 'utf8');
 };
+
+const resolvePackageManagerField = async (
```
askoufis, line 45: I chose not to add support for `npm` or `yarn` just yet. They may come in a future PR.

packages/pnpm-plugin/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+export default {
+  hooks: {
+    updateConfig(config: object) {
+      Object.entries(defaultConfig).forEach(([key, value]) => {
+        if (
+          typeof value === 'string' ||
+          typeof value === 'number' ||
+          typeof value === 'boolean'
+        ) {
+          // @ts-expect-error
```
SpiralLogic, line 14: do these need a description ?

askoufis: We haven't made descriptions mandatory in `eslint-config-seek`, so no. We should probably add our own rule for it in sku though.

packages/pnpm-plugin/package.json
```diff
@@ -1,11 +1,25 @@
...
+    "./config": {
+      "import": "./dist/config.mjs",
+      "require": "./dist/config.cjs"
+    },
+    "./package.json": "./package.json"
+  },
   "files": [
-    "pnpmfile.cjs"
+    "pnpmfile.cjs",
+    "pnpmfile.mjs"
```
SpiralLogic: I think this needs to include dist as well

packages/pnpm-plugin/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+          config[key].push(...value);
+
+          return;
+        }
+
+        if (typeof value === 'object' && value !== null) {
+          // @ts-expect-error
+          config[key] ??= {};
+          // @ts-expect-error
+          Object.assign(config[key], value);
```
SpiralLogic, line 33: if config[key] already has a key that exists in defaultConfig it will overwrite it with the default, is that intended?

as opposed to: `Object.assign(config[key], { ...value, ...config[key] })`

askoufis: `Object.assign` already does what is effectively a spread right?

tests/node/__snapshots__/sku-create.test.ts.snap
```diff
@@ -103,6 +103,7 @@ exports[`create-sku vite > should create package.json 1`] = `
     "vitest": "VERSION_IGNORED",
   },
   "name": "new-project",
+  "packageManager": "pnpm@10.33.4",
```
SpiralLogic: will this break the snapshot when a different pnpm version runs on build?

askoufis: No, but I guess it will break if a new pnpm v10 version is released. Might bump this up to v11 soon (maybe in this PR, was waiting for v11 to be more widely used), which will have the same problem, so probably worth replacing it with `VERSION_IGNORED` or maybe just the major version.

packages/pnpm-plugin/src/index.ts
SpiralLogic: Image

the first one keeps the default config, the second overwrites the default with the user config

The code will use the default config for objects rather than the config provided as the argument to update config

askoufis: Hmm, maybe it's intentional in this case because you'd want the user (consumer of the pnpm plugin) to be able to preferentially set properties in their `pnpm-workspace.yaml` knowing that they won't be overridden by a plugin.

SpiralLogic (approved):

PR #1575, opened 2026-05-20
Create semantic text styling utils, replace `chalk` dep with native `styleText`

askoufis:
## Problem

This PR was prompted by an error in braid's storybook:

```
TypeError: highlight is not a function
│  at banner
│  (/Users/askoufis/code/braid-design-system/node_modules/.pnpm/sku@15.15.1_5a2b91c15c68fb5f5611153b65ee0692/node_modules/sku/dist/config/storybook.cj
s:1169:17)
```

The `highlight` function is here:
https://github.com/seek-oss/sku/blob/765ca72bd657fb625cc83a392f6b2919a7314416/private/utils/src/console/banner.ts#L14

I think the error is being caused by the following sequence of events:
- sku's storybook entrypoint is CJS, because historically storybook didn't like ESM config (no longer an issue as of v9-ish)
- It eventually calls the `banner` function from our private sku utils package which gets inlined into the entrypoint as a result
- `banner` imports `chalk` which is ESM-only
- The entrypoint is CJS, but since `chalk` is ESM, require(esm) kicks in, and the `require` returns an object with a `default` property that doesn't get destructured correctly (because the assumption in sku is that everything is ESM, which is wrong in only this specific case), causing the error

## Fix

To fix this, we _could_ build an ESM entrypoint in addition to the CJS one for the storybook config. In fact we were planning to (and eventually will) replace the CJS entrypoint with an ESM entrypoint in the next major version. However, a better fix that was also something on our todo list would be to remove `chalk` entirely in favour of the native [`styleText`](https://nodejs.org/api/util.html#utilstyletextformat-text-options) util. So I did that.

We now have semantic console color utils in the private utils package under their own `@sku-private/utils/console` entrypoint. The utils are named semantically, rather than just by colour. There's also a little bit of consolidation too, e.g. some `red bold` styles are now just `red` (via `critical`), and `dim` usage has been replaced with `gray` (via `secondary`). I felt like these were subtle changes that are worth doing for the sake of consistency.

_Maybe_ there's some use cases for having a `dim` or `deemphasized` tone as well as a `secondary` tone, but they're similar enough that just one "less bright neutral" tone should be fine.

However, both `info` and `accent` render `blue` text right now, and I felt that changing one of them in this PR was a bit too disruptive. Also I haven't put much thought into which one to change or what to change it to, so that might be a followup. _Maybe_ `info` should be `whiteBright` or something. Regardless, we're in a much better position to make broad changes to console styling, and eventually build some better abstractions on top of the existing stuff we have (IMO `banner` is a bit ugly).

> [!NOTE]
> Tested a snapshot in braid and it works.


packages/sku/src/program/commands/build-ssr/build-ssr.action.ts
```diff
@@ -82,7 +82,7 @@ export const buildSsrAction = async ({
       csp: cspEnabled,
     });
 
-    console.error(chalk.red(error));
+    console.error(critical(String(error)));
```
askoufis, line 85: `chalk` is more permissive with its input type compared to our utils. We only accept `string`, so some coercion will be necessary here and there.

packages/sku/src/utils/allocatePort.ts
```diff
@@ -9,7 +9,7 @@ const allocatePort = async ({
   host,
   strictPort = false,
 }: {
-  port?: number;
+  port: number;
```
askoufis, line 12: No existing usage of `allocatePort` passes an undefined `port`, so this type is now stricter.

private/utils/src/console/managedConfigBanner.ts
askoufis, line 1: I think this module existed because something else used `prependWithManagedConfigBanner`, but right now it's only used in one place so it has been consolidated with that usage.

eslint.config.js
```diff
@@ -90,7 +90,7 @@ export default [
             {
               group: ['**/*.action.js'],
               message:
-                'Please dynamically import action modules to keep the CLI fast',
+                'Please dynamically import action modules to keep the CLI fast.',
```
askoufis, line 93: These messages are interpolated after a sentence with a full stop, so it looks weird if the message doesn't also have a full stop.

packages/sku/src/program/commands/test/vitest-test-handler.ts
```diff
@@ -48,17 +48,16 @@ const lazyLoadVitest = async () => {
...
-        styleText(
-          'red',
+        critical(
           `Could not find ${vitest}. Please install it in your project's devDependencies.`,
         ),
       );
       process.exit(1);
     }
-    throw error;
+    throw critical;
```
michaeltaranto: Why is this no longer throwing the error?

private/utils/src/console/banner.ts
```diff
@@ -1,37 +1,28 @@
...
-      break;
-    }
-    default: {
-      throw new Error(`Banner type not implemented: ${type}`);
-    }
+  const highlightMap = {
+    info,
+    critical,
+    caution,
+  };
```
michaeltaranto: Given `highlightMap` is static, you can hoist it out of `banner` and also extract its keys for the `type` itself to keep them in sync.

```ts
const highlightMap = {
  info,
  critical,
  caution,
};

export const banner = (
  type: keyof typeof highlightMap,
  heading: string,
  messages: string[] = [],
) => {
```

packages/sku/src/program/commands/test/vitest-test-handler.ts
askoufis: Bad rename I think 😅. Originally had `critical` named as `error`.

michaeltaranto (approved):

PR #1580, opened 2026-05-26
Lint with `knip`

askoufis:
Last time I tried this I kept hitting internal errors in `knip`, but this I got it working!

Main changes involved:
- Adding some missing deps
- Removing unused deps
- Adding scripts to fixtures so `sku` is flagged as a used dep
- Replacing import path aliases like `src/App` with `./App` where appropriate (i.e. where we aren't explicitly testing resolution of these aliases)

I've had to ignore a few unresolved paths/unused dep errors caused by the fact that `tsconfig.json` files in fixtures are gitignored, so they only exist in CI during `test` which happens _after_ `lint`, which means tsconfig paths don't resolve. Eventually we'll replace most (maybe all?) of these paths with package imports, but until then we have to live with some ignores here and there.

Currently `knip` is only linting deps, but will try turn on unused code/exports in a future PR.


knip.jsonc
```diff
@@ -0,0 +1,189 @@
+{
+  "$schema": "https://unpkg.com/knip@6/schema-jsonc.json",
+  "tags": ["-knipignore"],
```
askoufis, line 3: `knip`'s init script adds a `lintignore` tag by default, but I like the more explicit `knipignore` so you can know what tool the ignore is used for.

knip.jsonc
```diff
@@ -0,0 +1,189 @@
...
+      // Dep is referenced as a string
+      "ignoreDependencies": ["@babel/preset-react"],
+    },
+    "tests": {
+      "entry": ["**/*.test.ts"],
+      "project": ["**/*.ts"],
+      // Used in tests
+      "ignoreDependencies": ["webpack", "webpack-dev-server", "webpack-cli"],
+    },
+    "fixtures/assertion-removal": {
```
askoufis, line 37: A blanket `src/render.*`, `src/client.*`, `sku.config.*` for all fixtures _sort of_ works, but you get warnings in cases where fixtures don't have them. Once we consolidate onto vite and migrate all fixtures to typescript we can probably revisit this config to try and reduce duplication.

SpiralLogic (approved):
does this one need a changeset?

askoufis:
> does this one need a changeset?

Don't think so - only dev deps have changed.

PR #1583, opened 2026-05-29
`webpack`: Improve CSS minification and autoprefixing tests

askoufis:
Split these changes out from a separate PR that might not go anywhere.

We don't _really_ test our minification explicitly because our snapshots format the CSS, so any whitespace minification ends up reverted. I've added a bit more CSS to the `styling` fixture so we have coverage on a few CSS bundling features sku enables:
- Minification: `#0000ff` -> `#00f`
- Autoprefixing: `height: stretch`
- Disabled `calc` minification: minifiers don't handle certain cases. This was disabled long ago in https://github.com/seek-oss/sku/pull/589, but now we explicitly test for it and have a bit more context as to _why_ we turn it off.

### DIFF
--- a/packages/sku/src/services/webpack/config/utils/loaders.ts
+++ b/packages/sku/src/services/webpack/config/utils/loaders.ts
@@ -56,6 +56,11 @@ export const makeExternalCssLoaders = (
                   cssnano({
                     preset: [
                       'default',
+                      // The calc optimizer can make incorrect assumptions, so we disable it.
+                      // See https://github.com/postcss/postcss-calc/issues/238.
+                      //
+                      // Turning off calc optimizations is currently not possible in lightningcss.
+                      // See https://github.com/parcel-bundler/lightningcss/issues/12.
                       {
                         calc: false,
                       },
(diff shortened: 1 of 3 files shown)


packages/sku/src/services/webpack/config/utils/loaders.ts
```diff
@@ -56,6 +56,11 @@ export const makeExternalCssLoaders = (
                   cssnano({
                     preset: [
                       'default',
+                      // The calc optimizer can make incorrect assumptions, so we disable it.
+                      // See https://github.com/postcss/postcss-calc/issues/238.
+                      //
+                      // Turning off calc optimizations is currently not possible in lightningcss.
```
askoufis, line 62: The PR mentioned in the description was to move to lightningcss for webpack, but unfortunately calc optimization can't be disabled (yet), so I've parked that branch for now. 

`vite` already uses lightningcss, so it would allow us to remove the `cssnano` and `autoprefixer` deps, resulting in a net decrease of 38 deps in `sku`'s dep tree.

SpiralLogic (approved):

PR #1584, opened 2026-06-01
Update polyfill registry

askoufis:
Adding a few more polyfill deps I've seen here and there.


SpiralLogic (approved):

packages/sku/src/utils/polyfillRegistry.ts
```diff
@@ -58,4 +66,8 @@ export const POLYFILL_REGISTRY: PolyfillRegistry = {
     reason:
       'requestAnimationFrame is well established and works across many devices and browser versions. It has been available across browsers since July 2015.',
   },
+  'url-search-params-polyfill': {
+    reason:
+      'URLSearchParms is well established and works across many devices and browser versions. It has been available across browsers since April 2018.',
```
SpiralLogic:
```suggestion
      'URLSearchParams is well established and works across many devices and browser versions. It has been available across browsers since April 2018.',
```

PR #1585, opened 2026-06-02
`vite`: Fix loadable css ordering bug

askoufis:
This PR fixes a bug that could cause stylesheets to be ordered incorrectly in static HTML. In some cases this resulted in styling issues caused by improper positioning of the Braid reset styles (i.e. not first).

The culprit was our vite loadable implementation - we were registering chunk assets _before_ traversing into them, resulting in a backwards dependency order. The fix was relatively simple - recurse into the chunk's imports first so we emit dependencies in the correct order. Also added a list of seen chunks to prevent duplicate work and potential infinite recursion.

The added fixture was initially a reproduction of the bug - it seems to only manifest when an app has both a loadable component in its source code and in a dependency. There may be some more nuance to this (i.e. shared styles between them via braid), but that's the general situation.

I've validated the fix in an internal repo where the bug originally occurred and it fixed the chunk order.

### DIFF
--- a/fixtures/vite-loadable-css-order/src/client.tsx
+++ b/fixtures/vite-loadable-css-order/src/client.tsx
@@ -1,0 +1,14 @@
+import { hydrateRoot } from 'react-dom/client';
+import { BrowserRouter } from 'react-router';
+
+import App from './App';
+import type { RouteName } from './routes';
+
+export default ({ route }: { route: RouteName }) => {
+  hydrateRoot(
+    document.getElementById('app')!,
+    <BrowserRouter>
+      <App route={route} />
+    </BrowserRouter>,
+  );
+};
--- a/fixtures/vite-loadable-css-order/src/client.tsx
+++ b/fixtures/vite-loadable-css-order/src/client.tsx
@@ -1,14 +1,13 @@
 import { hydrateRoot } from 'react-dom/client';
 import { BrowserRouter } from 'react-router';
 
-import App from './App';
-import type { RouteName } from './routes';
+import { App } from './App';
 
-export default ({ route }: { route: RouteName }) => {
+export default () => {
   hydrateRoot(
     document.getElementById('app')!,
     <BrowserRouter>
-      <App route={route} />
+      <App />
     </BrowserRouter>,
   );
 };
--- a/private/test-utils/package.json
+++ b/private/test-utils/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@sku-private/test-utils",
   "private": true,
+  "version": "0.0.0",
   "type": "module",
   "main": "index.ts",
   "exports": {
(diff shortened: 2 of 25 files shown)


tests/browser/__snapshots__/translations.test.ts.snap
```diff
@@ -18,12 +18,12 @@ exports[`translations > bundler vite > should render en 1`] = `
      >
      <link
        rel="modulepreload"
-       href="/vite-client.js"
+       href="/rolldown-runtime.js"
```
askoufis, line 21: A few fixtures have had these `link` ordering changes. They're benign, and likely more correct as they better reflect dependency order.

tests/node/vite-loadable-css-order.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    );
+    const restIndex = cssContents.findIndex(({ href }) =>
+      href.includes('vite-client'),
+    );
+
+    expect(resetIndex).toBeGreaterThanOrEqual(0);
+    expect(restIndex).toBeGreaterThanOrEqual(0);
+
+    // The reset must load before other styles
+    expect(resetIndex).toBeLessThan(restIndex);
```
askoufis, line 52: This is basically the whole point of the test: the stylesheet containing the CSS reset should be before any other stylesheets.

private/test-utils/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@sku-private/test-utils",
   "private": true,
+  "version": "0.0.0",
```
askoufis, line 4: This fixes a bug that was preventing the create package from being deployed - see [here](https://github.com/seek-oss/sku/actions/runs/26798712128/job/79000610676#step:8:34). Seems like PNPM doesn't like when a workspace dep doesn't have a version and that dep is referenced by a public package.

Haven't attempted to find a pnpm issue about this because the error is quite bad. If I don't find one I'll probably raise an issue.

fixtures/vite-loadable-css-order/src/client.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { BrowserRouter } from 'react-router';
+
+import App from './App';
+import type { RouteName } from './routes';
+
+export default ({ route }: { route: RouteName }) => {
+  hydrateRoot(
+    document.getElementById('app')!,
+    <BrowserRouter>
+      <App route={route} />
```
SpiralLogic: app doesn't look like it takes a route prop, they are imported in the app component file drectly

askoufis: Ah yep, leftover from an older version of the fixture. Fixed in https://github.com/seek-oss/sku/pull/1585/commits/a50d28dcfbcc038a31866acc05dd585818836b02.

SpiralLogic (approved):

PR #1586, opened 2026-06-03
Configure pnpm v11 instead of v10, update pnpm config

askoufis:
PNPM v11 has had enough releases and bug fixes that I'm now more confident in initializing new apps with it rather than v10. I've updated the pnpm config a bit, and tweaked the changeset added in #1574 since that change is still unreleased.


askoufis:
Not sure what's causing all the test failures. Seems like the CLI is getting stuck maybe? I guess that would eventually mean prettier doesn't run so quotes aren't correct. 🤷

askoufis:
Turns out the issue was caused by using pnpm v11 in the `sku-create` fixture, but still using v10 at the repo-level. Updating to v11 in the repo seems to have fixed most of the issues. The PR is now much larger, but most of that is snapshot updates from all the dep shuffling/deduping.

I also had to pass through the environment to the `sudo` command because otherwise PNPM doesn't know it's running in CI, causing an error when it presents an interact prompt.

Finally I had to update our pnpm config dep detection logic to handle pnpm v11's CLI output.

tests/node/sku-create.test.ts
```diff
@@ -119,7 +119,7 @@ describe.each(['webpack', 'vite'])('create-sku %s', (template) => {
 
     expect(await result.findByText('❯ Vite')).toBeInTheConsole();
 
-    result.userEvent.keyboard('[Enter]');
+    await result.userEvent.keyboard('[Enter]');
```
askoufis, line 122: `keyboard`'s return type is `void | Promise<void>`. This was a failed attempt to fix test failures, but I still think the change is correct.

.github/workflows/validate.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
       - name: Set up hosts
         # Couldn't get `sudo pnpm setup-test-hosts` to work
-        run: sudo ${{ steps.pnpm-setup.outputs.bin_dest }}/pnpm setup-test-hosts
+        run: sudo -E ${{ steps.pnpm-setup.outputs.bin_dest }}/pnpm setup-test-hosts
```
askoufis, line 56: `sudo` is running `pnpm`, which now needs to know if it's in a CI environment in some cases. The `-E` passes through the existing environment to `sudo`, in particular the `CI: true` envar set in this workflow.

SpiralLogic:
There are still a few mentions of pnpm 10:
https://github.com/seek-oss/sku/blob/a5b0195c6d85632bcd43be89b0aeb23cb717b9b9/packages/sku/src/services/packageManager/pnpmConfig.ts#L16-L19

should that be pnpm 11?

SpiralLogic (approved):

askoufis:
> There are still a few mentions of pnpm 10:
> 
> https://github.com/seek-oss/sku/blob/a5b0195c6d85632bcd43be89b0aeb23cb717b9b9/packages/sku/src/services/packageManager/pnpmConfig.ts#L16-L19
> 
> should that be pnpm 11?

Nope, that's a minimum version check and should stay the same.

PR #1591, opened 2026-06-11
removing init command

williamlark:
Removing the `init` command completely in the next major version. It's been deprecates since the v15 release and will be removed entirely in v16. The assumption is that if people need to create a new project and wonder why  `sku  init` isn't working then the sku site or github is the first place they would go if they run into issues, which shows the new command.


williamlark:
Closing for now till we're ready to start reviewing v16 PRs

PR #1602, opened 2026-06-29
Enable subpath imports and deprecate cjs config files

williamlark:
sku currently handles import alias (e.g., importing from `src/*`) with a vite and webpack plugin. Node can handle this natively now with subpath imports. This PR removes the plugins we were using and leverages subpath imports  for aliases now.

The `pathAliases` option is now available for both webpack and vite (previously only vite). This option will now add the entries into the tsconfig `paths` option so editors can pick up the imports, and also updates the package.json `imports` option with the paths. Handling the package.json field is quite aggressive since we will overwrite any `imports` entries with whatever is in `imports`, but it allows the least amount of friction for people to add new aliases.

I've also created a codemod that changes all `src/*` imports to `#src/*` and adds the src path alias to sku.config. The template for the `create` package has also been  updated so projects will include a `src` path alias by default. Users can opt out of the src alias simply by deleting that entry in their sku.config.

Finally, I've deprecated the cjs format of sku.config files (configs using `module.exports`). Handling cjs config files would have added more complexity to the new codemod, and the ESM file should be an easy change that any project can do, so we agreed it was safe to deprecate the cjs format.


.pnpm-store/v11/index.db
SpiralLogic, line 1: did you mean to add this?

packages/sku/src/context/createSkuContext.ts
```diff
@@ -113,32 +92,7 @@ export const createSkuContext = ({
 
   validateConfig(skuConfig);
 
-  if (isCompilePackage && skuConfig.rootResolution) {
```
SpiralLogic, line 112: is this check no longer needed. it doesn't seem to be moved to where this logic lives now

.changeset/remove-root-resolution.md
```diff
@@ -0,0 +1,29 @@
+---
+'sku': major
+---
+
+Remove `rootResolution` in favour of native subpath imports
+
+Node.js can now resolve path aliases natively via [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) declared in your `package.json` `imports` field. Subpath import specifiers must be prefixed with `#`, so root-resolution imports like `import x from 'src/foo'` become `import x from '#src/foo'`.
+
+The [`pathAliases`](https://seek-oss.github.io/sku/#/./docs/configuration?id=pathaliases) option is now available for both the `webpack` and `vite` bundlers. It generates TypeScript's `paths` configuration for editor/type support, and `sku` mirrors it into your `package.json` `imports` field so the aliases resolve natively at build time.
```
askoufis:
```suggestion
The [`pathAliases`](https://seek-oss.github.io/sku/#/./docs/configuration?id=pathaliases) option is now available for both the `webpack` and `vite` bundlers. `sku` now generates `tsconfig.json#paths` configuration so TypeScript can resolve these imports, and `sku` mirrors it into your `package.json` `imports` field so the aliases resolve natively at build time.
```

.changeset/remove-root-resolution.md
```diff
@@ -0,0 +1,29 @@
+---
+'sku': major
+---
+
+Remove `rootResolution` in favour of native subpath imports
+
+Node.js can now resolve path aliases natively via [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) declared in your `package.json` `imports` field. Subpath import specifiers must be prefixed with `#`, so root-resolution imports like `import x from 'src/foo'` become `import x from '#src/foo'`.
```
askoufis, line 7: I think the `#<field>` notation is common enough these days that we can use it to talk about specific fields in a JSON file:

```suggestion
Node.js can now resolve path aliases natively via [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) declared in your `package.json#imports` field. Subpath import specifiers must be prefixed with `#`, so root-resolution imports like `import x from 'src/foo'` become `import x from '#src/foo'`.
```

.changeset/remove-root-resolution.md
```diff
@@ -0,0 +1,29 @@
...
+
+The [`pathAliases`](https://seek-oss.github.io/sku/#/./docs/configuration?id=pathaliases) option is now available for both the `webpack` and `vite` bundlers. It generates TypeScript's `paths` configuration for editor/type support, and `sku` mirrors it into your `package.json` `imports` field so the aliases resolve natively at build time.
+
+#### Migration
+
+Run the `migrate-root-resolution` codemod to update your project. It rewrites `src/…` imports to `#src/…` and adds the matching `#src/*` entry to the `pathAliases` option in your sku config:
```
askoufis:
```suggestion
Run the `migrate-root-resolution` codemod to update your project. It rewrites `src/...` imports to `#src/...` and adds the matching `#src/*` entry to the `pathAliases` option in your sku config:
```

tests/node/__snapshots__/lint-format.test.ts.snap
```diff
@@ -23,6 +21,8 @@ import myself from '.'; // index
 
 import vanillaStyles from './vanillaStyles.css'; // styles
 
+import utils from '#src/utils'; // internal
+
```
askoufis, line 25: This will cause a lot of formatting churn. We probably want to customize `pathGroups` in the `import-x/order` rule to group `#`-prefixed paths into the `internal` group. See https://github.com/un-ts/eslint-plugin-import-x/blob/master/docs/rules/order.md#pathgroups.

packages/sku/src/utils/pathAliasImports.ts
```diff
@@ -0,0 +1,51 @@
+import { readFile, writeFile } from 'node:fs/promises';
+
+import { getPathFromCwd } from '@sku-private/utils';
+import { applyEdits, modify } from 'jsonc-parser';
```
askoufis: `jsonc-parser` feels like overkill to me, especially because we're only using it to parse regular JSON. Is it really necessary?

packages/sku/src/services/typescript/tsconfig.ts
```diff
@@ -40,14 +59,11 @@ export default ({ rootResolution, tsconfigDecorator, tsPaths }: SkuContext) => {
...
-          }
-        : {}),
       ...(tsPaths ? { paths: tsPaths } : {}),
     },
   };
 
   return tsconfigDecorator(config);
 };
+
+export default createTSConfig;
```
askoufis:
```suggestion
```

packages/sku/src/context/createSkuContext.ts
askoufis: I think so. Now that Node will be handling path aliases, apps that consume a compile package (i.e. published source code) should be able to resolve these paths.

packages/sku/src/config/eslint/importOrder.ts
```diff
@@ -1,21 +1,6 @@
...
-    'import-x/internal-regex': internalRegex(paths),
-  },
-});
-
-export const createImportOrderConfig = ({
-  paths,
-  rootResolution,
-}: SkuContext): Linter.Config => ({
-  ...(rootResolution ? rootResolutionConfig(paths) : undefined),
+export const createImportOrderConfig = (): Linter.Config => ({
```
askoufis: This can just be static config now, no need for it to be a function anymore.

packages/sku/src/utils/validateSkuConfigFormat.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import {
+  isCjsSkuConfig,
+  validateSkuConfigFormat,
+} from './validateSkuConfigFormat.js';
+
+const { readFileSync } = vi.hoisted(() => ({ readFileSync: vi.fn() }));
+
+vi.mock('node:fs', () => ({ readFileSync }));
+
```
askoufis: Can we use `fs-fixture` here instead?

packages/sku/src/types/types.ts
```diff
@@ -538,25 +541,6 @@ export interface ViteSkuConfig {
...
 
-  /**
-   * Path alias mappings for module resolution.
-   * Each alias maps a pattern to a destination path relative to the project root.
-   *
-   * This configuration affects both bundler module resolution and TypeScript's
-   * `paths` configuration in tsconfig.json.
-   *
-   * Note: sku automatically provides a 'src/*' alias that maps to './src/*'.
-   * This option allows you to define additional custom aliases.
```
askoufis, line 549: This note wasn't kept in the move, but I think it should be.

packages/sku/src/services/vite/helpers/config/createConfig.ts
```diff
@@ -49,12 +44,6 @@ export const createConfig = (
   }
 
   return {
-    resolve: {
-      // ! vite v8+ supports tsconfigPaths out of the box, however we need to use the vite-tsconfig-paths plugin instead for vitest v3 support.
-      // ! Once we drop support for vitest v3, we can remove this plugin and use the built-in tsconfigPaths support.
-      // ! This will need to be added at the top level so that vanilla-extract picks it up. VE doesn't inherit config options from plugins at the moment (unless whitelisting the entire plugin)
-      // tsconfigPaths: true,
```
askoufis, line 56: Do we test that a tsconfig paths resolve correctly in VE files?

tests/node/__snapshots__/lint-format.test.ts.snap
askoufis: Actually, skuba didn't do anything to address this, and I'd rather not deviate on import ordering unless we have to. Churn in a major version is expected, and at least it's auto-fixable, so don't worry about this.

.pnpm-store/v11/index.db
williamlark: Nope, thank you!

packages/sku/src/utils/pathAliasImports.ts
williamlark: Yea I gave a little test and JSON parse/stringify seems to be enough. We control the prettier settings so the formatting we do with JSON.stringify seems to be enough to not have to run `format` whenever this adds entries (apart from the first time so it fixes the key ordering, but that's acceptable)

packages/sku/src/types/types.ts
williamlark: I removed it since I didn't think it was correct anymore. We do add the `src/*` alias in our create template, but sku doesn't automatically add it itself, it needs to be explicit. I thought that since it's now visible in the sku.config (and no longer magic behind the scenes) that it didn't need to be mentioned in the types.

askoufis: Ah ok, I got confused. I agree it's better that the config is now explicit 👍.

packages/sku/src/services/vite/helpers/config/createConfig.ts
williamlark: Tested and it works. I've added css and ts subpath imports into the styling fixture to make sure things work too.

.changeset/remove-root-resolution.md
askoufis, line 1: Was going to nitpick this changeset a bit - it doesn't follow the typical structure we use, but we'll likely rewrite the full release notes anyway so I'll save feedback until then.

.pnpm-store/v11/index.db
askoufis, line 1: This got added again? Feel free to add `.pnpm-store` to gitignore.

docs/docs/configuration.md
```diff
@@ -398,32 +398,45 @@ Type: `Record<string, string>`
 
 Default: `{}`
 
-Bundler: `vite`
-
 Custom path alias mappings for module resolution. Each alias pattern maps to a destination path relative to the project root.
 
-Sku automatically provides a `src/*` alias that maps to `./src/*`. The `pathAliases` option allows you to define additional custom aliases.
+This option generates `tsconfig.json#paths` so TypeScript can resolve these imports, and `sku`
+mirrors it into your `package.json#imports` field so the aliases resolve natively at build time.
```
askoufis:
```suggestion
This option generates `tsconfig.json#paths` so TypeScript can resolve these imports, and `sku` mirrors it into your `package.json#imports` field so the aliases resolve natively at build time.
```

docs/docs/configuration.md
```diff
@@ -398,32 +398,45 @@ Type: `Record<string, string>`
...
+
+Subpath import specifiers must be prefixed with `#`.
 
-**Note**: For Vite projects, `pathAliases` replaces the need for `rootResolution` configuration used in Webpack.
+> [!WARNING]
+> Because `sku` fully manages the `imports` field, any entries you add manually will be removed. Declare all of your subpath imports via `pathAliases` instead.
```
askoufis:
```suggestion
> Because `sku` fully manages the `imports` field, any entries you add manually will be removed.
> Declare all of your subpath imports via `pathAliases` instead.
```

askoufis (approved):

askoufis (approved):

PR #1605, opened 2026-06-29
Remove pre commit

williamlark:
Removing the pre-commit sku command. Users already needed to do some manual work to setup pre-commit hooks since sku removed husky, and monorepo projects may use their own tools instead of lint-staged, so bundling lint-staged for every project is unnecessary. The pre-commit docs have been updated so users can still set it up, but also gives them flexibility to set it up how they like.

Also, the sku monorepo has moved from lint-staged to nano-staged since it's a smaller, newer package, and the sku docs have now recommended it instead of lint-staged.


package.json
```diff
@@ -76,5 +75,8 @@
   "packageManager": "pnpm@11.5.1",
   "skuSkipPostInstall": true,
   "skuSkipConfigure": true,
-  "skuSkipValidatePeerDeps": true
+  "skuSkipValidatePeerDeps": true,
+  "dependencies": {
+    "nano-staged": "catalog:"
+  }
```
askoufis: Should move this to `devDependencies` so it doesn't feel lonely 😢

docs/docs/extra-features.md
```diff
@@ -231,24 +231,41 @@ A report is generated in the `/report` directory when `sku build` is run.
 
 ## Pre-commit hook
 
-To speed up the feedback loop on linting and formatting errors, `sku` provides a `pre-commit` script that can be run to catch simple problems before CI.
-To make use of this hook, it's recommended to install [husky](https://www.npmjs.com/package/husky) as a development dependency and configure it as follows:
+> **Deprecation notice:** The built-in `sku pre-commit` command has been removed. It saw very little use and bundled `lint-staged` (and its many transitive dependencies) into every `sku` install. Setting this up yourself is straightforward and gives you full control over which commands run before each commit.
```
askoufis: Feels weird to call this deprecation when it has actually just been removed entirely (and without warning). I'd rather just be blunt about it:
```suggestion
> The `sku pre-commit` command was removed in v16. It saw very little use and bundled `lint-staged` (and its many transitive dependencies) into every `sku` install. Setting this up yourself is straightforward and gives you full control over which commands run before each commit.
```

askoufis (approved):

package.json
williamlark: What a sneaky guy. He almost had his own space 👾

PR #1607, opened 2026-06-29
Replace `cross-spawn` with `tinyexec` and linting changes

williamlark:
This PR also changes the way  linting works in two ways:
*  Sku lint no longer bails on the first failing check and now reports all errors in one pass, still exiting non-zero if any are found.
* Skips the TypeScript check when linting specific file paths. Because TypeScript ignores tsconfig.json when explicit filenames are passed (making it impossible to scope to one path), `sku lint <path>` now skips the TypeScript check entirely so it only lints the targeted path.

Also replace `cross-spawn` with `tinyexec` to reduce dependancies. Changing this dep made me find the above changes.


packages/sku/src/services/prettier.ts
```diff
@@ -45,36 +45,35 @@ const runPrettier = async ({
...
+    return;
+  }
+
+  if (listDifferent && exitCode === 1) {
+    console.error(
+      critical('Error: The file(s) listed above failed the prettier check'),
+    );
+    suggestScript('format');
+    return;
   }
```
askoufis, line 77: Don't we still need a check for a non-zero exit code?

williamlark: Ah yes, I  added a zero check for an early return

askoufis: But what about the other exit codes? I guess what I'm saying is before we had a catch-all, but now there's only a log as a fallback - an error we don't handle will just silently fail (I think).

private/testing-library/utils.ts
```diff
@@ -1,20 +1,36 @@
...
     }
   });
 };
+
+// TODO refactor the tests to use this in expect. e.g., expect(hasExitCode(process, 0)).toBe(true)
+/**
+ * Checks if a command exited with the given code.
+ * Useful in tests since it will output the debug information if it fails.
+ */
+export const hasExitCode = (
```
askoufis: nit: just a naming suggestion.

```suggestion
export const hasExpectedExitCode = (
```

tests/node/lint-format.test.ts
```diff
@@ -1,114 +1,321 @@
...
-      const lint = await sku('lint');
-      expect(
-        await lint.findByText('src/typescriptFile.ts(1,7): error'),
-      ).toBeInTheConsole();
+    describe('when there are no lint errors', () => {
+      let lint: RenderResult;
+      let fixture: Fixture;
+
+      beforeAll(async () => {
+        fixture = await createSrcFixture({ 'passing.ts': passingFile });
```
askoufis: I don't know if it still works with the wrapping `createSrcFixture`, but you could use `using` here to automatically clean up the fixture.

.changeset/pink-kiwis-boil.md
```diff
@@ -0,0 +1,9 @@
+---
+'sku': minor
+---
+
+`lint|format`: Run all lint checks instead of bailing on the first failure
+
+Previously, a single failing lint check would halt the entire process, so you'd only see errors from the first check to fail and would have to fix them before any other errors surfaced.
+
+To surface everything at once, `sku` now runs all lint checks even when some fail. You'll see errors from every check in a single run, and the process still exits with a non-zero code if any errors are found.
```
askoufis: Not sure if `format` should be mentioned in this changeset. It was changed in a different way that I don't even think is worth mentioning because it's kind of subtle and shouldn't matter in most cases, but even if it did need a changeset I think it should be separate.

tests/node/lint-format.test.ts
williamlark: I think it would, but it won't work here since I'm declaring it in the upper scope with let.

askoufis: You could hoist the fixture creation out of the `beforeAll`, and then you _could_ use `using`, but if it's too much hassle then don't worry about it.

williamlark: After trying that I realised that we don't even need the files during the test itself, so I can create and clean up directly in the beforeEach. I've used `await using` in the places that use the var, and cleaned up manually for those that don't for clarity

.changeset/pink-kiwis-boil.md
williamlark: I've split format out to it's own changeset

.changeset/hip-jobs-bathe.md
```diff
@@ -0,0 +1,17 @@
...
+
+Since linting a path should only lint that path, `sku` now skips the TypeScript check entirely when a file path is provided. To type-check as well, run `tsc` manually first:
+
+```json
+{
+  "scripts": {
+    "lint-path": "tsc && sku lint <path>"
+  }
+}
+```
```
askoufis: Would this make more sense as a diff/migration guide?


````suggestion
```diff
{
  "scripts": {
-    "lint-path": "sku lint <path>"
+    "lint-path": "tsc && sku lint <path>"
  }
}
```
````

File: tests/node/lint-format.test.ts [hunk 6]
```diff
@@ -1,114 +1,308 @@
...
+ */
+const createSrcFixture = (files: Record<string, string>) =>
+  createFixture(files, { tempDir: fixturePath() });
 
-  afterEach(async () => {
-    await fs.rm(srcDirectory, { recursive: true, force: true });
-  });
+/** The fixture's location relative to the directory `sku` is run from. */
+const relativePath = (fixture: Fixture, ...subpaths: string[]) =>
+  path.relative(fixturePath(), fixture.getPath(...subpaths));
```
askoufis: This comment feels incorrect. I think renaming the function can replace the comment entirely:

```suggestion
const relativePathFromFixture = (fixture: Fixture, ...subpaths: string[]) =>
  path.relative(fixturePath(), fixture.getPath(...subpaths));
```

.changeset/hip-jobs-bathe.md
```diff
@@ -0,0 +1,17 @@
+---
+'sku': major
+---
+
+`lint`: Skip the TypeScript check when file paths are provided
+
+Previously, `sku lint <path>` ran ESLint and Prettier on the given paths but still ran TypeScript across the *entire project*. This is because TypeScript ignores your `tsconfig.json` when filenames are passed explicitly, so the check can't be scoped to a single path.
+
+Since linting a path should only lint that path, `sku` now skips the TypeScript check entirely when a file path is provided. To type-check as well, run `tsc` manually first:
```
askoufis: We want to remove the typecheck because it's unexpected, which is _sort of_ what this says, but "linting a path should only lint that path" feels very circular IMO.

```suggestion
This unexpected behaviour has now been removed: `sku` now skips the TypeScript check entirely when a file path is provided. To replicate the previous behaviour, run `tsc` manually first:
```

tests/node/lint-format.test.ts
```diff
@@ -1,114 +1,308 @@
...
+  let mutable = true;
 
+  expect(mutable).toBe(true);
+});
+`;
+
+/**
+ * A file that passes every linter: valid types, correctly formatted and free of
+ * lint errors.
+ */
```
askoufis: The variable names feel like enough here. I don't think the comments are necessary.

williamlark: Yea I thought the same but I want it to be clear that there are no other errors present in the file apart from the linter being tested, which I think could get lost. I could add a comment above them all stating that instead of repeating it for each file if you prefer, but I'd like to have _something_

.changeset/hip-jobs-bathe.md
williamlark: Yea, I like it

tests/node/lint-format.test.ts
askoufis: I feel like the variable names communicate that clearly, but a single comment above all of them is fine.

williamlark: After trying to summaries the comments it felt a little redundant, so I've just removed them. Agree now that their not needed.

askoufis (approved):

.changeset/great-bats-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': major
+---
+
+`format`: Format will throw on Prettier errors instead of silently passing
```
askoufis: Oops I forgot to submit this comment. This makes it sound like it will throw on prettier formatting errors too, which is not the case. I guess it's specifically "errors that stop prettier from formatting your code, such as invalid permissions or missing config".

PR #1608, opened 2026-06-29
Replace death

williamlark:
Replace `death` package with internal implementation (reducing ~deaths~ deps)

The SIGQUIT test was hanging.  Seems the webpack-dev-server doesn't respond to SIGQUIT events. Probably not an issue, but just to be sure I've handled it for the webpack-dev-server too


jahredhope (approved):
CI issues. But once they are cleared up I'm supportive of removing death.

askoufis:
With https://github.com/seek-oss/sku/pull/1623 merged it's worth rebasing/merging to make sure it still works with the changes in this PR.

askoufis (approved):

PR #1611, opened 2026-06-29
Typescript 6

williamlark:
Bumping the Typescript version to v6, and describing what changes a user should be aware of in the changesets. They will likely not have to change anything.

This should get projects ready for Typescript 7.


tests/browser/path-aliases.test.ts
```diff
@@ -41,8 +41,8 @@ describe('pathAliases', () => {
       });
     });
 
-    it('should preserve existing baseUrl behavior alongside paths', async () => {
-      expect(tsconfig.compilerOptions.baseUrl).toBeDefined();
+    it('should resolve imports via paths without baseUrl', async () => {
```
SpiralLogic: nit: it's not really testing resolution, just that the config has been generated correctly

SpiralLogic (approved):

pnpm-workspace.yaml
```diff
@@ -44,6 +44,7 @@ catalog:
   jiti: ^2.4.2
   react: ^19.1.0
   react-dom: ^19.1.0
+  typescript: ^6.0.3
```
askoufis: We should keep this patch-flexible.
```suggestion
  typescript: ~6.0.3
```

tests/browser/path-aliases.test.ts
williamlark: True, I'll  reword it

fixtures/custom-src-paths/src/client.tsx
```diff
@@ -1,7 +1,6 @@
 import { hydrateRoot } from 'react-dom/client';
 
-/** @knipignore Can't seem to ignore this via knip config */
-import App from 'another-folder/App';
```
askoufis: I would've thought this PR would make the `knipignore` unnecessary - making the import relative now means this fixture isn't testing src paths at all.

packages/sku/src/services/typescript/tsconfig.ts
```diff
@@ -34,18 +64,20 @@ export default ({ rootResolution, tsconfigDecorator, tsPaths }: SkuContext) => {
...
 
+      // Side-effect imports for CSS/assets are resolved by bundlers, not TypeScript
+      noUncheckedSideEffectImports: false,
+
       // misc
       strict: true,
       forceConsistentCasingInFileNames: true,
       jsx: 'react-jsx',
-      lib: ['dom', 'dom.iterable', 'es2022'],
+      lib: ['dom', 'es2022'],
```
askoufis, line 90: I forgot that `dom.iterable` is now included by default - good catch. For reference: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#the-dom-lib-now-contains-domiterable-and-domasynciterable.

packages/sku/src/services/typescript/tsconfig.ts
```diff
@@ -34,18 +64,20 @@ export default ({ rootResolution, tsconfigDecorator, tsPaths }: SkuContext) => {
...
+      // ? should we update this to es2025
```
askoufis: I'd say no. es2022 aligns best with our browser support policy at the moment, so I'm happy to keep it. IMO stopping consumers from using APIs that are too new (and that aren't likely to be polyfilled) by restricting available types is a good thing.

packages/sku/src/services/typescript/tsconfig.ts
```diff
@@ -1,8 +1,38 @@
...
+    const packageJson: PackageJson = requireFromCwd('./package.json');
+    const allDeps = {
+      ...packageJson.dependencies,
+      ...packageJson.devDependencies,
+    };
+    return Boolean(allDeps[`@types/${types}`]);
+  } catch {
+    return false;
+  }
+};
```
askoufis:
```suggestion
const hasTypesInstalledForPackage = (packageName: string) => {
  try {
    // Trying to import @types/{packageName}/package.json can lead to false positives, so instead we'll check the dependencies directly.
    const packageJson: PackageJson = requireFromCwd('./package.json');
    const allDeps = {
      ...packageJson.dependencies,
      ...packageJson.devDependencies,
    };
    return Boolean(allDeps[`@types/${packageName}`]);
  } catch {
    return false;
  }
};
```

packages/sku/src/openBrowser.ts
```diff
@@ -1,7 +1,7 @@
 // Inspired by create-react-app
 // https://github.com/facebook/create-react-app/commit/d2de54b25cc25800df1764058997e3e274bd79ac
 
-import open from 'open';
+import open from 'tiny-open';
```
askoufis, line 4: Seems like this change from master got assimilated into the PR.

williamlark: Should be good now. I've merged in master to the major-v16 branch so it's all up to date

askoufis (approved):
Can remove some `@ts-expect-error` annotations now ([CI](https://github.com/seek-oss/sku/actions/runs/28559656204/job/84674530804#step:9:23) is failing), but otherwise LGTM.

PR #1613, opened 2026-06-30
`v16`: Remove automatic `.eslintignore` migration during `sku configure`

askoufis:
With ESLint v9 well-established, there's no longer a good reason to keep this automatic migration code around. It has been around since v14, so consumers migrating from v14->v15 can still use it if necessary.


williamlark (approved):
So nice to see this gone 🌟

SpiralLogic (approved):

PR #1618, opened 2026-06-30
`v16`: Node 24, replace `pretty-ms` and `escape-string-regexp` with native APIs

askoufis:
The `@ts-expect-error` annotations should error once the TS 6.0 PR is merged.


williamlark (dismissed):

askoufis:
My initial implementation assumed that the formatter would appropriately split the duration into hours, minutes, seconds, etc., but it doesn't, so I've implemented that and added some tests. It roughly matches how `pretty-ms` worked (doesn't show milliseconds if the duration is more than 1 second), but doesn't care about day-length durations because that's unlikely given we're timing app builds.

williamlark (approved):

PR #1622, opened 2026-06-30
`v16`: Convert loadable entrypoint to JavaScript

askoufis:
I think this was just a leftover thing we did from the days before we started bundling `sku`. I can't think of a good reason to keep doing it, so the entrypoint is now MJS.

I also removed the `vite/client` entrypoint from `tsdown`, as that's configured manually in `package.json` because it points to a `.d.ts` file.

### DIFF
--- a/packages/sku/tsdown.config.ts
+++ b/packages/sku/tsdown.config.ts
@@ -10,16 +10,13 @@ export default defineConfig([
         from: 'src/services/vite/client.d.ts',
         to: 'dist/vite/',
       },
-      {
-        from: 'src/@loadable/component/index.ts',
-        to: 'dist/@loadable/component/',
-      },
     ],
     // Need to use unbundled mode because `webpack/entry/server/index.ts` calls webpackHot.accept which needs a known path to the server app at runtime.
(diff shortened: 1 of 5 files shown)


SpiralLogic (approved):

packages/sku/tsdown.config.ts
```diff
@@ -29,7 +26,6 @@ export default defineConfig([
       'jest/js-transform': 'src/config/jest/jsBabelTransform.ts',
       'jest/ts-transform': 'src/config/jest/tsBabelTransform.ts',
       postinstall: './src/postinstall.ts',
-      'vite/client': 'src/services/vite/index.ts',
```
williamlark, line 32: Yea this was definitely wrong. Good catch

tests/browser/__snapshots__/display-names-prod.test.ts.snap
williamlark, line 1: Looks like this fixture is missing makeStableHashes in the webpack sku config.

williamlark (approved):

PR #1603, opened 2026-06-29
Adding legacy react warning

williamlark:
Sku  fully supports React 19 and we'd like applications to move to React 19. This adds a banner when React 18 is detected to encourage users to upgrade.

PR #1627, opened 2026-06-30
Replace accidental `debug` usage with `obug`

askoufis:
I think this snuck in via #1603 because it wasn't up to date with the upstream branch.


williamlark (approved):
thank you  🐛

PR #1628, opened 2026-06-30
`v16`: Replace `cssnano` with `lightningcss` for webpack CSS minification

askoufis:
`lightningcss` has been the default minifier in Vite since v8, so consumers have been installing it as a dep regardless of whether they use Vite. I felt that now was a good opportunity to align minified CSS output between webpack and Vite, and moving webpack users to it means we can remove a good handful of deps that were brought in by `cssnano` and `autoprefixer`.

Testing minfied output against some internal apps, I have not noticed any changes that could cause styling issues. The `calc` optimizations that we turn off in `cssnano` cannot be disabled in `lightningcss`, but I don't think the issues this can cause are common or relevant these days anyway. I'll leave more comments inline on some potentially unexpected but valid changes.


tests/browser/__snapshots__/braid-design-system.test.ts.snap
```diff
@@ -13458,7 +13458,10 @@ exports[`braid-design-system > bundler webpack > build > should generate the exp
 ._19azws3d[type="number"] {
   -moz-appearance: textfield;
 }
-._19azws3d[type="number"]::-webkit-inner-spin-button,
+._19azws3d[type="number"]::-webkit-inner-spin-button {
+  -webkit-appearance: none;
+  margin: 0;
+}
 ._19azws3d[type="number"]::-webkit-outer-spin-button {
```
askoufis, line 13465: I _think_ `lightningcss` is splitting these up because it doesn't recognize this vendor-specific pseudo-selector. In CSS, if 1 pseudo-selector is invalid then the entire selector is invalid, so to prevent all styles potentially not applying, it splits them up to reduce the blast-radius. Not a big deal, just noting it.

tests/browser/__snapshots__/braid-design-system.test.ts.snap
```diff
@@ -15699,7 +15702,7 @@ html.r2i18210 .fx8zm71l {
   top: -0.105em;
 }
 ._16185315 {
-  top: -0.16499999999999998em;
+  top: -0.165em;
```
askoufis, line 15705: `lightningcss` tends to round quite aggressively, and apparently incorrectly in some cases - see https://github.com/parcel-bundler/lightningcss/issues/949 for more info. I haven't seen any issues in our apps caused by this, but it might be worth making sure it doesn't break capsize styles or something (they may be quite precision-dependent).

SpiralLogic (approved):

tests/browser/__snapshots__/braid-design-system.test.ts.snap
```diff
@@ -15865,46 +15868,46 @@ html.r2i18210 ._778n5bf {
...
   height: var(--_1hw07lg0);
-  outline: var(--_9h6w0d10) solid transparent;
-  transition: outline-color 0.125s ease;
   width: var(--_1hw07lg0);
+  outline: var(--_9h6w0d10) solid transparent;
+  transition: outline-color 0.125s;
 }
 ._1hw07lg4[type="checkbox"]:checked ~ ._1hw07lg5 {
-  background: transparent;
+  background: 0 0;
```
askoufis, line 15887: `background-color` defaults to `transparent`, and this is equivalent to `background: none`, which is then minified further to `background: 0 0` (`background-position` defaults to `0 0`).

williamlark (approved):

tests/browser/__snapshots__/braid-design-system.test.ts.snap
askoufis: Update: it seems to only recognize the single-colon version. See [here](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22._19azws3d%5Btype%3D%5C%22number%5C%22%5D%3A-webkit-inner-spin-button%2C%20._19azws3d%5Btype%3D%5C%22number%5C%22%5D%3A-webkit-outer-spin-button%20%7B%5Cn%20%20-webkit-appearance%3A%20none%3B%5Cn%20%20margin%3A%200%3B%5Cn%7D%5Cn%5Cn._19azws3d%5Btype%3D%5C%22number%5C%22%5D%3A%3A-webkit-inner-spin-button%2C%20._19azws3d%5Btype%3D%5C%22number%5C%22%5D%3A%3A-webkit-outer-spin-button%20%7B%5Cn%20%20-webkit-appearance%3A%20none%3B%5Cn%20%20margin%3A%200%3B%5Cn%7D%5Cn%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D).

PR #1630, opened 2026-07-07
`csp`: Add `cspDelivery` config option

kjots:
This change adds a new configuration option, `cspDelivery`, that allows control of how the Content Security Policy is delivered. It can be set to one of two values:

- `tag`: The CSP will be embedded directly in the rendered HTML content via a `<meta http-equiv="Content-Security-Policy" …>` tag. This is the default and matches the previous behaviour.
- `header`: The CSP will be written to a JSON metadata file alongside the rendered HTML content. This metadata can be used at deployment and/or request time to include a `Content-Security-Policy` header in the response for the rendered HTML content.

Note: This change will be released alongside #1631 which adds support for a report-only Content Security Policy.

### DIFF
--- a/packages/sku/src/types/types.ts
+++ b/packages/sku/src/types/types.ts
@@ -540,7 +540,7 @@ export interface ViteSkuConfig {
   vitePlugins?: PluginOption[];
 
   /**
-   * The way the content security policy is delivered.  Only relevant if {@link SkuConfigBase#cspEnabled} is set to `true`.
+   * The way the content security policy is delivered. Only relevant if {@link SkuConfigBase#cspEnabled} is set to `true`.
    *
    * @default 'tag'
    * @link https://seek-oss.github.io/sku/docs/configuration#cspdelivery
--- a/packages/sku/src/types/types.ts
+++ b/packages/sku/src/types/types.ts
@@ -540,6 +540,14 @@ export interface ViteSkuConfig {
   vitePlugins?: PluginOption[];
 
   /**
+   * The way the content security policy is delivered.  Only relevant if {@link SkuConfigBase#cspEnabled} is set to `true`.
+   *
+   * @default 'tag'
+   * @link https://seek-oss.github.io/sku/docs/configuration#cspdelivery
+   */
+  cspDelivery?: 'tag' | 'header';
+
+  /**
    * This function provides a way to modify sku's Vite configuration.
    * It should only be used in exceptional circumstances where a solution cannot be achieved by adjusting standard configuration options.
    *
--- a/tests/browser/__snapshots__/security-controls.test.ts.snap
+++ b/tests/browser/__snapshots__/security-controls.test.ts.snap
@@ -91,6 +91,100 @@ exports[`security-controls > build-ssr > should start a server with content-secu
 </html>
 `;
 
+exports[`security-controls > bundler vite > csp-delivery > start > should start an app with security controls 1`] = `
+content-security-policy: script-src 'self' https://some-cdn.com 'sha256-Z2/iFzh9VMlVkEOar1f/oSHWwQk3ve1qk/C2WdsC4Xk=' 'sha256-UBWhRaCtYn0qJIs3GeFksNFLt/GkChOdRUDOMGEwb4A=' 'nonce-RANDOM_NONCE';
+===================================================================
+--- sourceHtml
++++ clientHtml
+@@ -1,68 +1,82 @@
+ <!DOCTYPE html>
+ <html>
+   <head>
+     <script type="module">
+       import { injectIntoGlobalHook } from "/@react-refresh";
+ injectIntoGlobalHook(window);
+ window.$RefreshReg$ = () => {};
+ window.$RefreshSig$ = () => (type) => type;
+     </script>
+     <script
+       type="module"
+       src="/@vite/client"
+     >
+     </script>
+     <meta charset="UTF-8">
+     <title>
+       hello-world
+     </title>
+     <meta
+       name="viewport"
+       content="width=device-width, initial-scale=1"
+     >
+     <script
+       type="module"
+-      src="/@id/__x00__/index.html?html-proxy&index=0.js"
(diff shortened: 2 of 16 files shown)


docs/docs/csp.md
```diff
@@ -8,6 +8,19 @@
 
 Set `cspEnabled: true` in your `sku.config.js`.
 
+### Delivery
+
+The `cspDelivery` option controls how the CSP is delivered and can be set to one of two values:
+
+- `tag`: The CSP will be embedded directly in the rendered HMTL content via a `<meta http-equiv="Content-Security-Policy" …>` tag.
```
SpiralLogic:
```suggestion
- `tag`: The CSP will be embedded directly in the rendered HTML content via a `<meta http-equiv="Content-Security-Policy" …>` tag.
```

SpiralLogic:
It doesn't look like there are any tests for the new stuff?

packages/sku/src/types/types.ts
```diff
@@ -156,6 +156,14 @@ export interface SkuConfigBase {
    */
   cspEnabled?: boolean;
 
+  /**
+   * The way the content security policy is delivered.  Only relevant if {@link cspEnabled} is set to `true`.
+   *
+   * @default 'tag'
+   * @link https://seek-oss.github.io/sku/#/./docs/configuration?id=cspdelivery
```
SpiralLogic: There isn't a cspdelivery id in the docs yet, should these be added together?

docs/docs/csp.md
kjots: Fixed.

kjots:
> It doesn't look like there are any tests for the new stuff?

No, there is not. The current test fixture support is based around the HTML response content and as-is cannot incorporate either the HTTP headers returned from the dev server or the metadata JSON generated by the build. With no standardised way to assert on the headers or the metadata, there is nothing to test with this change.

There are also some parts of the test fixtures that rely on the CSP being embedded in the HTML. For example, the serializer will [replace any randomly-generated nonce value](https://github.com/seek-oss/sku/blob/69f57910886062a9405a8c8a60d7ae754cf203ac/private/playwright/formatHtml.ts#L5) with a stable string, so that the snapshots do not change from test to test. However this requires the nonce to be stored in the `meta` tag, and with the new delivery option this will not be the case. So this also will need to be changed in some manner to preserve the existing snapshot behaviour with the new option.

Basically in order to be able to test these changes the existing test support will need to be re-architected somewhat - it is an open question as to whether such support should block this change, or if we go ahead as-is and fill in the test support with a future change.

packages/sku/src/types/types.ts
kjots: Yes, I presume the contents of the `docs/docs/configuration.md` file is updated automatically from `packages/sku/src/types/types.ts` but I am not aware of how - I will probably seek some support on this during the next pairing session.

kjots: The `docs/docs/configuration.md` file has been updated.

tests/browser/security-controls.test.ts
```diff
@@ -77,6 +77,65 @@ describe('security-controls', () => {
...
+          const content = await readFile(indexJsonPath, 'utf-8');
+          const data = JSON.parse(content);
+          const metadata = data.metadata?.csp ?? null;
+
+          if (!metadata) {
+            throw new Error('Unable to select CSP metadata');
+          }
+
+          cspHeader = metadata;
+        });
```
williamlark: Can this be done in beforeAll as well? It's only build once so the content wouldn't change between them

williamlark (approved):
Nice work!

tests/browser/security-controls.test.ts
kjots: Good pickup, updated to use `beforeAll` for all `build` tests.

kjots:
> It doesn't look like there are any tests for the new stuff?

Following up on the [previous reply](https://github.com/seek-oss/sku/pull/1630#issuecomment-4922240586), there are now tests for the new stuff.

tests/browser/__snapshots__/security-controls.test.ts.snap
```diff
@@ -91,6 +91,100 @@ exports[`security-controls > build-ssr > should start a server with content-secu
...
+       src="{cwd}/packages/sku/dist/entries/vite-client.mjs"
+     >
+     </script>
+     <script
+       id="__SKU_CLIENT_CONTEXT__"
+       type="application/json"
+     >
+       {"environment":"","dynamicScriptNonce":"RANDOM_NONCE"}
+     </script>
++    <script nonce>
```
kjots, line 180: This appears to be the expected behaviour - when the CSP is delivered via a header, the nonce value is _accepted_ (i.e. the script will run when the `nonce` attribute is set to the expected value, and will _not_ run when it is not), however the attribute value itself is _cleared_ immediately after the element is attached to the DOM and so therefore cannot be captured in the snapshot. See also https://stackoverflow.com/a/55673767.

This [does not occur](https://github.com/seek-oss/sku/blob/4950d6e463f12f6f3c151a831c576ff14902dfac/tests/browser/__snapshots__/security-controls.test.ts.snap#L183) when the CSP is delivered via a `<meta>` tag, possibly because the nonce value is already embedded within the HTML content.

packages/sku/src/types/types.ts
```diff
@@ -539,6 +539,14 @@ export interface ViteSkuConfig {
    */
   vitePlugins?: PluginOption[];
 
+  /**
+   * The way the content security policy is delivered.  Only relevant if {@link SkuConfigBase#cspEnabled} is set to `true`.
```
askoufis:
```suggestion
   * The way the content security policy is delivered. Only relevant if {@link SkuConfigBase#cspEnabled} is set to `true`.
```

askoufis (approved):
~Should a test be added for this change?~ Ignore this, forgot to clear it out from an old pending review before submitting.

PR #1631, opened 2026-07-07
`csp`: Add support for report-only Content Security Policy

kjots:
This change adds two new configuration options, `cspReportOnlyEnabled` and `cspReportOnlyExtraScriptSrcHosts`, that allow the generation of a "report-only" CSP delivered via the `Content-Security-Policy-Report-Only` header.

Note: This change will be released alongside #1630 which adds support for Content Security Policy delivery via HTTP header.

### DIFF
--- a/packages/sku/src/services/webpack/entry/csp.ts
+++ b/packages/sku/src/services/webpack/entry/csp.ts
@@ -12,6 +12,7 @@ const hashScriptContents = (scriptContents: BinaryLike) =>
 
 interface CreateCSPHandlerOptions {
   extraHosts?: string[];
+  reportOnlyExtraHosts?: string[];
   isDevelopment?: boolean;
 }
 
@@ -19,6 +20,7 @@ export type CSPHandler = {
   registerScript: (script: string) => void;
   createCSP: () => string;
   createCSPTag: () => string;
+  createReportOnlyCSP: () => string;
   createUnsafeNonce: () => string;
   processHtml: (html: string) => HTMLElement;
   updateHtml: (root: HTMLElement) => string;
@@ -27,12 +29,14 @@ export type CSPHandler = {
 
 export default function createCSPHandler({
   extraHosts = [],
+  reportOnlyExtraHosts = [],
   isDevelopment = false,
 }: CreateCSPHandlerOptions = {}): CSPHandler {
   let cspCreated = false;
-  const hosts = new Set();
-  const shas = new Set();
-  const nonces = new Set();
+  const hosts = new Set<string>();
+  const reportOnlyHosts = new Set<string>();
+  const shas = new Set<string>();
+  const nonces = new Set<string>();
 
   const addScriptContents = (contents: BinaryLike | undefined) => {
     if (contents) {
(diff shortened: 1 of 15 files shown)


packages/sku/src/services/webpack/entry/csp.ts
```diff
@@ -93,7 +107,7 @@ export default function createCSPHandler({
     parse(script).querySelectorAll('script').forEach(processScriptNode);
   };
 
-  const createCSP = () => {
+  const createCSPForHosts = (set: Set<unknown>) => {
```
askoufis: Maybe it's worth annotating the `Set` types on lines 36-39 so `unknown` doesn't flow through to here.

```suggestion
  const createCSPForHosts = (set: Set<string>) => {
```

tests/browser/__snapshots__/security-controls.test.ts.snap
```diff
@@ -185,6 +185,104 @@ content-security-policy: script-src 'self' https://some-cdn.com 'sha256-Z2/iFzh9
...
+===================================================================
+--- sourceHtml
++++ clientHtml
+@@ -1,72 +1,86 @@
+ <!DOCTYPE html>
+ <html>
+   <head>
+     <meta
+       http-equiv="Content-Security-Policy"
+       content="script-src 'self' https://some-cdn.com 'sha256-Z2/iFzh9VMlVkEOar1f/oSHWwQk3ve1qk/C2WdsC4Xk=' 'sha256-UBWhRaCtYn0qJIs3GeFksNFLt/GkChOdRUDOMGEwb4A=' 'nonce-RANDOM_NONCE';"
```
SpiralLogic, line 199: when do the other CSP directives get injected? like frame, object-src, default-src etc?

kjots: As far as I know there is no support for any other CSP directives at this time, only `script-src` with custom hosts and hashes/nonces are available. 

I don't know if there are any plans to add such support.

packages/sku/src/types/types.ts
```diff
@@ -547,6 +547,24 @@ export interface ViteSkuConfig {
...
 
+  /**
+   * **Unavailable for libraries**
+   *
+   * Enable report-only content security policy feature. More info at https://seek-oss.github.io/sku/docs/csp
+   *
+   * @default false
+   * @link https://seek-oss.github.io/sku/docs/configuration#cspreportonlyenabled
+   */
+  cspReportOnlyEnabled?: boolean;
```
SpiralLogic, line 558: where is the report-uri defined for the report-only report to go to? do consumers need to add this themselves and should there be some documentation for that?

askoufis (approved):
Just a small comment, otherwise LGTM.

packages/sku/src/types/types.ts
kjots: At present there is no support for the `report-to` feature for CSPs, report-only CSP violations can only be monitored via the console or through the [`securitypolicyviolation`](https://developer.mozilla.org/en-US/docs/Web/API/Element/securitypolicyviolation_event) event.

I assume we will add support for the [`report-to`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/report-to) directive and the [`Reporting-Endpoints`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints) header for both standard and report-only CSPs at a future date.

packages/sku/src/services/webpack/entry/csp.ts
kjots: Done.

PR #1634, opened 2026-07-09
Adding mocking support to root-resolution codemod

williamlark:
The codemod didn't catch test mock imports. This adds support for both vitest and jest.

No changeset added since this codemod isn't live yet and we have a changeset for the release already.


tests/node/codemods/migrate-root-resolution.test.ts
```diff
@@ -18,6 +18,13 @@ runCodemodTests('migrate-root-resolution', [
       const cjs = require('src/cjs');
       import local from './src/local';
       import external from 'pkg/src/thing';
+      vi.mock('src/mocked');
```
askoufis, line 22: Can we also test the case where a mock factory is provided?

packages/codemod/src/codemods/migrate-root-resolution.ts
```diff
@@ -34,7 +34,21 @@ const transformSource = async (source: string): Promise<string | null> => {
...
               has: {
                 field: 'function',
-                any: [{ kind: 'import' }, { regex: '^require$' }],
+                any: [
+                  { kind: 'import' },
+                  { regex: '^require$' },
+                  { pattern: 'vi.mock', kind: 'member_expression' },
+                  { pattern: 'vi.importActual', kind: 'member_expression' },
+                  { pattern: 'jest.mock', kind: 'member_expression' },
+                  { pattern: 'jest.requireActual', kind: 'member_expression' },
```
askoufis, line 45: There's also [`vi.doMock`](https://vitest.dev/api/vi.html#vi-domock) and [`jest.doMock`](https://jestjs.io/docs/jest-object#jestdomockmodulename-factory-options) APIs.

askoufis (approved):

PR #1635, opened 2026-07-09
`vite`: Add support for serving assets from a `public` directory

askoufis:
Braid (and other apps) utilize a public asset directory for serving non-hashed assets such as favicons and opengraph assets. This is supported for webpack apps but support was never implemented for vite apps. Until now!

The comment in the configuration docs has been removed as I don't think disabling this feature entirely makes sense. It's useful and already used internally.


fixtures/public-path/sku.config.js
```diff
@@ -4,6 +4,6 @@ export default {
   clientEntry: 'src/client.jsx',
   renderEntry: 'src/render.jsx',
   publicPath: '/static/my-app',
-  port: 4001,
+  port: 4004,
```
askoufis, line 7: Unrelated change: this port clashes with the `custom-output-target` fixture's port, so I've made it unique.

packages/sku/src/services/vite/plugins/devServer.ts
```diff
@@ -14,7 +14,9 @@ export const devServerPlugin = ({
   skuContext: SkuContext;
 }): PluginOption => ({
   name: makePluginName('dev-server'),
+  apply: 'serve',
```
askoufis, line 17: These _feels_ like a benign change: dev server (i.e. `start`) config shouldn't bleed into `build` config, so I've scoped this to `serve`. Happy to revert it if there are concerns.

.changeset/wise-towns-ask.md
```diff
@@ -0,0 +1,9 @@
+---
+'sku': minor
+---
+
+`vite`: Add support for `public` config
```
askoufis:
```suggestion
`vite`: Enable support for configuring a public assets directory via the `public` option in sku config
```

.changeset/wise-towns-ask.md
```diff
@@ -0,0 +1,9 @@
+---
+'sku': minor
+---
+
+`vite`: Enable support for configuring a public assets directory via the `public` option in sku config
+
+Previously, the [`public`] configuration option was only supported in Webpack apps. It now also support in Vite apps.
```
askoufis:
```suggestion
Previously, the [`public`] configuration option was only supported in Webpack apps. It is now also support in Vite apps.
```

.changeset/wise-towns-ask.md
```diff
@@ -0,0 +1,9 @@
+---
+'sku': minor
+---
+
+`vite`: Enable support for configuring a public assets directory via the `public` option in sku config
+
+Previously, the [`public`] configuration option was only supported in Webpack apps. It is now also support in Vite apps.
```
askoufis:
```suggestion
Previously, the [`public`] configuration option was only supported in Webpack apps. It is now also supported in Vite apps.
```

packages/sku/src/utils/buildFileUtils.ts
```diff
@@ -29,9 +29,24 @@ export const copyPublicFiles = async ({
...
+    .crawl(paths.public)
+    .withPromise();
+
+  for (const file of files) {
+    const relativePublicPath = relative(paths.public, file);
+    const sourcePath = relative(process.cwd(), file);
+    const destinationPath = join(paths.relativeTarget, relativePublicPath);
+
+    console.log(`Copying ${sourcePath} to ${destinationPath}`);
+  }
```
williamlark: Could just log that were moving the entire folder. Probably not super useful to see every file thats copied.

williamlark (approved):

packages/sku/src/utils/buildFileUtils.ts
askoufis: https://github.com/seek-oss/sku/pull/1635/changes/6b401dd68a9a56114c121787cb710b89ff5ddac3

Wasn't sure about the exact log message wording, so I settled on:

```
Copying public assets from "assets" to "dist"
```

askoufis:
Realised that by only configuring `publicDir` during local dev, imports to `publicDir` assets won't resolve during build. Setting it globally fixes this, and it's also now tested in the fixture. See https://github.com/seek-oss/sku/pull/1635/changes/c9acbc8d4796aba3d9faa08659a898f273e7e09a.

I don't think webpack supports importing (unhashed) assets like this, so I didn't test it in webpack.

PR #1638, opened 2026-07-10
Bump browserlists to latest

williamlark:
Updating browserlists to the latest version. We updated the browserlists target to `24.16.0` in another branch  but the current minimum pin we have doesn't support that version.

Users can update this dep on their own with a lockfile refresh, but I don't  think it makes sense for sku to have a version in its range that doesn't supoprt it.

No changeset since it'll be bundled with the node version bump release.

### DIFF
--- a/packages/sku/src/services/vite/helpers/browserslist-to-esbuild.test.ts
+++ b/packages/sku/src/services/vite/helpers/browserslist-to-esbuild.test.ts
@@ -12,8 +12,8 @@ test('works by passing browsers as array', () => {
   expect(target).toMatchInlineSnapshot(`
     [
       "chrome103",
-      "edge147",
-      "firefox150",
+      "edge148",
+      "firefox121",
       "ios11",
       "opera131",
       "safari18.5",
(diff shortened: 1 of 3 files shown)


packages/sku/src/services/vite/helpers/browserslist-to-esbuild.test.ts
```diff
@@ -12,8 +12,8 @@ test('works by passing browsers as array', () => {
   expect(target).toMatchInlineSnapshot(`
     [
       "chrome103",
-      "edge147",
-      "firefox150",
+      "edge148",
+      "firefox121",
```
jeremy-putt, line 16: 121?

jahredhope (approved):

packages/sku/src/services/vite/helpers/browserslist-to-esbuild.test.ts
williamlark: Guess so! Firefox 121 apparently has had a surge of users. 
We test `browserslist-config-seek` as the last test snapshot here, and that snapshot hasn't changed so this is fine.

PR #1640, opened 2026-07-13
Install fake react 18 dependency

askoufis:
Our upkeep score has tanked because we are installing react 18 again. Instead of installing the dep, we now make a fake dep inside the `legacy-react` fixture during the test.

Also removed a stale `react-18` snapshot.


tests/node/legacy-react.test.ts
```diff
@@ -1,11 +1,26 @@
...
+    JSON.stringify(fakeReactPackageJson),
+    'utf8',
+  );
+
   const skuProcess = await sku('configure');
   expect(
     await skuProcess.findByText('React 18 or below detected'),
   ).toBeInTheConsole();
+
+  await rmdir(fixturePath('node_modules/react'), { recursive: true });
```
askoufis: Cleanup ordering isn't too critical for this fixture since it's all within `node_modules`.

fixtures/legacy-react/package.json
```diff
@@ -4,10 +4,11 @@
...
   "type": "module",
   "scripts": {
-    "start": "sku start"
+    "start": "sku start",
+    "COMMENT": "echo "
   },
-  "dependencies": {
-    "react": "18.3.1"
+  "comments": {
+    "react": "A fake react dependency is inected during the test so we don't need to install the real react"
```
williamlark:
```suggestion
    "react": "A fake react dependency is injected during the test so we don't need to install the real react"
```

tests/node/legacy-react.test.ts
williamlark: rmdir options are deprecated so I think rm is preferred now 
```suggestion
  await rm(fixturePath('node_modules/react'), { recursive: true, force: true });
```

williamlark (approved):

PR #1643, opened 2026-07-14
Pin webpack

williamlark:
Webpack seems to have issues at the moment, but is fixed in [this PR](https://github.com/webpack/webpack/pull/21407). Until that's out I've pinned webpack to fix our CI issues. Can unpin after that linked PR is merged.

I'm hoping they release it before the sku v16 release, so I've not made a changeset yet. Can add one though if we need.


askoufis (approved):

askoufis:
Does a dedupe help to reduce the additions to the lockfile? Seems like a lot.

PR #1651, opened 2026-07-15
V16 changeset

williamlark:
* Ordering the changesets so the most important ones are up the top using a prefixed filename (`0-`). 
* Added a scope to all the changesets and updated the file name to match (e.g., `deps-`). This will group the same type of changesets together (could update the changeset.js to group by this in future if we wanted)
* Added or moved the **BREAKING CHANGE** text. I tried to place it above where the actual breaking change is for the consumer. For a lot of these it's just the description, for others it's a bit more specific to what will (possibly) break someones app
* Added a preamble file


.changeset/.PREAMBLE.md
```diff
@@ -0,0 +1,7 @@
+There are a lot of changes in this major release, however most consumers should only need to run the following to address the majority of issues:
+
+```sh
+pnpm dlx @sku-lib/codemod@ migrate-root-resolution . && sku format
```
askoufis: Just realized it should be `pnpm sku format`.

```suggestion
pnpm dlx @sku-lib/codemod@ migrate-root-resolution . && pnpm sku format
```

.changeset/deps-solid-cooks-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'sku': patch
+---
+
+`deps`: Update `@vanilla-extract/vite-plugin` dependency to `^5.2.4`
```
askoufis: We should bump this to [5.2.5](https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fvite-plugin%405.2.5) to get the bug fix. Might as well do it in this PR.

.changeset/start-itchy-cameras-matter.md
```diff
@@ -2,10 +2,9 @@
 'sku': major
 ---
 
-SSR server now accessible via webpack dev server
-
-When running `sku start` users would previously access the SSR server directly. Static assets would then be loaded from the webpack dev server on a separate host.
+`start`: SSR server now accessible via webpack dev server
 
+When running `sku start`, users would previously access the SSR server directly. Static assets would then be loaded from the webpack dev server on a separate host.
 `sku start` will now proxy requests from the webpack dev server to the consumer's app (the SSR server), allowing the dev server to act as a single entrypoint.
```
askoufis: My bad, all these `start`s should be `start-ssr`s.

williamlark: You mean start-sdd?

.changeset/.PREAMBLE.md
askoufis:
```suggestion
pnpm dlx @sku-lib/codemod migrate-root-resolution . && sku format
```

askoufis (approved):

pnpm-workspace.yaml
```diff
@@ -32,7 +32,7 @@ catalog:
   '@vanilla-extract/css': '^1.20.1'
   '@vanilla-extract/integration': '^8.0.9'
   '@vanilla-extract/jest-transform': '^1.1.21'
-  '@vanilla-extract/vite-plugin': ^5.2.4
+  '@vanilla-extract/vite-plugin': ^5.2.5
```
jahredhope, line 35: This seems weirdly unrelated. An accident or intentional?

williamlark: Yea intentional. Adam mentioned it in a comment here https://github.com/seek-oss/sku/pull/1651#discussion_r3584757170

PR #1652, opened 2026-07-15
`csp`: Use fixed nonce value for tests

kjots:
This change updates the `createUnsafeNonce` function of the CSP handler to check for a `SKU_CSP_NONCE` environment variable and, should it exist, use that as the value of the returned nonce. In the absence of such an environment variable, a randomly generated nonce value is created and used instead as per the existing behaviour.

This change also updates the Vitest config to set the above environment variable to a fixed value, and backs out the manual replacement of randomly generated nonces with the same fixed value in the formatted HTML snapshots. Since the environment is passed through transparently both to Node worker threads and to the render processes scoped to the test fixtures, these changes together eliminate the randomness of CSP nonce values in the testing environment altogether without the need to update any tests or snapshots[^1].

Partially inspired by #1646.

[^1]: One test has been updated to assert explicitly on the fixed nonce value, however this test would still have passed without this change since (part of) the fixed value conformed to the match pattern.

### DIFF
--- a/packages/sku/src/services/webpack/entry/csp.test.ts
+++ b/packages/sku/src/services/webpack/entry/csp.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest';
+import { afterEach, beforeEach, describe, it, expect, vi } from 'vitest';
 import createCSPHandler from './csp.js';
 
 describe('createCSPHandler', () => {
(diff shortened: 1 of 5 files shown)


packages/sku/src/services/webpack/entry/csp.test.ts
```diff
@@ -93,6 +93,10 @@ describe('createCSPHandler', () => {
   });
 
   describe('createUnsafeNonce', () => {
+    beforeEach(() => {
+      vi.stubEnv('SKU_CSP_NONCE', undefined);
```
kjots, line 97: For those wondering why the `SKU_CSP_NONCE` environment variable is set everywhere _except_ here, I feel like it is better for the nonce value to be stable by default across all tests, and to only re-introduce the randomness for tests that actually require it (e.g. when validating uniqueness, etc).

jahredhope (approved):

PR #1659, opened 2026-07-16
Make create tests install local sku via packed tarball

jahredhope:
## Purpose

`@sku-lib/create` integration tests install unversioned `sku` from the registry, so they couple to the latest published package and WIP create↔configure fixes fail or lie until release. The existing `linkWorkspacePackages` fixture workaround is fragile and breaks once create writes a nested `pnpm-workspace.yaml`.

## Changes

- Add an internal test-only env hook `SKU_CREATE_SKU_SPECIFIER` so create installs sku from a caller-supplied dependency specifier (typically `sku@file:<abs-path-to-.tgz>`) instead of the registry
- Rework `sku-create` integration tests to `pnpm pack` local `packages/sku` into a unique `os.tmpdir()` dir, pass the specifier override, and drop the fragile `linkWorkspacePackages` fixture harness
- After create succeeds, assert the new project’s `lint` passes (quality gate lives in the harness, not a create-time hard fail)
- Thread custom env through `@sku-private/testing-library` create helper (merge with `process.env` so PATH is preserved)
- Fix soft-mode format messaging: after sku’s lint/format exit-code consolidation (#1607), `sku format` exits `1` for failures. Soft mode previously treated exit `1` as “warnings”; it now warns with failure-style messaging for any non-zero exit and still continues
- Add OpenSpec change artifacts for `create-sku-tarball-override` (proposal / design / delta spec / tasks)


jahredhope:
## Follow-up from offline feedback: soft-mode format messaging

After sku’s lint/format exit-code consolidation (https://github.com/seek-oss/sku/pull/1607), sku format exits 1 for failures. Soft-mode create previously treated exit 1 as “warnings” and other non-zeros as failures, so real format errors were mis-labelled. Soft mode now warns with failure-style messaging for any non-zero exit and still continues when SKU_CREATE_STRICT is unset.

private/testing-library/render-create.ts
```diff
@@ -17,6 +17,13 @@ export const scopeToFixture = (dir: string) => {
       render(createSkuBin, [projectName, ...args], {
         ...options,
         cwd: fixturePath(options.cwd ?? ''),
+        spawnOpts: {
```
kjots, line 21: I have a concern with this that, along with the earlier change to [`sku()`](https://github.com/seek-oss/sku/pull/1646/changes#diff-603813c63c5046146cdca1f3ba1f3193cccbb6d75fc4f6558e0f1d3ba3973ddcR45), we are effectively forking the default behaviour of `render()` but only for _some_ usages:

- For `sku()` and, with this change, `create()`, the existing environment is preserved if `spawnOpts.env` is provided.
- For all other usages, the existing environment is clobbered if `spawnOpts.env` is provided.

I'm thinking we should add a helper that preserves the environment by default and update _all_ usages to use that instead of calling `render()` directly, so we don't have to keep track of where the environment is preserved and where it is not.

kjots: Agreed on Slack for me to implement this change separately.

kjots (approved):

private/testing-library/render-create.ts
kjots: Raised #1663 to address this concern, once that PR is merged it will not be necessary to set `spawnOpts` here.

PR #1661, opened 2026-07-20
docs: Convert diff code blocks into colored diffs

askoufis:
With doscify we could either have diff highlighting or syntax highlighting, but with Vitepress [we can have both](https://vuejs.github.io/vitepress/v1/guide/markdown#colored-diffs-in-code-blocks)!

Example:
image


jahredhope (approved):
Image

PR #1663, opened 2026-07-21
`testing-library`: Always preserve environment when using `render`

kjots:
This change adds a new helper function, `renderWithEnvironment`, to the `@sku-private/testing-library` package. This helper will always preserve the current environment when spawning a command using `render` from `cli-testing-library`, even if `spawnOpts.env` is provided which ordinarily would replace the existing environment. This change also updates all usages of `render` in `@sku-private/testing-library` to use `renderWithEnvironment` instead.

See [this comment](https://github.com/seek-oss/sku/pull/1659/changes#r3618574324) for some background on this change.

### DIFF
--- a/private/testing-library/render-cli.ts
+++ b/private/testing-library/render-cli.ts
@@ -1,6 +1,7 @@
-import { render, waitFor, type RenderOptions } from 'cli-testing-library';
+import { waitFor, type RenderOptions } from 'cli-testing-library';
 import { createRequire } from 'node:module';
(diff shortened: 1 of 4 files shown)


kjots:
There remains a [single direct use](https://github.com/seek-oss/sku/blob/0d1215dbfd65127633c1dfd139e3be22f80a9083/tests/node/codemods/interactive-cli.test.ts#L12) of `render` following this change, however this function does not accept options and therefore cannot provide `spawnOpts.env`, so the existing environment will be preserved by default.

askoufis (approved):

PR #1665, opened 2026-07-21
Fix eslint tsconfig errors

williamlark:
This fixes our eslint being very slow on cold loads. It was taking ages for me on a cold run, and turns out it was due to it reading heaps of configs and rerunning the search a bunch of times. This then lead to making our tsconfigs use the composite setting to speed up tsc linting.

For eslint:
> The result: A from-scratch `pnpm lint:eslint` run dropped from 137.4s to 13.4s (~10x faster)

For the typescript change I compared the difference between using `--incramental --noEmit` versus `--build`. The pros for noEmit is that it doesn't emit any files at all and is a little simpler in the config settings, however the graph below shows the speed differences between the two (ignore the "current" label, that's incorrect).

Speed results - _gonna update this description with what this means TBA_
CleanShot 2026-07-24 at 16 09 07

noEmit is slightly faster on cold cache, but the speed difference is pretty significant between the two on a warm cache, with `--build` being far quicker. Our eslint run is still the bottleneck (~16s) but I don't feel like eslint being slow means we need to hold tsc back. I also think the changes I made to make `--build` work are easy to maintain. If we add a new public package we just have to add it to the root tsconfig as a reference. Hence I think this is the way to go.


openspec/changes/archive/2026-07-23-eslint-tsconfig-references/design.md
```diff
@@ -0,0 +1,71 @@
+# Design: eslint-tsconfig-references
+
+`import-x/no-duplicates` alone takes **101.4s**, and all `import-x/*` rules combined account for 99.3% of rule time, confirming the resolver is the bottleneck. The current `project: '**/*/tsconfig.json'` glob matches 97 tsconfigs (45 after the `node_modules` ignore), and 35 of those are fixtures eslint never lints. The multi-project code path never caches because the cache is keyed by `configFile`, which stays `undefined` when `project` matches more than one tsconfig, so the resolver ends up re-running this glob 960 times per run.
+
+## Architecture Decisions
+
+### Decision: A `tsconfig.eslint.json` at repo root
+
+A new file, not a modification of the root `tsconfig.json` (which `lint:tsc` uses as a catch-all typecheck):
```
askoufis: I think using `tsconfig.eslint.json` as the root would speed up `tsc` as well. Having a separate tsconfig just for eslint feels a bit strange: they should both share tsconfigs if possible.

askoufis: This is what we do internally.

williamlark: Yea it would, but would need to change the script to use `tsc --buld` and modify the tsconfigs a bit which I tried to call out [here](https://github.com/seek-oss/sku/pull/1665/changes#diff-2b79e50306e4149815ec46ce006ac5c30779a6e76f142fdafdc59f0f896179e3R35). This gets us a step closer to doing that and fixes the eslint time, with minimal impact since I figured the alternative would be more effort. 

Though, I can just try the full migration and see what issues pop up.

openspec/changes/archive/2026-07-23-eslint-tsconfig-references/specs/lint-typescript-projects/spec.md
```diff
@@ -0,0 +1,56 @@
...
+- GIVEN the change is applied
+- WHEN `pnpm lint:tsc` runs
+- THEN the set of files typechecked is identical to before the change
+- AND the compiler options applied are identical to before the change
+
+#### Scenario: Build output is unchanged
+
+- GIVEN the change is applied
+- WHEN `pnpm build` runs
+- THEN the build outputs are byte-equivalent to before the change
```
jahredhope: I don't mind the other docs going a bit OTT but these specs are living docs that will be read by AI a lot, yet are really nailing home a non-consumer concern.

Ideally specs would have a consumer impact and these specs are more about our internal dev experience.

Would you be interested in cutting some of the bottom ones as being unnecessarily detailed? They aren't wrong, I'm only saying this because it's also low criticality.

williamlark: Yea I was thinking the same and was even considering deleting the specs entirely for this change and only leaving the design and proposal (which will eventually be archived) since it doesn't really impact any consumer facing apis or contain strong scenarios to keep around 💭 I'll finish up actioning Adams comment above and we can see what's most useful in this case

packages/codemod/tsconfig.json
```diff
@@ -1,17 +1,11 @@
...
-    "module": "nodenext",
-    "moduleResolution": "nodenext",
-    "allowSyntheticDefaultImports": true,
-    "resolveJsonModule": true,
-    "declaration": true,
     "types": ["node"]
   },
-  "include": ["./src/**/*", "tsdown.config.ts"],
-  "exclude": ["node_modules", "**/__testfixtures__"]
+  "include": ["./src/**/*", "./tsdown.config.ts", "./package.json"],
```
askoufis, line 9: Is `package.json` necessary here because we import it in the code?

eslint.config.js
```diff
@@ -142,7 +142,9 @@ export default [
     settings: {
       'import-x/resolver': {
         typescript: {
-          project: '**/*/tsconfig.json',
+          tsconfig: {
+            configFile: 'tsconfig.json',
+          },
```
askoufis: If this file is just the root level `tsconfig.json`, is this necessary anymore? I'd expect it to be the default value.

package.json
```diff
@@ -6,13 +6,12 @@
     "plop": "pnpm --filter '@sku-private/plop' plop",
     "postinstall": "node ./postinstall.js",
     "prepare": "husky",
+    "prelint": "echo 'Running lint...'",
```
askoufis: Was this leftover from testing, or do you want some kind of pre-lint log?

packages/codemod/tsconfig.json
williamlark: Yea it raises an error without it

eslint.config.js
williamlark: Yea, good catch. I found a clobbered rule and the react version was 18, so I've fixed those things up too.

package.json
williamlark: Yea was testing install hooks - not related at all haha I've removed it now.

askoufis (approved):

PR #1668, opened 2026-07-21
Combine fixture dist

williamlark:
Turns out we don't need to split fixtures into dist and dist-ssr/dist-build/dist-etc since vitest runs sequentially anyway. This removes some of the extra ignores we were injecting into the tests since they aren't required.

I combined the translations tests into the same file so that they don't run concurrently on the same fixture (it was only the only test that did that).

There was still .gitignore churn from .ssl being added, so I've added .ssl into the gitignore always instead of programatically since I don't see why anyone would want to commit it anyway and it's the simplest solution.


askoufis (approved):

williamlark:
I'm just merging some of the ssr server ports in the places they can be combined

packages/sku/src/utils/configureApp.ts
```diff
@@ -122,9 +98,10 @@ export default async (skuContext: SkuContext) => {
 
   // Generate self-signed certificate and ignore
   const selfSignedCertificateDirName = '.ssl';
+  // no reason to ever commit the certificate so we can always ignore it
+  gitIgnorePatterns.push(selfSignedCertificateDirName);
```
askoufis, line 102: Need a changeset for this.

askoufis (approved):

packages/sku/src/utils/configureApp.ts
williamlark: Done. patch or minor? I went patch since I see it as a fix rather than a feature

.changeset/three-trains-pay.md
```diff
@@ -0,0 +1,7 @@
+---
+'sku': patch
+---
+
+`configure`: `.ssl` always added to `.gitignore`
```
askoufis:
```suggestion
`configure`: Always add `.ssl` to `.gitignore`
```

packages/sku/src/utils/configureApp.ts
askoufis: Yeah I think `patch` is fine.

PR #1670, opened 2026-07-22
Initial Spec for Managed Data Mode SSR

jahredhope:
## Purpose

Add first-class SSR support with a new high-level API that better supports best practice whilst being built on the new vite bundler. See [Proposal](https://github.com/seek-oss/sku/blob/vite-ssr-sdd/openspec/changes/vite-ssr/proposal.md) for more.

## Process

This PR is being raised early for feedback on the specification before it gets merged ahead of implementations. Ideally this should reduce the size of the implementation review.

**SPECIFICATION ONLY** - To review the complexity of review we are looking to merge the initial specification FIRST. Reviewers are welcome to view the implementation itself at `vite-ssr-impl`

Reviewers are encouraged to:

1. Review the [Proposal](https://github.com/seek-oss/sku/blob/vite-ssr-sdd/openspec/changes/vite-ssr/proposal.md), for the high level purpose and scope of this change
2. Review the [Design](https://github.com/seek-oss/sku/blob/vite-ssr-sdd/openspec/changes/vite-ssr/design.md), for the various technical decisions made so far
3. Review the various [Specifications](https://github.com/seek-oss/sku/blob/vite-ssr-sdd/openspec/changes/vite-ssr/specs/vite-ssr/spec.md), for the living requirements that will exist after the implementation has been merged
4. Where desirable, review the tasks and implementation to build context for above. However please note that implementation is most likely to change, possibly being broken up into smaller PRs. Implementation is available in `vite-ssr-impl`.

For advice on reviewing OpenSpec Specs see: [Guide - Reviewing Changes](https://openspec.dev/docs/reviewing-changes)

### DIFF
--- a/openspec/changes/vite-ssr/design.md
+++ b/openspec/changes/vite-ssr/design.md
@@ -1,6 +1,6 @@
 ## Context
 
-SSR commands are blocked today.
+Vite SSR commands are blocked today.
 Webpack SSR uses Express `renderCallback`, string HTML, and CSP meta tags.
 
 This change introduces **Managed Data Mode**.
--- a/openspec/changes/vite-ssr/design.md
+++ b/openspec/changes/vite-ssr/design.md
@@ -1,0 +1,1455 @@
+## Context
+
+SSR commands are blocked today.
+Webpack SSR uses Express `renderCallback`, string HTML, and CSP meta tags.
+
+This change introduces **Managed Data Mode**.
+Sku owns the server, Document shell, streaming/hydration, assets, and CSP headers.
+Sku wires React Router Data Mode for routing and data.
+Apps own routes, data, and providers.
+
+Managed Data Mode is first shipped as **SSR** (`buildType: 'ssr'` on Vite).
+The same application contract is intended to underpin a future Static path.
+That contract includes `sku/runtime`, request entries, and hooks.
+Render strategy and API surface are separate concerns.
+See Decision 27.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- **Managed Data Mode** first shipped as SSR via `buildType: 'ssr'`.
+- Public import `sku/runtime`.
+- Public types and symbols without an `Ssr` infix.
+- Create template `ssr`.
+- First-class `routesEntry` with a named `routes` export.
+- First-class request-entry contracts.
+- First-class multi-site route trees via flat `routes`, optional `sites`, and `getSite`.
+- Same spirit as first-class multi-language.
+- Full-document streaming with `hydrateRoot(document, …)`.
+- Shell-derived CSP headers.
+- Per-route and vocab chunks via sku-owned `@vocab/vite` resolve.
+- Create template and Migrating docs.
+- Experimental first release.
+- Single config `port` only.
+- Config `serverPort` is rejected.
+- Named `Component` on lazy pages in template and docs.
+- CJS interop docs.
+- Accurate config JSDoc.
+- React Router 8 as an optional peerDependency `^8` for SSR consumers only.
+- No hard sku dependency on React Router.
+- Keep the shared Express dep on 4.
+- No 4 → 5 bump in this change.
+- No Jest transforms for React Router 8 in this change.
+- SSR requires Vitest.
+- Request entries `export default` one object via `defineServerEntry` / `defineClientEntry`.
+- Those helpers are zero-runtime inference helpers.
+- Optional sync getters receive Express `{ req }` where needed.
+- Getters include `getSite`, `getLanguage`, `getClientContext`, and `getReactContext`.
+- Optional entry hooks include `middleware`, `onListen`, `onHydrate`, and `getRouterContext`.
+- Always-on sku `SkuProvider` outside the router.
+- It carries `site`, serialised `clientContext`, and env `reactContext`.
+- Typed hooks via `createSkuContexts<typeof server, typeof client>()` on `sku/runtime`.
+- Sku does not wrap the route tree.
+- `createStaticHandler` is pre-built per site.
+- Three value channels: `getClientContext`, dual-entry `getReactContext`, and dual-entry `getRouterContext`.
+- `getClientContext` is the serialised isomorphic React seed.
+- `getReactContext` may differ per environment.
+- `getRouterContext` supplies values for React Router loaders and actions.
+- Later getters receive already-resolved sibling values.
+- `defineServerEntry` infers types from getter returns.
+- `defineClientEntry<typeof server>` extracts `Site` / `ClientContext` from the server entry.
+- Hooks read types from `typeof` the entry objects.
+- Router-aware isomorphic wrapping lives in the app’s own root layout route in `routesEntry`.
+- That covers Vocab, Apollo provider mount, and page chrome.
+- App-owned streaming data transports via `useInsertHtml` on `sku/runtime`.
+- The hook is nonce-able and a no-op off the SSR path.
+- Apollo streaming hydration is proven by a fixture.
+- Server-run queries are not refetched on hydrate.
+- Post-hydration queries still fetch.
+- Docs prefer render-time React data loading via Suspense.
+- Clients come from `useReactContext` / `useClientContext`.
+- Router-aware providers live in the app’s root layout route.
+- Loaders and `getRouterContext` are opt-in.
+- Docs include a red warning against putting Express `req` into router context.
+- Migrating guidance covers server-only loaders, Braid reset order, client-only libraries via `getReactContext`, Jest → Vitest, and `#` `pathAliases`.
+- Config `polyfills` applies to the SSR browser client.
+- SSR `sku start` covers SSR-CSS via a virtual stylesheet on Document assets.
+- No `transformIndexHtml` on the SSR path.
+- SSR `sku start` telemetry parity for `start.initial` / `start.rebuild`.
+- Optional server-entry `onListen` covers the same post-`listen` window as webpack `onStart`.
+- Opt-in config `expressTrustProxy` is a boolean that sets Express hop count `1`.
+- The create template sets `expressTrustProxy: true`.
+
+**Non-Goals:**
+
+- Webpack mode for this `buildType`, or a Webpack SSR backfill / an updated Webpack-SSR create template.
+- Converting the static `vite` create template.
+- Full infra/deploy product guides (keep sku’s existing docs scope).
+- Framework Mode / RSC.
+- Absolute / `CDN` `publicPath`.
+- First-class React Router basename config.
+- SSR `serverPort`.
+- Expanding baked-in CJS interop defaults beyond Apollo.
+- Consumer Document override.
+- Runtime server↔client route-tree equality checking.
+- Dual-entry `routes` re-exports from `serverEntry` / `clientEntry`.
+- Differing server vs client route modules as a product feature.
+- Sku-owned site resolution from config `hosts` / `sites[].host` (local-dev listen/setup only).
+- Sku-owned per-site path expansion libraries (apps own path shape; sku owns membership filtering).
+- Requiring a non-empty config `sites` array for SSR (empty soft-defaults to `'default'`).
+- Per-site JS bundles.
+- Returning routes from a request-entry getter or bag.
+- A combined request-entry resolver returning site + language + `clientContext` values (see Decision 12).
+- Returning a provider component from request-entry getters (see Decision 12).
+- Sku reading site / language / `clientContext` from a conventional `req` field, or a sku-provided push API such as `setRequestContext(req, …)`.
+- Tolerating a missing `serverEntry` / `clientEntry` file (paths still resolve via normal module resolution).
+- Per-getter named exports on request entries or `defineGet*` per-property helpers.
+- Mounting sku’s provider inside the route tree as a pathless layout (see Decision 12).
+- Consumer-authored Async Local Storage as the documented way to reach request state from React.
+- App-authored dual-entry `Providers` / `SkuProvidersProps` (see Decision 12a).
+- A public `useInitialLanguage` / language-in-React-context hook in v1.
+- Union route tree + site allowlist middleware as the documented multi-site product story.
+- Parent → child inheritance of `sites` (must set explicitly on each divergent route).
+- Overloading config `routes` (static prerender path lists) as the SSR `RouteObject` entry.
+- Sku-owned listen logging by default (apps log in `onListen`).
+- An `onBeforeListen` server-entry hook (module top-level covers pre-bind setup).
+- Soft-defaulting Express `trust proxy` for SSR without config (opt-in via `expressTrustProxy`).
+- Supporting the config `public` assets folder for SSR (until a definitive need).
+- Automatic `*.server.ts` client strip.
+- Auto-injecting Braid reset into sku’s SSR server entry.
+- A new Jest → Vitest codemod beyond existing tooling / docs.
+- Making Express `req` the loader `request` argument (stays Fetch `Request`).
+- Treating Framework Mode’s server-only `getLoadContext(req, res)` as sufficient for sku Data Mode.
+- Requiring `getClientContext`, `getReactContext`, or `getRouterContext` (all optional; omit → `undefined` / empty defaults).
+- Requiring `middleware`, `onListen`, or `onHydrate` (all optional).
+- Requiring `getSite` when config has zero or one site (sku uses the sole resolved name).
+- Passing Fetch `Request` into `getSite` / `getLanguage` / `getClientContext` (Express `req` only).
+- Passing `res` into getters / `getReactContext` / `getRouterContext` in v1.
+- Async request-entry getters (sync-only; `getRouterContext` MAY still be async because it seeds loader context).
+- Treating raw Express `req` (or other non-isomorphic platform objects) in `RouterContextProvider` as a supported pattern.
+- Shipping Jest support for React Router 8 (transforms / ESM interop).
+- Forcing webpack fixtures or non–Vite-SSR apps onto React Router 8.
+- Treating React Router loaders as the default teaching path for page content.
+- Upgrading sku’s shared Express dependency from 4 → 5 (deferred; would break webpack SSR).
+- Supporting `@sku-lib/vite/loadable` (Collector / `LoadableProvider` / `preloadPlugin` module-id injection) as an SSR document-preload source.
+- Supporting `dangerouslySetViteConfig` or `vitePlugins` for SSR (static Vite unchanged).
+- A sku-owned Apollo dependency, provider, config option, or version pin (sku ships `useInsertHtml` and apps own the client / transport).
+- `@apollo/client-integration-react-router`’s loader transport (see Decision 21a).
+- Streaming (turbo-stream) loader-data serialization to carry transported query refs.
+- Two-pass `getDataFromTree` SSR (incompatible with streaming).
+- Sku auto-attaching the CSP nonce to app-injected scripts (transports expose their own script props).
+- Consumer Vite config injection for SSR module identity (sku owns `optimizeDeps.exclude`).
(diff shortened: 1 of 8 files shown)


openspec/changes/vite-ssr/specs/managed-data-mode/spec.md
```diff
@@ -0,0 +1,456 @@
...
+Omitting `onHydrate` MUST mean no hydrate side effects (not an error).
+
+Omitting `middleware` MUST mean no consumer middleware layer (not an error).
+
+Omitting `onListen` MUST mean no post-listen callback (not an error).
+SSR defines `onListen` call timing and failure behaviour.
+
+Sku MUST always render `SkuProvider` outside the router — `Document` → `SkuProvider` → router — with `site`, `clientContext`, and `reactContext` for that document.
+
+Sku MUST export `createSkuContexts<typeof server, typeof client>()` from `sku/runtime` so apps can obtain typed `useSite` / `useClientContext` / `useReactContext` bound to that provider.
```
askoufis, line 214: I'd want to look at what `sku/runtime` actually exports so I can better judge whether that name is appropriate. The main question is have is would that entrypoint be useful for SSG (in the future), or should it be scoped to SSR right now (instead of later), or does it not matter?

openspec/changes/vite-ssr/specs/managed-data-mode/spec.md
```diff
@@ -0,0 +1,456 @@
...
+- **WHEN** the server entry omits `getSite` (single-site)
+- **THEN** `useSite()` is typed as `string`
+
+### Requirement: Shared Managed Data Mode modules keep one identity under Vite
+
+App code that imports shared Managed Data Mode state from `sku/runtime` (hooks from `createSkuContexts`, `useInsertHtml`, `usePreloadRoute`, CSP nonce helpers) and sku’s own Managed Data Mode runtime (`SkuProvider`, insert-html queue/provider, preload registry, request-context runner) MUST observe the **same** module instances.
+
+Sku MUST:
+
+1. Prefer the public `sku/runtime` specifier for sku runtime call sites that touch that shared state (re-export `@internal` symbols as needed), and
```
askoufis: Brought this up briefly yesterday, but if the intention is to have a stable entrypoint for _private_ API, a subpath import should be preferred over a public export with APIs marked as `@internal`.

openspec/changes/vite-ssr/specs/ssr/spec.md
```diff
@@ -0,0 +1,832 @@
...
+
+- **WHEN** React Router records a route error on the static handler context
+- **THEN** the HTML response status is `context.statusCode`
+- **AND** the body is the nearest `ErrorBoundary`
+
+### Requirement: Hydration bootstrap is production-safe
+
+Sku MUST scrub Promises from loader/action data before bootstrap stringify.
+
+Sku MUST omit `Error.stack` from production hydration error payloads.
```
askoufis, line 209: Keen to understand this better. Wouldn't this make it harder to debug hydration issues in prod, or is that not relevant to this statement?

openspec/changes/vite-ssr/specs/ssr/spec.md
```diff
@@ -0,0 +1,832 @@
...
+
+When `languages` is configured for build-time vocab splitting, sku MUST register a vocab language chunk on the Document only when server `getLanguage` returns `language`.
+
+Language is optional: if `getLanguage` is omitted or returns `undefined`, sku MUST NOT register a language chunk.
+
+Sku MUST NOT validate the returned language against config and MUST NOT default to a sole configured language.
+
+Language MUST be server-local for Document registration only (no client forward, no `getSkuLanguage` / `__SKU_LANGUAGE__` / baked languages define).
+
+When vocab / `languages` is active, sku MUST resolve `@vocab/vite` from sku’s install tree and MUST alias bare `@vocab/vite/runtime` (prefer the export file) onto that copy in the shared Vite `resolve.alias` (client and SSR).
```
askoufis, line 387: This might be worth porting over to Vite SSG. I _think_ at the moment Vite SSG users need to install `@vocab/vite`.

openspec/changes/vite-ssr/design.md
askoufis, line 1: This doc is too long to read in a normal amount of time. I've raised any questions I had in the spec rather than digging into the design specifically, LMK if I've missed something by not reading the exact design notes for anything I commented on.

openspec/changes/vite-ssr/specs/managed-data-mode/spec.md
```diff
@@ -0,0 +1,456 @@
...
+- **AND** does not choose the tree from the request `Host` header alone
+
+### Requirement: Optional server and client request exports
+
+`serverEntry` / `clientEntry` MUST each **`export default`** one object.
+Sku MUST read that default export and call optional properties on it.
+
+Sku MUST export `defineServerEntry` / `defineClientEntry` from `sku/runtime` as zero-runtime identity helpers that infer types from getter returns and type later sibling args (`NoInfer` on input positions).
+`defineServerEntry` MUST infer `Site` from `getSite`, `Language` from `getLanguage`, `ClientContext` from `getClientContext`, and `ReactContext` from `getReactContext`, and MUST type later sibling `site` args as that `Site`.
+`defineClientEntry` MUST accept an optional `ServerEntry` type argument (`defineClientEntry<typeof server>`).
```
SpiralLogic, line 169: some of these requirements seem like they are implementation details as opposed to observable behaviours. should they be in the design document? for example this one and the NoInfer above.

openspec/changes/vite-ssr/specs/ssr/spec.md
```diff
@@ -0,0 +1,832 @@
...
+
+#### Scenario: Stream path skips HTML transform
+
+- **WHEN** an SSR document response is generated
+- **THEN** the body is the React render stream
+- **AND** `transformIndexHtml` is not invoked for that response
+
+### Requirement: SSR start injects collected SSR CSS via Document assets
+
+On SSR `sku start`, sku MUST mount `vitePluginSsrCss` (or equivalent) so CSS reachable from the SSR module graph is available as the virtual stylesheet used by static Vite start.
```
SpiralLogic, line 109: same thing here, these are more implementation detail than observable behaviour. could be more something like:

```On sku start, streamed documents for CSS-importing apps MUST include a stylesheet link for collected SSR CSS, and that link MUST update correctly under HMR. Production documents MUST get CSS from the client manifest path instead.```

File: openspec/changes/vite-ssr/specs/managed-data-mode/spec.md (hunk 5 above)
Comment (jahredhope) [reply openspec/changes/vite-ssr/specs/managed-data-mode/spec.md:169]: It's a questionable call in my view.

If it was an internal function I'd agree.

In this case it's not only an exposed integration point, it's a very critical one. Not only the function signatures, but the inference that is later used to build type safe hooks into user context.

It's still questionably detailed, especially everything in brackets (`(NoInfer on input positions)`, but I'd still side on keeping them for the same reason that they exist: When they weren't specified the solution was flawed and the user experience suffered.

File: openspec/changes/vite-ssr/tasks.md [hunk 7]
```diff
@@ -0,0 +1,125 @@
...
+- [ ] 2.3 Add `optimizeDeps.exclude` for `'sku'` and `'sku/runtime'` (shared constant). Assert it in unit tests.
+- [ ] 2.4 Import `virtual:sku/polyfills` at the top of the SSR browser client entry only.
+
+## 3. `routesEntry` and site-scoped trees
+
+- [ ] 3.1 Add config `routesEntry` (default `src/routes.tsx`). Alias `__sku_alias__routesEntry` into both SSR graphs.
+- [ ] 3.2 Export `SkuRouteObject = RouteObject & { sites?: string[] }`. Require named `routes` array. Hard-error when missing or non-array.
+- [ ] 3.3 Soft-default empty/omitted config `sites` to `['default']`. Keep multi-site behaviour when `sites.length > 1`.
+- [ ] 3.4 At init, pre-build one route tree per resolved site name. Omit `sites` ⇒ all sites. No parent→child inheritance. Strip `sites` before React Router.
+- [ ] 3.5 Fail closed on missing/invalid/unknown site. Do not select the tree from config hosts.
```
SpiralLogic, line 30: what does fail-closed unknown site mean? Is there a definition for the observable behaviour of this?

openspec/changes/vite-ssr/specs/managed-data-mode/spec.md
```diff
@@ -0,0 +1,456 @@
+## ADDED Requirements
+
+### Requirement: Managed Data Mode naming
+
+Product docs and public APIs MUST describe this architecture as **Managed Data Mode** (sku-owned Document + React Router Data Mode wiring; apps own routes, data, and providers).
+
+**SSR** MUST refer only to the render strategy selected by `buildType: 'ssr'`.
+
+Docs MUST note that Managed Data Mode is intended to be shared later by a new Static path on the same contract.
```
SpiralLogic, line 9: should docs related things be in tasks rather than specs?

openspec/changes/vite-ssr/tasks.md
jahredhope: It means throw an error if there is no sku config site for the site returned.

If `getSite` returns "banana", when that site isn't configured in sku.config.ts, the request should immediately return an error. Rather than falling back to a common tree.

The alternative option considered was to build up a known common site tree (all routes that have no sites array), this would have space improvements, (as many sites will all share the same tree), and allow a default/unknown tree. But I felt it was better to make it a hard error and optimise towards fixing the invalid code.

openspec/changes/vite-ssr/specs/managed-data-mode/spec.md
jahredhope: Can docs? Yes.
Should docs? Sometimes.

In this case it isn't a great argument for it being in the spec, but it was a clear decision made during the specification flow to setup managed data mode as "For SSR FOR NOW", not "Only for SSR". So it's not bad that there is a spec clearly showing a decision that will impact user experience (not code experience, but the docs are an important part of the sku offering)

openspec/changes/vite-ssr/specs/managed-data-mode/spec.md
jahredhope: Yes I expected it to be used for static. This was the key reason to move away from `sku/ssr`

openspec/changes/vite-ssr/specs/ssr/spec.md
jahredhope: To avoid confusion, this isn't referring to hydration errors, as in "the client tried to hydrate and things went wrong".

But the payload the loader/action returned so that the client can appropriate hydrate, specifically an error payload as the loader/action errorred.

This is similar (slightly different) to what React Router Framework Mode does. Though it goes further and makes the error itself generic, something we might want to consider doing. See [React Router Framework Mode - Error Sanitisation](https://reactrouter.com/how-to/error-boundary#error-sanitization).

openspec/changes/vite-ssr/specs/ssr/spec.md
jahredhope: I agree. Will raise this as a separate issue.

openspec/changes/vite-ssr/specs/managed-data-mode/spec.md
jahredhope: Thanks, yes. I've updated the spec, it's a bit more specific than it should be, but I don't want it to be forgotten when it comes to adapting the implementation.

I also haven't tested whether the implementation will actually hold.

openspec/changes/vite-ssr/specs/ssr/spec.md
jahredhope: Addressed in https://github.com/seek-oss/sku/pull/1670#discussion_r3709394051

openspec/changes/vite-ssr/design.md
```diff
@@ -0,0 +1,1466 @@
+## Context
+
+SSR commands are blocked today.
```
williamlark:
```suggestion
Vite SSR commands are blocked today.
```

openspec/changes/vite-ssr/design.md
```diff
@@ -0,0 +1,1466 @@
...
+This change introduces **Managed Data Mode**.
+Sku owns the server, Document shell, streaming/hydration, assets, and CSP headers.
+Sku wires React Router Data Mode for routing and data.
+Apps own routes, data, and providers.
+
+Managed Data Mode is first shipped as **SSR** (`buildType: 'ssr'` on Vite).
+The same application contract is intended to underpin a future Static path.
+That contract includes `sku/runtime`, request entries, and hooks.
+Render strategy and API surface are separate concerns.
+See Decision 27.
```
williamlark, line 15: Not sure what Decision 27 is referring to. Needed to reference here at all?

openspec/changes/vite-ssr/proposal.md
```diff
@@ -0,0 +1,129 @@
...
+Webpack SSR’s low-level `renderCallback` + string document + `#app` hydrate cannot stream Suspense HTML.
+It also cannot let React own `<html>`/`<head>`/`<body>`.
+
+This change introduces **Managed Data Mode**.
+Sku owns Document, streaming/hydration, the Node server, and CSP.
+Sku wires React Router Data Mode for routing and data.
+Apps own routes, data, and providers.
+
+Managed Data Mode is first shipped as **SSR** (`buildType: 'ssr'`).
+The same contract is intended to be shareable later by a new Static path.
```
williamlark, line 13: What does "a new Static path" mean?

openspec/changes/vite-ssr/proposal.md
```diff
@@ -0,0 +1,129 @@
...
+That contract includes `sku/runtime`, request entries, and typed context hooks.
+Render strategy (`ssr` / `static`) and the Managed Data Mode API are separate concerns.
+It ships **experimental** (not for production).
+
+Multi-site apps need different path sets per site.
+A single unfiltered route list either over-matches or registers foreign paths on every host.
+This change makes site-scoped trees first-class via `routesEntry` + optional route `sites` + app `getSite`.
```
williamlark, line 20: Is `routesEntry` a react-router property or a new sku config option?

Is this implying that there will be three different places routes and sites can be configured?

openspec/changes/vite-ssr/design.md
```diff
@@ -0,0 +1,1466 @@
...
+
+### 4a. Flat `routes` + optional `sites` → pre-built site trees
+
+Multi-site apps need different React Router path sets per site (e.g. site-only pages).
+A single unfiltered `RouteObject[]` either over-matches unsupported paths or registers foreign paths on every host.
+
+Config `hosts` / `sites[].host` are **local-dev listen and setup-hosts only**.
+Sku MUST NOT derive production (or request) site from them for route-tree selection.
+
+**Apps own** site resolution (from Express `req`, headers, app config, etc.) via sync `getSite({ req })`, and per-site **path shape** when paths differ by site (factories remain fine for `/jobs` vs `/emploi`).
```
williamlark, line 259: ```factories remain fine for `/jobs` vs `/emploi`)```

What does this mean?

SpiralLogic (approved):

openspec/changes/vite-ssr/proposal.md
```diff
@@ -0,0 +1,129 @@
...
+Env-differing values come from dual-entry `getReactContext`.
+Isomorphic mounts (Apollo, Vocab) live in the app’s root layout route.
+Details: `design.md` Decision 12a.
+
+## What Changes
+
+- Experimental SSR ships with Vite `buildType: 'ssr'`.
+- `buildType: 'static'` keeps today’s static Vite behaviour.
+- A future Managed Data Mode Static path is out of scope for this change.
+- Commands are `sku start` / `sku build`.
```
williamlark, line 47:
```suggestion
- Commands are `sku start` / `sku build`, deprecating `sku start-ssr`/`sku build-ssr`. start-ssr and build-ssr will still work while in experimental phase, but will funnel into the new buildType logic internally.
```

This is assuming that the buildType change doesn't have to be vite only. build-srr and start-ssr would just add the `buildType: 'ssr' to the config, and then `if buildType is ssr and bundler is webpack do current webpack stuff`. Could mark *-ssr commands as deprecated right off the bat if we wanted though.

openspec/changes/vite-ssr/design.md
jahredhope: The numbered headings of this document are all decisions.
[Decision 27.](https://github.com/seek-oss/sku/blob/9e76092681057ed1ed2b41ec8b00fd2d0fa440ca/openspec/changes/vite-ssr/design.md#27-naming-managed-data-mode-skuruntime-ssr-render-type)

This is where we cover naming decisions, such as the use of `sku/runtime` over `sku/ssr` or `sku/app`

openspec/changes/vite-ssr/proposal.md
jahredhope: Managed Data Mode Static

As in: a new type of app the consumer can choose that is Static instead of SSR.

openspec/changes/vite-ssr/proposal.md
jahredhope: `routesEntry` is one of the three main entries of a Managed Data Mode SSR app.

- Server — Server specific stuff
- Client — Client specific stuff
- Routes — Isomorphic stuff

It's probably easier to read the docs: https://github.com/seek-oss/sku/blob/vite-ssr-impl/site/docs/ssr/entries.md

openspec/changes/vite-ssr/design.md
jahredhope: The `sites` field solves a common problem of many apps where the app wants to have some routes available on some other routes.

However, it only helps when you are toggling a route on and off by site, and not another less common problem:

We use English in URLs (`/jobs`). But what if you really care about the translated experience and want to translate the URL (`/emploi` in French). Effectively you want to change the path of a route based on the language. `sites` doesn't help with that.

A factory (`const routes = createRoutes()`) is about the closest thing we have to support for that scenario, but it really doesn't help much.

openspec/changes/vite-ssr/proposal.md
jahredhope: The proposal currently is to make `buildType: ssr` inherently the Managed Data Mode SSR and ONLY supported by Vite.

`sku start-ssr`/`sku build-ssr` aren't deprecated separately to webpack-ssr, they are inherently tied to the old API which is only supported by webpack-ssr.

A future change MAY decide to back port non-Managed Data Mode to vite,  but that's currently an explicit non-goal for this change.

askoufis (approved):

openspec/changes/vite-ssr/specs/managed-data-mode/spec.md
```diff
@@ -0,0 +1,463 @@
...
+- **WHEN** a route sets `sites: ['au']`
+- **AND** config defines sites `au` and `nz`
+- **THEN** the route is present only in the `au` pre-built tree
+
+#### Scenario: No sites inheritance from parent
+
+- **WHEN** a parent route sets `sites: ['au']`
+- **AND** a child omits `sites`
+- **THEN** sku does not copy the parent’s `sites` onto the child
+- **AND** structural exclusion still applies (if the parent is absent for a site, its subtree is absent)
```
williamlark, line 113: If it's not inherited, does the child not display at all unless its sites are specified?

openspec/changes/vite-ssr/specs/managed-data-mode/spec.md
```diff
@@ -0,0 +1,463 @@
...
+#### Scenario: Multi-site missing getSite hard-errors at init
+
+- **WHEN** config defines more than one site
+- **AND** the server entry omits `getSite`
+- **THEN** sku fails with a hard error at init naming the `getSite` property
+
+#### Scenario: Non-string getSite fails closed
+
+- **WHEN** `getSite` returns a non-string value
+- **THEN** sku fails closed with a hard error for that request
```
williamlark, line 143: Guessing this will be typed check on top of this runtime check?

PR #1674, opened 2026-07-29
Improve package manager detection

williamlark:
I've refactored the package manager detection to reduce warning when running within an agent. Previously, if you ran skus binary directly from node_modules it would lead to `npm_config_user_agent` not being set. Sku would fall back to npm if this was the case, leading to errors in the terminal.

Obvisouly this is not how people should be running sku, but this can happen with AI agents, and there are other ways for them to run sku without setting a npm_config_user_agent due to sandbox limitations. Ai agent output was how this issue was found.

An example of previous broken output, tested by building the braid fixture with the sku bin directly (`./node_modules/.bin/sku`). This could also happen  inside an AI agent sandbox environment in a few different ways.

CleanShot 2026-07-29 at 17 32 01

 <br>
 <br>

After the changes, there  are no errors and we get correct telemetry installation output.

CleanShot 2026-07-29 at 17 36 30

## Code logic changes
- Resolution now uses `package-manager-detector`, which detects the **projects** package manager from it's `packageManager` field or  lockfile. That is now the source of truth for any command run inside a project.
- The **running** package manager is only a fallback for when no project is detected, which is the case for `@sku-lib/create`, where a fresh directory has no lockfile. If neither is available, it still defaults to npm.
- The "Lockfile mismatch" warning became a mismatch between the *running* and *project* package managers, and it only fires when a user agent is actually present.
- `packageManagerVersion` prefers the running agent's version when the running agent matches the project's (so it reflects the actual binary in use), then falls back to the version in the `packageManager` field. 
- A detected but unsupported manager (bun or deno) now throws the existing "Unsupported package manager" error rather than silently behaving like npm.
- I didn't add any tests for this since a pretty tricky one to test nicely without mocking a whole bunch of stuff. I don't think it's worth the effort for the value it brings

### DIFF
--- a/.changeset/wild-moons-attend.md
+++ b/.changeset/wild-moons-attend.md
@@ -1,0 +1,8 @@
+---
+'sku': patch
+'@sku-lib/create': patch
+---
+
+Improve package manager detection
+
+Sku now works out which package manager your project uses from its `packageManager` field or lockfile, before falling back to the package manager that invoked sku. This should improve the accuracy of sku's output when running from a coding agent.
--- a/.changeset/wild-moons-attend.md
+++ b/.changeset/wild-moons-attend.md
@@ -5,4 +5,4 @@
 
 Improve package manager detection
 
-Sku now works out which package manager your project uses from its `packageManager` field or lockfile, before falling back to the package manager that invoked sku. This should improve the accuracy of sku's output when running from a coding agent.
+The package manager detection has been improved, and should now correctly detect the package manager your project uses, even when running from a coding agent.
--- a/.changeset/wild-moons-attend.md
+++ b/.changeset/wild-moons-attend.md
@@ -5,4 +5,4 @@
 
 Improve package manager detection
 
-The package manager detection has been improved, and should now correctly detect the package manager your project uses, even when running from a coding agent.
+Package manager detection has been improved, and should now correctly detect the package manager your project uses, even when running from a coding agent.
--- a/private/utils/src/packageManager/packageManager.ts
+++ b/private/utils/src/packageManager/packageManager.ts
@@ -69,6 +69,46 @@ const getFallbackPackageManager = (): SupportedPackageManager => {
   return fallback;
 };
 
+const resolveDetectedPackageManager = () => {
+  const projectPackageManager = getProjectPackageManager();
+  const runningPackageManager = getRunningPackageManager();
+
+  const name =
+    projectPackageManager?.name ??
+    runningPackageManager?.name ??
+    getFallbackPackageManager(); // some edge cases might use this fallback, but unlikely
+
+  if (!isSupportedPackageManager(name)) {
+    throw new Error(
+      `Unsupported package manager: ${name}. Supported package managers are: ${supportedPackageManagers.join(
+        ', ',
+      )}`,
+    );
+  }
+
+  const isRunningResolvedPackageManager = runningPackageManager?.name === name;
+
+  if (
+    projectPackageManager &&
+    runningPackageManager &&
+    !isRunningResolvedPackageManager
+  ) {
+    console.warn(
+      caution(
+        `Package manager mismatch: sku was run with ${strong(runningPackageManager.name)}, but this project uses ${strong(name)}`,
+      ),
+    );
+  }
+
+  // The running version is the most accurate source of the version, so we use it if it's available.
+  const version =
+    (isRunningResolvedPackageManager ? runningPackageManager.version : null) ??
+    projectPackageManager?.version ??
+    null;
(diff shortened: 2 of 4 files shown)


kjots:
I think the title of this PR should reflect the nature of the change (e.g. "Improved package manager detection") rather than one of the motivating reasons for it - there is really nothing "agent" related to this change, the detection algorithm is just more comprehensive and will resolve a number of other use cases (e.g. running tests via an IDE's built-in test runner, which also does not make use of a package manager to launch `vitest` (ask me how I know)).

williamlark:
> I think the title of this PR should reflect the nature of the change
> 

Yea can do. The description evolved a bit to be less about AI agents but I didn't update the title to reflect that

private/utils/src/packageManager/packageManager.ts
```diff
@@ -1,85 +1,129 @@
...
 import { resolveCommand } from 'package-manager-detector/commands';
 import { INSTALL_PAGE } from 'package-manager-detector/constants';
+import { detectSync, getUserAgent } from 'package-manager-detector/detect';
 import semver from 'semver';
 import { caution, strong } from '../console/styles.ts';
 
 export type SupportedPackageManager = 'yarn' | 'pnpm' | 'npm';
 
-const supportedPackageManagers = ['yarn', 'pnpm', 'npm'];
+const supportedPackageManagers: SupportedPackageManager[] = [
```
kjots: These could be derived from the `lockfileByPackageManager` record to reduce duplication, e.g.:

```typescript
const lockfileByPackageManager = {
  pnpm: 'pnpm-lock.yaml',
  yarn: 'yarn.lock',
  npm: 'package-lock.json',
} as const satisfies Record<string, string>;

export type SupportedPackageManager = keyof typeof lockfileByPackageManager;
const supportedPackageManagers = Object.keys(lockfileByPackageManager) as SupportedPackageManager[];
```

private/utils/src/packageManager/packageManager.ts
```diff
@@ -1,85 +1,129 @@
...
   }
 
-  if (userAgent.includes('yarn')) {
-    packageManager = 'yarn';
-  }
+  // User agents typically look like `pnpm/9.12.1 npm/? node/v20.17.0 linux x64`
+  const version =
+    process.env.npm_config_user_agent?.split(' ')[0].split('/')[1] || null;
+
+  return { name, version };
```
askoufis: `getUserAgent` does the same thing as line 43 but doesn't return the version. Is there any reason to double up on this?

private/utils/src/packageManager/packageManager.ts
```diff
@@ -1,85 +1,129 @@
...
- */
-const resolvePackageManager = () => {
-  const userAgentPackageManager = getPackageManagerFromUserAgent();
-  const packageManager = validatePackageManager(
-    userAgentPackageManager.packageManager,
+const getFallbackPackageManager = (): SupportedPackageManager => {
+  const fallback = 'npm';
+  console.warn(
+    caution(`No package manager detected, assuming ${strong(fallback)}`),
   );
```
askoufis, line 69: I see that `console.warn` was already used in this file before your change, but it's kind of redundant here because `caution` already applies the styling we want.

williamlark: The other difference is that they go to different channels. Might make a difference if  an agent is tailing stderr or something. console.warn does feel more correct here anyway

private/utils/src/packageManager/packageManager.ts
williamlark: True, I didn't realise it did the same thing. I've merged them together with the logic from package-manager-detector (checking the AGENTS constant)

private/utils/src/packageManager/packageManager.ts
askoufis: I guess if we want it to go to stderr I'd rather use `console.error`, but I'm not too concerned either way.

williamlark: I think i'd prefer to keep it as is.

.changeset/wild-moons-attend.md
```diff
@@ -0,0 +1,8 @@
+---
+'sku': patch
+'@sku-lib/create': patch
+---
+
+Improve package manager detection
+
+The package manager detection has been improved, and should now correctly detect the package manager your project uses, even when running from a coding agent.
```
askoufis:
```suggestion
Package manager detection has been improved, and should now correctly detect the package manager your project uses, even when running from a coding agent.
```

askoufis (approved):

PR #1675, opened 2026-07-30
Add support for `report-to` Content Security Policy directive

kjots:
This change adds two new configuration options, `cspReportTo` and `cspReportOnlyReportTo`, that allow the reporting of CSP violations via the browser [Reporting API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API).


packages/sku/src/services/vite/helpers/prerender/prerenderWorker.ts
```diff
@@ -94,17 +115,31 @@ await Promise.all(
 
         if (cspHandler) {
           const root = cspHandler.processHtml(html);
+          const reportingEndpoints: ReportingEndpoint[] = [];
 
           if (cspEnabled) {
             if (cspDelivery === 'tag') {
               html = cspHandler.updateHtml(root);
```
SpiralLogic, line 122: if a reportTo is configured and also tag is chosen, won't updateHtml which calls createCSP still append report-to on the meta tags?

site/docs/csp.md
```diff
@@ -25,6 +25,22 @@ The `cspDelivery` option is only available when using Vite.
 
 If you need to allow scripts that are only known client side (e.g. scripts loaded by tag managers) you can add their URLs to the `cspExtraScriptSrcHosts` array in `sku.config.js`.
 
+### Report To
+
+The `cspReportTo` option allows reports of CSP violations to be captured via the browser [Reporting API].
+This option can be configured either as an _endpoint name_, a _URL_, or as a tuple of both, with the following outcome:
+
+- If only an _endpoint name_ is specified, then this value will be included in the CSP as the value of the [`report-to`] directive, and no [`Reporting-Endpoints`] header will be emitted.
+- If only a _URL_ specified, then an endpoint name will be generated automatically and included in the CSP as the value of the [`report-to`] directive, and a [`Reporting-Endpoints`] header will be emitted containing the generated endpoint name and the provided URL.
```
SpiralLogic:
```suggestion
- If only a _URL_ is specified, then an endpoint name will be generated automatically and included in the CSP as the value of the [`report-to`] directive, and a [`Reporting-Endpoints`] header will be emitted containing the generated endpoint name and the provided URL.
```

site/docs/csp.md
```diff
@@ -25,6 +25,22 @@ The `cspDelivery` option is only available when using Vite.
...
+- If both an _endpoint name_ and a _URL_ is specified, then the provided endpoint name will be be included in the CSP as the value of the [`report-to`] directive, and a [`Reporting-Endpoints`] header will be emitted containing both the provided endpoint name and URL.
```
SpiralLogic:
```suggestion
- If both an _endpoint name_ and a _URL_ is specified, then the provided endpoint name will be included in the CSP as the value of the [`report-to`] directive, and a [`Reporting-Endpoints`] header will be emitted containing both the provided endpoint name and URL.
```

SpiralLogic (approved):

packages/sku/src/services/vite/helpers/prerender/prerenderWorker.ts
kjots: Hmm, yes, `report-to` would be included in the CSP but no `Reporting-Endpoints` header would be emitted. I'll fix this.

kjots: This should be fixed now, for both the [middware](https://github.com/seek-oss/sku/blob/ed11bce05a7e74f80cb63b5debbca6a65b8f619b/packages/sku/src/services/vite/plugins/middleware.ts#L121) and the [worker](https://github.com/seek-oss/sku/blob/ed11bce05a7e74f80cb63b5debbca6a65b8f619b/packages/sku/src/services/vite/helpers/prerender/prerenderWorker.ts#L84) the `cspReportTo` value will only be passed to the handler when `cspDelivery === 'header'`.

SpiralLogic:
all of the new imports are using .ts should they use .js to be consistent?

kjots:
> all of the new imports are using .ts should they use .js to be consistent?

I didn't even notice that, they both appear to work regardless.

For this change, yes they should be `.js` since that is what all the existing imports use, however I would suggest that we should at some point update to use `.ts` since that is what is actually being imported (the compiled `dist` files appear to all be `.mjs` so `.js` is even more wrong than `.ts` in my opinion).

Either way, we should probably put a linting rule in place to ensure the correct extension is used at all times.

SpiralLogic (approved):

jahredhope (approved):

PR #1676, opened 2026-07-31
storybook: Add `viteFinal` API for storybook config, update docs

askoufis:
Added a `viteFinal` API so consumers that use Storybook have a migration path off webpack. Updated both the storybook and vite docs appropriately, and added some tests to the `storybook-config` suite that test the new API (`tests/browser/storybook-config.test.ts` changes are best viewed with `Hide whitespace` enabled).


site/docs/storybook.md
```diff
@@ -1,24 +1,58 @@
...
 
+[Storybook]: https://storybook.js.org/
+
 ## Installing Storybook Dependencies
 
-To set up Storybook, you will need to install the following dev dependencies:
+Install the appropriate dependencies for your configured [`bundler`]:
 
-```sh
+::: code-group
```
williamlark, line 12: This is nice ✨

williamlark (approved):

packages/sku/src/config/storybook/viteConfig.ts
```diff
@@ -0,0 +1,44 @@
+import { mergeConfig } from 'vite';
+import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';
+import { fixViteVanillaExtractDepScanPlugin } from '../../services/vite/plugins/esbuild/fixViteVanillaExtractDepScanPlugin.ts';
```
SpiralLogic: should this import a js to be consistent with everywhere else?

SpiralLogic (approved):

packages/sku/src/config/storybook/viteConfig.ts
askoufis: I'm not too fussed about consistency, but I've fixed it anyway.

PR #1685, opened 2026-08-10
Use brand accent for code block highlight color, modernize CSS

askoufis:
Inspired by @jahredhope being annoyed about the grey highlight color, I customized the color so it's now based off a new `--brand-soft-base` token. Also took the opportunity to derive other colors from that token too, as well as use some modern CSS syntax where practical.

Before:

image

After:

image

### DIFF
--- a/site/.vitepress/theme/custom.css
+++ b/site/.vitepress/theme/custom.css
@@ -1,26 +1,26 @@
 :root {
-  --vp-c-brand-1: #8300fc;
-  --vp-c-brand-2: #6a00d1;
-  --vp-c-brand-3: #8300fc;
+  --brand-base: rgb(131 0 252);
+  --vp-c-brand-1: var(--brand-base);
+  --vp-c-brand-2: rgb(106 0 209);
+  --vp-c-brand-3: var(--brand-base);
 
-  --brand-soft-base: rgb(131 0 252);
-  --vp-c-brand-soft: rgb(from var(--brand-soft-base) r g b / 0.14);
+  --vp-c-brand-soft: rgb(from var(--brand-base) r g b / 0.14);
 
-  --vp-c-bg-alt: #f8f8f8;
-  --vp-c-bg-soft: #f8f8f8;
-  --vp-c-bg-mute: #efefef;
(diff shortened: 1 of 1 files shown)


site/.vitepress/theme/custom.css
```diff
@@ -13,11 +15,12 @@
...
     160deg,
-    rgba(131, 0, 252, 0.18) 0%,
-    rgba(131, 0, 252, 0.04) 55%,
-    rgba(248, 248, 248, 0.9) 100%
+    rgb(from var(--brand-soft-base) r g b / 0.18) 0%,
+    rgb(from var(--brand-soft-base) r g b / 0.04) 55%,
+    rgb(248 248 248 / 0.9) 100%
   );
   --vp-home-hero-image-filter: blur(44px);
+  --vp-code-line-highlight-color: rgb(from var(--brand-soft-base) r g b / 0.06);
```
askoufis: This is the line that actually sets the highlight color. Everything else is refactoring.

site/.vitepress/theme/custom.css
```diff
@@ -2,7 +2,9 @@
   --vp-c-brand-1: #8300fc;
   --vp-c-brand-2: #6a00d1;
   --vp-c-brand-3: #8300fc;
-  --vp-c-brand-soft: rgba(131, 0, 252, 0.14);
+
+  --brand-soft-base: rgb(131 0 252);
```
jahredhope: rgb or hex, can we pick a direction?

I'd be happy to migrate all colours in this file to rgb if that's the preference.

jahredhope (commented):
This change makes me very happy. Thank-you.

site/.vitepress/theme/custom.css
askoufis: I agree. Turns out `rgb(131 0 252) === #8300fc` there isn't a need for `--brand-soft-base`.

askoufis: https://github.com/seek-oss/sku/pull/1685/changes/8abdde3be0f8f378b4f834a1d66c478e1f58610d

jahredhope (approved):

PR #1687, opened 2026-08-11
Implement Vite SSR Change including Managed Data Mode SSR

jahredhope:
Implements Vite SSR change as defined in https://github.com/seek-oss/sku/pull/1670 and refined with https://github.com/seek-oss/sku/pull/1686

## Purpose

Add a new high level SSR API called Managed Data Mode, that wraps React Router and provides a slightly more "batteries included" experience when developing an SSR app.

We still leave many core features to the user such as logging and state-management specific solutions, whilst providing hooks to ensure doing so is possible.

As this is a major new API we are deliberately cutting down support for anti-patterns or extreme customisation so as to better ensure consumers are providing exceptional use-cases back to support for iteration.

## Changes (see full [openspec change ](https://github.com/seek-oss/sku/tree/vite-ssr-impl/openspec/changes/vite-ssr)for full details)

<details>
<summary>Config, peers, and rejects</summary>

- Add `buildType: 'ssr' | 'static'`. Reject webpack + `ssr`, suffixed `-ssr` when `buildType` is set, and absolute/`CDN` `publicPath`.
- SSR single-port: bake `__SKU_DEFAULT_SERVER_PORT__` from `port`. Reject / untype `serverPort`. Keep `PORT` override at runtime.
- Hard-error when SSR has a configured `public` directory. Disable Vite `publicDir` and `copyPublicFiles` for SSR only.
- Hard-error when SSR sets `dangerouslySetViteConfig` or `vitePlugins`. Omit both from the SSR plugin graph.
- Add optional peer `react-router` `^8` for SSR consumers.
- Add optional boolean `expressTrustProxy`. When `true`, set Express hop count `1` before listen.
- Soft-default empty/omitted config `sites` to `['default']`.

</details>

<details>
<summary>Package surface (<code>sku/runtime</code>)</summary>

- Export Managed Data Mode APIs from `sku/runtime` (not a strategy-branded subpath). Drop `Ssr` from public type names.
- Keep sku-only shared-state symbols off public `sku/runtime`; mount via private package `imports` (e.g. `#runtime/*`).
- `optimizeDeps.exclude` for `'sku'` and `'sku/runtime'` to prevent module duplication.
- Import `virtual:sku/polyfills` at the top of the SSR browser client entry only.
- Public hooks: `createSkuContexts`, `useInsertHtml`, `usePreloadRoute`, `getCspNonce`.

</details>

<details>
<summary><code>routesEntry</code> and site-scoped trees</summary>

- Config `routesEntry` (default `src/routes.tsx`) aliased into both SSR graphs.
- `SkuRouteObject = RouteObject & { sites?: string[] }`. Require named `routes` array.
- Pre-build one route tree per resolved site name. Omit `sites` ⇒ all sites. No parent→child inheritance. Strip `sites` before React Router.
- Fail closed on missing/invalid/unknown site. Do not select the tree from config hosts.
- Optional `mapRoutePath` (`{ path, site, parentSegments }` → `string[]`) after sites filter; clones preserve `lazy` / `handle` / `moduleId`; empty array omits; identity when omitted.
- Index homes: `mapRoutePath` for `index: true` with `path: ''`; non-empty maps become path clones without `index`.
- During pre-build, default omitted `caseSensitive` to `true`; preserve explicit `true` / `false`; apply to `mapRoutePath` clones.
- Prefer inline `lazy` in `routesEntry`; page modules own `loader` / `action` / `Component` (no per-page `route.ts` stubs as the happy path).

</details>

<details>
<summary>Request entries and always-on <code>SkuProvider</code></summary>

- Types: `SkuServerEntry` / `SkuClientEntry` with zero-runtime `defineServerEntry` / `defineClientEntry` and sibling `NoInfer` typing.
- Infer `Site` / `Language` / `ClientContext` / `ReactContext` from getter returns. `defineClientEntry<typeof server>` projects client `site` / `clientContext` from the server entry.
- Call order before `query()`: `getSite` → `getLanguage` → `getClientContext` → `getReactContext` → optional `getRouterContext`. Do not serialise `reactContext`.
- Require `getSite` at init only when resolved sites length is `> 1`.
- Always mount `SkuProvider` outside the router with `site` + `clientContext` + `reactContext`. No app-authored `Providers` export.
- Three value channels: `getClientContext`, `getReactContext`, `getRouterContext`.
- Wire optional dual `getRouterContext` into `query({ requestContext })` and `createBrowserRouter({ getContext })`.
- Optional `middleware` (array), `onHydrate`, and `onListen`.
- Optional React Router `instrumentations` pass-through on server (route-only into `createStaticHandler`) and client (`createBrowserRouter`).

</details>

<details>
<summary>Start and build server runtime</summary>

- `sku start` / `sku build` for SSR: Vite middlewareMode, sibling `client/` + `server/`, Document stream, document hydrate.
- Shared aliases `__sku_alias__serverEntry` / `__sku_alias__clientEntry` / `__sku_alias__routesEntry`.
- Build `createStaticHandler` per site at init. Per request select handler and call only `query()` / `createStaticRouter`.
- Abort-before-write. Forward loader/action Responses and headers. Errored-route `statusCode` + ErrorBoundary. `waitForAll`. `httpsDevServer`.
- Skip `transformIndexHtml` on SSR. Manifest assets. Auto `moduleId` for lazy routes.
- Mount `vitePluginSsrCss` on the SSR serve graph; virtual stylesheet URL in Document `assets.css` on `sku start`.
- Mount `telemetryPlugin` with `type: 'ssr'`. Deliver page-load + HMR clients via client entry / bootstrap.
- Promise-scrub loader/action data. Strip production `Error.stack`. Harden Express→Fetch adapter.
- Hydrate client from bootstrap `site` (same as SSR). Call optional `onHydrate({ clientContext })` only.
- Express 4 shared server runtime (no Express 5 bump in this change).

</details>

<details>
<summary>Assets, <code>publicPath</code>, and production layout</summary>

- Treat `publicPath` as the static asset prefix only. Bake `__SKU_PUBLIC_PATH__`. Do not use Vite `BASE_URL` as React Router basename.
- `sku start` ignores `publicPath` and serves Vite bootstrap from `/`. Production keeps `config.base` + assets under `publicPath`.
- After client build, bake the Vite client manifest into the server output so production can start without sibling `client/`.
- Mount `express.static(publicPath)` only when sibling `client/` exists, after request-context and before app-entry `middleware`.

</details>

<details>
<summary>CSP</summary>

- Shell CSP headers (enforcing and/or Report-Only). Lazy single nonce only when requested. Async Local Storage holds nonce only.
- Align production defines with webpack (`__SKU_CSP__` including report-only fields).
- Support `cspReportTo` / `Reporting-Endpoints` for URL-bearing endpoints.
- Export `getCspNonce` from `sku/runtime`.

</details>

<details>
<summary>Vocab and language</summary>

- Register language chunk only when `getLanguage` returns a language. No allowlist / sole-language default. No client forward of language.
- When vocab / `languages` is active, resolve `@vocab/vite/runtime` from sku and alias the export file (not the package root).
- Validate SSR + vocab without a consumer direct `@vocab/vite` dependency.

</details>

<details>
<summary>Streaming HTML insert + intent preload</summary>

- Render-scoped injection queue + React context shared by `render` and `sku/runtime`, outermost around `Document`.
- `useInsertHtml()` from `sku/runtime`: silent no-op off the SSR path; never throws.
- Node transform on the response pipe writes queued markup before the next React chunk; flush remainder at stream end (`onShellReady` and `waitForAll`).
- Register the selected site route tree from the client entry. `usePreloadRoute` supports both `lazy` shapes; no-op when unregistered + dev warning on client invoke.

</details>

<details>
<summary><code>onListen</code></summary>

- Optional `onListen` on `SkuServerEntry` (`{ app, httpServer, port }` → `void | Promise<void>`).
- Call once after middleware + HTML mounted and `listen` succeeds. Await promise. Failure fails startup. Do not re-call on server-entry HMR.
- Covers the webpack `onStart` window for Managed Data Mode SSR.

</details>

<details>
<summary>Create template</summary>

- `@sku-lib/create` template `ssr` with dual entry scaffold via `define*Entry` + `createSkuContexts`.
- Flatten layout: pathless root layout route, named `Component` on lazy pages, inline `lazy`.
- Omit config `sites` and `getSite` (soft-default). Set `expressTrustProxy: true`.
- Static `vite` template left unchanged.

</details>

<details>
<summary>Docs</summary>

- New `/ssr/` product docs: index, entries, routing, providers, data-loading, middleware, CSP, runtime-api, error-pages, multi-language, logging, deploy-to-production, troubleshooting.
- Migrating: migrate-from-webpack-ssr, migrate-from-static-app, webpack-ssr reference.
- Document Managed Data Mode vs SSR naming, unsupported SSR options, Express 4 + React Router 8 peer policy, experimental warning.
- Configuration / multi-site / multi-language docs updated for `routesEntry`, `mapRoutePath`, and case-sensitive defaults.

</details>

<details>
<summary>Fixtures and tests</summary>

- New/updated fixtures: `ssr-document`, `ssr-data`, `ssr-sites`, `stream-insert-html` (Apollo Client 4 + `useInsertHtml`), `logging`, translations SSR adapters, security-controls coverage as applicable.
- Browser tests: streaming Suspense, document hydrate, multi-site, CSP (+ report-only / report-to), redirects, `waitForAll`, errored-route status, loader cookies, HTTPS start, `onListen` / `expressTrustProxy`, site filtering, baked-manifest production server without sibling `client/`, preload intent, module identity under workspace link.
- Create template generate tests and snapshots updated for `ssr`.

</details>

<details>
<summary>Release</summary>

- Changeset: experimental / not-for-production Managed Data Mode SSR (`sku` + `@sku-lib/create` minor).
- End-state public names only; do not label unreleased API churn as breaking.

</details>

<details>
<summary>Explicitly out of scope / deferred</summary>

- Webpack SSR backfill; Framework Mode / RSC; Express 5; absolute/`CDN` `publicPath`; app-authored `Providers`; sku-owned Apollo.
- Future Managed Data Mode Static path; optional compose slot above the router; public language-in-React-context hook; file-based route building; SSR observability beyond instrumentations.

</details>

## Reviewers guide

Recommended review order: **Proposal → Docs → Design → Specs → fixtures → implementation**.

OpenSpec reviewing tips: [Reviewing Changes](https://openspec.dev/docs/reviewing-changes).

### 1. Contract (read first)

1. [Proposal](https://github.com/seek-oss/sku/blob/vite-ssr-impl/openspec/changes/vite-ssr/proposal.md) — why, scope, impact
2. [Design](https://github.com/seek-oss/sku/blob/vite-ssr-impl/openspec/changes/vite-ssr/design.md) — decisions and non-goals
3. Specs (living requirements):
   - [managed-data-mode](https://github.com/seek-oss/sku/blob/vite-ssr-impl/openspec/changes/vite-ssr/specs/managed-data-mode/spec.md)
   - [ssr](https://github.com/seek-oss/sku/blob/vite-ssr-impl/openspec/changes/vite-ssr/specs/ssr/spec.md)
   - [csp](https://github.com/seek-oss/sku/blob/vite-ssr-impl/openspec/changes/vite-ssr/specs/csp/spec.md)

Spec PRs for context: #1670, #1686.

### 2. Try it before deep-diving code (recommended)

Branch packages are published under the `vite-ssr-impl` npm dist-tag.

**Scaffold a new app** (preferred — uses `@sku-lib/create`, not `sku create`):

```bash
pnpm dlx @sku-lib/create@vite-ssr-impl my-ssr-app --template ssr
cd my-ssr-app
pnpm start
```

Or install create locally, then run its bin:

```bash
pnpm add -D @sku-lib/create@vite-ssr-impl
pnpm exec create-sku my-ssr-app --template ssr
```

Then set `bundler: 'vite'` + `buildType: 'ssr'` and follow:

- [Getting started](https://github.com/seek-oss/sku/blob/vite-ssr-impl/site/docs/ssr/index.md)
- [Migrate from webpack SSR](https://github.com/seek-oss/sku/blob/vite-ssr-impl/site/docs/ssr/migrate-from-webpack-ssr.md)
- [Migrate from static](https://github.com/seek-oss/sku/blob/vite-ssr-impl/site/docs/ssr/migrate-from-static-app.md)

### 3. Implementation chapters

We recommend focusing on reviewing the chapters that match your concern. Suggested order for a full pass:

| #   | Chapter           | What to read                                                                                                     |
| --- | ----------------- | ---------------------------------------------------------------------------------------------------------------- |
| 1   | Public API        | `packages/sku/src/services/vite/ssr/types.ts`, `defineEntry.ts`, `packages/sku/src/runtime.ts`, `skuContext.tsx` |
| 2   | Request path      | `streamDocument.tsx`, `render.tsx`, `ssrServerShared.ts`, `bootstrap.ts`                                         |
| 3   | Routes / sites    | `buildSiteRouteTrees.ts`, `entries/ssr-client.tsx`, `entries/ssr-server.tsx`, `plugins/ssr.ts`                   |
| 4   | Dev / prod wiring | `createDevSsrServer.ts`, `startProductionSsrServer.ts`, `validateConfig.ts`                                      |
| 5   | Seams             | `csp.ts`, `insertHtml.tsx`, `createInsertHtmlTransform.ts`, `resolveAssets.ts`, baked client manifest            |
| 6   | Proof             | `tests/browser/*ssr*`, `fixtures/ssr-*`, `fixtures/stream-insert-html`                                           |
| 7   | Adoption          | `site/docs/ssr/*`, `packages/create/templates/ssr`                                                               |

**Noise / skim:** `pnpm-lock.yaml`, create snapshots, fixture page chrome unless you are reviewing that chapter.

### 4. Split review by expertise (optional)

- **API / types / DX** — chapters 1 + 7
- **Streaming / server lifecycle** — chapters 2–4
- **CSP / assets / deploy** — chapter 5 + [deploy-to-production](https://github.com/seek-oss/sku/blob/vite-ssr-impl/site/docs/ssr/deploy-to-production.md)
- **Behaviour proof** — chapter 6


packages/sku/src/services/vite/ssr/ssrServerShared.ts
```diff
@@ -0,0 +1,360 @@
...
+        },
+        manifest,
+      );
+
+      if ('response' in result) {
+        await sendResponse(result.response, res);
+        return;
+      }
+      if (controller.signal.aborted) {
+        return;
```
SpiralLogic: when the client disconnects after render() resolves but before pipe, renderToPipeableStream keeps running. 
```suggestion
        result.abort();
        return;
```

jahredhope: Very good catch. Fixed.

SpiralLogic:
I found some more issues around request disconnects, retries cancelations etc, so I just put that on another branch and raised a PR to this one for you to review @jahredhope  https://github.com/seek-oss/sku/pull/1688

fixtures/translations/package.json
```diff
@@ -5,15 +5,17 @@
   "scripts": {
     "build": "sku build",
     "build:vite": "sku build --config sku.config.vite.ts",
+    "build:vite-ssr": "sku build --config sku.config.vite-ssr.ts",
     "serve": "sku serve",
+    "serve:vite-ssr": "sku serve --config sku.config.vite.ts",
```
askoufis: This scripts is a bit confusing: it's `vite-ssr` but it's using `sku serve` and the non-ssr config.

fixtures/logging/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@sku-fixtures/logging",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "build": "sku build --config sku.config.ts",
+    "start": "sku start --config sku.config.ts",
```
askoufis: These config flags are redundant (same issue is present in the other fixtures added in the PR). Is there a reason they're specified?

pnpm-workspace.yaml
askoufis, line 1: Probably best to make a changeset for these `react` dep updates (not `react-router` though).

private/playwright/appSnapshot.ts
```diff
@@ -1,10 +1,36 @@
...
 };
 
+const CLIENT_HTML_SETTLE_ATTEMPTS = 40;
+const CLIENT_HTML_SETTLE_INTERVAL_MS = 50;
+
+/**
+ * Wait until `page.content()` is unchanged across two reads so client effects
+ * (e.g. `useEffect`) and late style injection are reflected in snapshots.
+ */
+const waitForClientHtmlToSettle = async (page: Page) => {
```
askoufis, line 16: Was this added to address VE style loading flakiness during dev? If so, this was fixed upstream and shouldn't be necessary. If not then I'm keen to know what problem it addresses.

site/docs/ssr/csp.md
```diff
@@ -0,0 +1,62 @@
+# Content Security Policy
+
+:::danger Experimental — not for production
```
askoufis: We've used [blockquote-style alerts](https://vuejs.github.io/vitepress/v1/guide/markdown#github-flavored-alerts) in other parts of the docs, so would be good to be consistent.

site/docs/ssr/csp.md
```diff
@@ -0,0 +1,62 @@
...
+Managed Data Mode SSR is available for evaluation and testing. Do not use it in production yet; the API and behaviour may change.
+In the meantime, continue using [Webpack SSR](./webpack-ssr.md).
+:::
+
+On SSR, Content Security Policy is delivered as **HTTP headers** (`Content-Security-Policy` / `Content-Security-Policy-Report-Only`).
+Meta `http-equiv` CSP is not used on this path.
```
askoufis: `path` is a bit ambiguous here. Is it referring to "the SSR path", as in, an SSR app?

packages/sku/src/program/commands/test/vitest-test-handler.ts
```diff
@@ -20,6 +20,9 @@ export const vitestHandler = async ({
 
   const overrideableOptions: TestUserConfig = skuContext.vitestDecorator({
     css: true,
+    // Vitest intercepts console by default and can drop setupFiles logs (and other
+    // output not tied to a test). Keep raw console so setupTests and debugging work.
+    disableConsoleIntercept: true,
```
askoufis: I'd rather discuss the tradeoffs to disabling this before doing it. Not sure if it belongs in this PR.

fixtures/ssr-document/src/routes.tsx
```diff
@@ -0,0 +1,28 @@
...
+      { path: 'about', lazy: () => import('./pages/about/about.js') },
+      { path: 'details', lazy: () => import('./pages/details/details.js') },
+      {
+        path: 'buffered',
+        lazy: () => import('./pages/buffered/buffered.js'),
+        handle: {
+          waitForAll: true,
+        },
+      },
+      { path: 'boom', lazy: () => import('./pages/error/error.js') },
```
askoufis, line 22: Loading this page briefly flashes a stack trace (part of its static HTML) before it gets replaced with:

```
💿 Hey developer 👋

You can provide a way better UX than this when your app throws errors by providing your own ErrorBoundary or errorElement prop on your route.
```

Is this coming from `react-router`? Would be nice if we could make it a nicer experience.

fixtures/ssr-document/src/pages/buffered/buffered.tsx
```diff
@@ -0,0 +1,24 @@
+import { Suspense, use, cache } from 'react';
+
+const getDeferredMessage = cache(
```
askoufis: [The `cache` docs](https://react.dev/reference/react/cache) state that it can only be used with RSCs, but we're using it here without using RSCs. I think the functionality is still valuable, so if we _can_ let users use it that's great, but should we be more explicit (in the docs) about our current lack of RSC support and/or the fact that you can still use `cache` if you want?

fixtures/ssr-data/src/RootLayout.tsx
```diff
@@ -0,0 +1,20 @@
+import { Link, Outlet } from 'react-router';
+
+export const RootLayout = () => (
+  <>
+    <link
+      rel="icon"
+      href="data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg'/>"
+    />
```
askoufis, line 8: Saw this `link` in another fixture as well. Is it used for anything?

packages/create/src/generators/templates.ts
```diff
@@ -16,6 +17,31 @@ import type { Template } from '../types/index.js';
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
+/** Base files incompatible with SSR (static render / #app hydrate / App shell). */
+const BASE_FILES_SKIPPED_FOR_SSR = new Set([
```
askoufis, line 21: This feels like a bit of a smell. If the SSR template has to ignore so many files from the base template, maybe a way to disable the `base` template inheritance would be a simpler way to do this.

packages/create/README.md
```diff
@@ -0,0 +1,23 @@
...
+pnpm dlx @sku-lib/create my-app
+```
+
+## Templates
+
+| Template | Flag                 | Notes                                                                                                                         |
+| -------- | -------------------- | ----------------------------------------------------------------------------------------------------------------------------- |
+| Vite     | `--template vite`    | Static / SSG app (default interactive choice). Vitest.                                                                        |
+| SSR      | `--template ssr`     | Server-rendered Managed Data Mode app (`buildType: 'ssr'`). Experimental — not for production. Use `sku start` / `sku build`. |
+| Webpack  | `--template webpack` | Webpack + Jest.                                                                                                               |
```
askoufis:
```suggestion
| Vite     | `--template vite`    | Static / SSG app (default interactive choice). Vite + Vitest.                                                                        |
| SSR      | `--template ssr`     | Server-rendered Managed Data Mode app (`buildType: 'ssr'`). Experimental — not for production. Use `sku start` / `sku build`. |
| Webpack  | `--template webpack` | Legacy static / SSG app. Webpack + Jest.                                                                                                               |
```

packages/sku/src/context/createSkuContext.ts
```diff
@@ -295,6 +298,8 @@ export const createSkuContext = async ({
     defaultClientEntry,
     serveCjsInteropDependencies,
     buildCjsInteropDependencies,
+    /** Explicit `compilePackages` from sku config (not defaults / auto-detected). */
+    configuredCompilePackages: skuConfig.compilePackages ?? [],
```
askoufis: We've had issues in the past around the number and usage of compile package-related APIs. They've caused confusion and bugs, so I'd be keen to figure out a way to either incorporate this alongside the existing APIs, or move things around enough to obviate the need for it entirely.

tests/node/lint-format.test.ts
```diff
@@ -254,9 +264,10 @@ describe('lint-format', () => {
       let lint: RenderResult;
 
       beforeAll(async () => {
-        await using fixture = await createSrcFixture({
-          // only fails vitest rules (correctly formatted)
-          'vitestRules.test.ts': dedent /* ts */ `
+        lint = await withSrcFixture(
```
askoufis, line 267: Is the `lint` render result even needed anymore now that the result is asserted in the scope of `withSrcFixture`?

tests/node/sku-create.test.ts
```diff
@@ -181,14 +222,77 @@ describe.each(['webpack', 'vite'])('sku-create %s', (template) => {
...
+        'utf-8',
+      );
+      const skuConfig = await fs.readFile(
+        fixturePath(projectName, 'sku.config.ts'),
+        'utf-8',
+      );
+
+      expect(skuConfig).toContain("buildType: 'ssr'");
+      expect(skuConfig).toContain('expressTrustProxy: true');
+      expect(routes).toContain('export const routes');
```
askoufis: AI absolutely _hates_ snapshot tests for some reason. I feel like we should just snapshot these files instead of having a billion `expect().toContain()` calls.

vitest.config.ts
```diff
@@ -1,12 +1,38 @@
...
+const repoRoot = path.dirname(fileURLToPath(import.meta.url));
 
 export default defineConfig({
   resolve: {
     tsconfigPaths: true,
+    // Unit tests import helpers relatively from source; sku mounts shared
+    // state via `#runtime/*` and apps via `sku/runtime`. Without these aliases
+    // Vitest resolves package exports/imports to dist and splits React context
+    // / ALS identity.
+    alias: {
```
askoufis: We've neglected to invest time into making testing a nicer experience by not having to build packages between test runs. Configuring `alias`es is part of the story, but I'd rather address it for all entrypoints in a separate PR, if that's a problem we want to solve.

tests/browser/logging.test.ts
```diff
@@ -0,0 +1,397 @@
...
+      skipCleanup(task.id);
+      await clearTelemetry();
+
+      const response = await fetch(baseUrl);
+      expect(response.status).toBe(200);
+      const html = await response.text();
+      expect(html).toContain('data-testid="home-page"');
+      expect(html).toContain('data-testid="home-work"');
+
+      const events = await waitForLogs((logs) =>
```
askoufis, line 120: Again, would it be simpler (both in terms of implementation and readability) to snapshot (and sanitize) the handful of logs instead of doing all these precise assertions? Maybe not for _all_ test in this file, but some of them?

packages/sku/src/services/vite/ssr/getModuleIds.ts
```diff
@@ -0,0 +1,33 @@
...
+  }: {
+    development: boolean;
+    requestLanguage?: string;
+  },
+): string[] => {
+  const moduleIds = matches.flatMap(({ route }) => {
+    const moduleId = (route.handle as SkuRouteHandle | undefined)?.moduleId;
+    if (development && route.lazy && !moduleId) {
+      console.warn(
+        `[sku] Lazy route at "${String(route.path ?? '(index)')}" is missing handle.moduleId. Prefer idiomatic lazy: () => import('./pages/about/about') so sku can auto-derive it, or set handle.moduleId explicitly to the Vite client manifest key (e.g. "src/pages/about/about.tsx") for production modulepreload links.`,
```
askoufis: This feels like something that would be better explained in the docs rather than (or potentially in addition to) a console warning. Thoughts on documenting this somewhere (or adding a link to those docs if they already exist)?

tests/browser/ssr-sites.test.ts
```diff
@@ -0,0 +1,184 @@
...
+        },
+      });
+
+      await waitFor(
+        async () => {
+          const response = await fetch('http://127.0.0.1:8217/');
+          expect(response.ok).toBe(true);
+          const html = await response.text();
+          expect(html).toContain('__SKU_SITE__');
+          expect(html).toContain('"au"');
```
askoufis: The full `'__SKU_SITE__="au"'` text is asserted in a single assertion above. Can this not be done here too?

packages/sku/src/runtime.test.ts
askoufis, line 1: Is this test file valuable? It's not really a pattern we use anywhere else.

packages/sku/src/services/vite/ssr/selectForSite.ts
```diff
@@ -0,0 +1,33 @@
...
+ * static handler. Missing or non-string `site`, or a missing map entry, fails closed.
+ */
+export const selectForSite = <T>(
+  siteMap: Record<string, T>,
+  site: unknown,
+  source: SiteSource,
+): T => {
+  assertSiteName(site, source);
+
+  if (!Object.prototype.hasOwnProperty.call(siteMap, site)) {
```
askoufis: What year is it, 2015? Jokes aside, is this actually necessary?

packages/sku/src/services/vite/plugins/lazyRouteModuleId/injectLazyRouteModuleId.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  it('injects handle.moduleId for idiomatic lazy: () => import()', () => {
+    const result = transform(`
+      const routes = [{
+        path: 'about',
+        lazy: () => import('./pages/about/about'),
+      }];
+    `);
+
+    expect(result?.injected).toBe(true);
```
askoufis, line 53: AST transforms are a place where I think a snapshot is _very_ valuable. My small human brain can't piece together all these assertions to figure out what the transform actually did.

packages/sku/src/services/vite/plugins/lazyRouteModuleId/injectLazyRouteModuleId.test.ts
```diff
@@ -0,0 +1,162 @@
...
+import { join } from 'node:path';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { injectLazyRouteModuleId } from './injectLazyRouteModuleId.js';
+
+describe('injectLazyRouteModuleId', () => {
+  let projectRoot: string;
+
+  beforeEach(() => {
+    projectRoot = mkdtempSync(join(tmpdir(), 'sku-lazy-module-id-'));
+    mkdirSync(join(projectRoot, 'src/pages/about'), { recursive: true });
```
askoufis: This should use `fs-fixture` instead (we already use it elsewhere in the repo).

packages/sku/src/types/types.ts
```diff
@@ -385,6 +398,31 @@ export interface SkuConfigBase {
...
+   * the server and client graphs. Optional `sites` on routes declares
+   * multi-site membership; apps select the tree via `getSite` (required when
+   * config has more than one site; sole config site when omitted on single-site).
+   * Optional named `mapRoutePath` maps one logical path (including index
+   * homes via `path: ''`) to concrete paths while sku pre-builds each site tree.
+   *
+   * @default "./src/routes.tsx"
+   * @link https://seek-oss.github.io/sku/configuration#routesentry
+   */
+  routesEntry?: string;
```
askoufis: Should this be defined inside `ViteSkuConfig` instead of `SkuConfigBase`?

packages/sku/src/services/vite/ssr/Document.tsx
```diff
@@ -0,0 +1,31 @@
...
+          href={href}
+          {...(href === SSR_CSS_VIRTUAL_HREF ? { 'data-ssr-css': true } : {})}
+        />
+      ))}
+    </head>
+    <body>{children}</body>
+  </html>
+);
+
+export default Document;
```
askoufis: Does `Document` need a default and a named export?

packages/sku/src/services/vite/ssr/entryTypeExtractors.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * Entry objects from `define*Entry` keep getters optional on the type, so
```
askoufis, line 2: This docstring is attached to the `SiteOf` type, but really it's describing all the types in this file.

packages/sku/src/services/vite/ssr/startProductionSsrServer.ts
```diff
@@ -0,0 +1,63 @@
...
+    serverDirectory,
+    CLIENT_MANIFEST_RELATIVE_PATH,
+  );
+  const manifest = JSON.parse(
+    await readFile(bakedManifestPath, 'utf8'),
+  ) as ClientManifest;
+  const entry = findEntryChunk(manifest);
+  const publicPath = __SKU_PUBLIC_PATH__;
+  const assets: RenderAssets = {
+    bootstrapModules: [path.posix.join(publicPath, entry.file)],
```
askoufis: Why is `posix.join` used here instead of the OS's `join`?

packages/sku/src/services/vite/ssr/resolveAssets.ts
```diff
@@ -0,0 +1,113 @@
+import { existsSync } from 'node:fs';
+import { resolve } from 'node:path';
+
+export interface ManifestChunk {
```
askoufis: `vite` exports a very similar `ManifestChunk` type. Is there a reason not to use it?

packages/sku/src/services/vite/ssr/resolveAssets.ts
```diff
@@ -0,0 +1,113 @@
...
+  imports?: string[];
+  dynamicImports?: string[];
+  isEntry?: boolean;
+  src?: string;
+}
+
+export type ClientManifest = Record<string, ManifestChunk>;
+
+const joinPublicPath = (publicPath: string, file: string) =>
+  `${publicPath.endsWith('/') ? publicPath : `${publicPath}/`}${file}`;
```
askoufis: Can this use `path.join` instead of string concatenation?

packages/sku/src/services/vite/ssr/resolveAssets.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      expect.stringContaining(
+        'Unknown route handle.moduleId "src/pages/missing/missing.tsx"',
+      ),
+    );
+  });
+
+  it('does not warn for path-like moduleIds that exist on disk', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const cwd = mkdtempSync(join(tmpdir(), 'sku-module-id-'));
+    mkdirSync(join(cwd, 'src', 'pages', 'about'), { recursive: true });
```
askoufis: More opportunities to use `fs-fixture` here.

packages/sku/src/services/vite/ssr/bootstrap.ts
```diff
@@ -0,0 +1,109 @@
+import { isRouteErrorResponse, type StaticHandlerContext } from 'react-router';
+import type {
+  DocumentAssets,
+  JsonValue,
+  SerializableHydrationState,
+} from './types.js';
+
+const escapeScriptValue = (value: unknown) =>
```
askoufis, line 9: `serialize-javascript` does this escaping for you (and other things). Is there a reason so not use it here (we use it elsewhere already)?

packages/sku/src/services/vite/ssr/createDevSsrServer.ts
```diff
@@ -0,0 +1,153 @@
...
+      middlewareMode: true,
+      hmr: { server: httpServer },
+    },
+  });
+
+  const serverModule = (await vite.ssrLoadModule(serverEntry)) as {
+    middleware?: SkuMiddleware;
+    onListen?: SkuOnListen;
+    render: RenderFunction;
+  };
```
askoufis: Is there a defined type somewhere that can be used here instead of a type literal?

packages/sku/src/services/vite/ssr/requireNamedExport.ts
```diff
@@ -0,0 +1,124 @@
+/**
+ * Assert a SSR entry module exposes a required named export.
+ * Missing values are a hard error (no soft-skip / sku noop).
+ */
+export const requireNamedExport = <T>(
```
askoufis: Are we re-inventing runtime validation here? What's the advantage of rolling our own validation instead of using `zod` or similar?

packages/sku/src/runtime.ts
```diff
@@ -0,0 +1,28 @@
...
+  SkuGetLanguage,
+  SkuGetSite,
+  SkuMiddleware,
+  SkuOnHydrate,
+  SkuOnListen,
+  SkuRouteObject,
+  SkuServerEntry,
+  SkuServerGetReactContext,
+  SkuServerGetRouterContext,
+} from './services/vite/ssr/types.js';
```
askoufis, line 20: Are _all_ of these types actually likely to be imported by a consumer? I'd rather be conservative with what's exported here if possible.

packages/sku/src/services/vocab/runVocab.ts
```diff
@@ -1,12 +1,42 @@
...
+import { createRequire } from 'node:module';
+import { dirname, join } from 'node:path';
 import { getVocabConfig } from './config.js';
 import type { SkuContext } from '../../context/createSkuContext.js';
 
+/**
+ * Vocab's `compile` always ignores node_modules, so shared packages with
+ * `.vocab` (via `compilePackages`) need a separate compile with their package root.
+ */
+const compileConfiguredCompilePackages = async (
```
askoufis: Translated packages should be published with compiled translations based off their own config. This feels like it's papering over a publishing issue rather than actually solving a problem.

packages/sku/src/services/vite/ssr/ssrServerShared.test.ts
```diff
@@ -0,0 +1,473 @@
...
+    } as unknown as Request;
+
+    const request = createWebRequest(req, new AbortController().signal);
+    expect(await request.text()).toBe('raw-body');
+  });
+});
+
+describe('createHtmlRenderMiddleware abort-before-write', () => {
+  it('does not write HTML when the client disconnects before headers', async () => {
+    let resolveRender!: (result: RenderResult) => void;
```
askoufis, line 211: Surprised this doesn't get picked up by `no-non-null-assertions`. Seems necessary in this case.

askoufis (commented):
I've mostly avoid reviewing the docs site changes as we already went through half of them together and we'll likely make more changes in the future. Have reviewed everything else though.

fixtures/ssr-data/src/RootLayout.tsx
jahredhope: It avoids showing a 404 noise in the the fixture. It's not relevant for tests, but it's noise for a human reviewing the tests.

I'm keen to keep it in, and extend it to other fixtures.

packages/sku/src/services/vite/ssr/selectForSite.ts
jahredhope: I think so, yes. The older method call can be cleaned up but we don't want `toString` to be a valid site.

packages/sku/src/services/vocab/runVocab.ts
jahredhope: Removed

packages/sku/src/runtime.ts
jahredhope: By the time we release GA I think they will be. As consumers love building abstractions. But happy to keep it short for now. Cut down to things required in fixtures.

packages/sku/src/services/vite/ssr/requireNamedExport.ts
jahredhope: Cut this back down to some basic if statements.

site/docs/ssr/csp.md
jahredhope: Sure. I used ::: initially, and didn't realise the preference had changed.

site/docs/ssr/csp.md
jahredhope: reworded

fixtures/translations/package.json
jahredhope: fixed

pnpm-workspace.yaml
jahredhope: added

private/playwright/appSnapshot.ts
jahredhope: It waits for hydration / `useEffect` so `clientRenderContent` snapshots include post-hydrate DOM.

packages/sku/src/program/commands/test/vitest-test-handler.ts
jahredhope: Reverted. Happy to discuss that tradeoff outside this PR.

fixtures/ssr-document/src/routes.tsx
jahredhope: Yes, React Router’s default error UI. Added an `ErrorBoundary` on the `ssr-document` root route and the create SSR template, and documented the flash on the error-pages page. sku still does not ship a default boundary.

fixtures/ssr-document/src/pages/buffered/buffered.tsx
jahredhope: Dropped `cache()` from the fixture — the buffered page now uses `use()` with a promise. Data-loading docs still state that sku does not support RSC, and that `cache()` can still memoize per request during document SSR.

packages/create/src/generators/templates.ts
jahredhope: Left the skip list for this PR. A non-inheriting SSR template is a larger generator change; I’d rather do that as follow-up.

packages/create/README.md
jahredhope: Applied.

packages/sku/src/context/createSkuContext.ts
jahredhope: Removed. It only existed to drive the extra vocab compile.

tests/node/lint-format.test.ts
jahredhope: Yes. `withSrcFixture` asserts the exit code; the later `it` still reads ESLint output from `lint`.

tests/node/sku-create.test.ts
jahredhope: SSR entry files are snapshotted now. The remaining `runIf` only checks that static-app files are omitted.

vitest.config.ts
jahredhope: Removed, along with `runtime.test.ts`. Broader source aliases can be a separate PR.

tests/browser/logging.test.ts
jahredhope: Snapshotted event/span names on the noisiest start-path test. Left the parent/child span asserts.

packages/sku/src/services/vite/ssr/getModuleIds.ts
jahredhope: Added an “Automatic modulepreload” section in routing docs. The console warning now links there.

tests/browser/ssr-sites.test.ts
jahredhope: Collapsed to `expect(html).toContain('__SKU_SITE__="au"')`.

packages/sku/src/runtime.test.ts
jahredhope: Deleted. Identity is covered by browser tests and the `optimizeDeps.exclude` assert.

packages/sku/src/services/vite/plugins/lazyRouteModuleId/injectLazyRouteModuleId.test.ts
jahredhope: `injectLazyRouteModuleId` now snapshots the transformed code.

packages/sku/src/services/vite/plugins/lazyRouteModuleId/injectLazyRouteModuleId.test.ts
jahredhope: Done for `injectLazyRouteModuleId` tests.

packages/sku/src/types/types.ts
jahredhope: Moved it. Webpack SSR uses `serverEntry`, not `routesEntry`.

packages/sku/src/services/vite/ssr/Document.tsx
jahredhope: Dropped the named export. Callers import the default.

packages/sku/src/services/vite/ssr/entryTypeExtractors.ts
jahredhope: Moved to a file-level comment.

packages/sku/src/services/vite/ssr/startProductionSsrServer.ts
jahredhope: It is a URL prefix, not a filesystem path. Shared `joinPublicPath` with `resolveAssets` and commented that it is always posix.

packages/sku/src/services/vite/ssr/resolveAssets.ts
jahredhope: Now using Vite’s `Manifest` / `ManifestChunk`.

packages/sku/src/services/vite/ssr/resolveAssets.ts
jahredhope: `joinPublicPath` now uses `path.posix.join`.

packages/sku/src/services/vite/ssr/resolveAssets.test.ts
jahredhope: Done.

packages/sku/src/services/vite/ssr/bootstrap.ts
jahredhope: Done, including the `undefined` vs `null` clientContext distinction. Also stopped marking it `ssr.external` so it is bundled into production `server.js` (fixtures do not depend on it).

packages/sku/src/services/vite/ssr/createDevSsrServer.ts
jahredhope: Not really. The type is named exports on another module which I don't think we have a way to pull the type from.

Added `SsrServerModule` next to `RenderFunction` and used it in the dev server. That is sku’s internal `ssr-server` entry, not `SkuServerEntry`.

packages/sku/src/services/vite/ssr/ssrServerShared.test.ts
jahredhope: Used a definite assignment (`let resolveRender!`) so the test can disconnect before `render()` resolves. `no-non-null-assertions` does not flag that form.

askoufis (approved):

PR #1688, opened 2026-08-13
Harden Vite SSR stream failure handling

SpiralLogic:
## Summary
- Make React SSR render attempts single-owner so retries, aborts, and callbacks cannot settle the same attempt more than once.
- Stop rendering or forwarding responses after request disconnects, including disconnects during streaming.
- Propagate insert-HTML transform failures through the stream pipeline and abort React.
- Add focused coverage for aborted signals, disconnect windows, render rejections, and insertion failures.

## Why
- Prevent hung render promises, retries after cancellation, and rendering work continuing after the client disconnects.
- Preserve genuine render failures for Express while suppressing expected cancellation errors.

### DIFF
--- a/packages/sku/src/services/vite/ssr/streamDocument.tsx
+++ b/packages/sku/src/services/vite/ssr/streamDocument.tsx
@@ -201,10 +201,12 @@ export const streamDocument = ({
             return;
           }
           options.onError?.(error);
-          // Suspense rejections never settle for onAllReady; retry once we
-          // see the error while still buffering for waitForAll.
-          if (waitForAll && state === 'rendering') {
-            retryFromError(error);
+          // Suspense rejections never settle for onAllReady. Recover once
+          // while still buffering; fail closed if this pass cannot produce
+          // a document (recovery has allowErrorRetry: false).
+          if (waitForAll && state === 'rendering' && !retryFromError(error)) {
+            rejectAttempt(error);
+            abortStream();
           }
         },
       },
--- a/packages/sku/src/services/vite/ssr/streamDocument.tsx
+++ b/packages/sku/src/services/vite/ssr/streamDocument.tsx
@@ -35,8 +35,16 @@ export type StreamDocumentArgs = {
   allowErrorRetry: boolean;
 };
 
+const RENDER_ATTEMPT_STATE = {
+  RENDERING: 'rendering',
+  RETRYING: 'retrying',
+  READY: 'ready',
+  ABORTED: 'aborted',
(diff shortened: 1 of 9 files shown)


packages/sku/src/services/vite/ssr/streamDocument.tsx
```diff
@@ -35,6 +35,17 @@ export type StreamDocumentArgs = {
   allowErrorRetry: boolean;
 };
 
+const RENDER_ATTEMPT_STATE = {
+  RENDERING: 'rendering',
+  RETRYING: 'retrying',
+  READY: 'ready',
+  ABORTED: 'aborted',
+  FAILED: 'failed',
+} as const;
```
jahredhope: Is there any value in this being an enum like object over a union? We'd usually just use a typed union of values.

packages/sku/src/services/vite/ssr/streamDocument.tsx
```diff
@@ -65,34 +76,67 @@ export const streamDocument = ({
...
+        return;
+      }
+      state = RENDER_ATTEMPT_STATE.FAILED;
+      removeAbortListener();
+      reject(error);
+    };
 
     const retryFromError = (error: unknown) => {
-      if (!allowErrorRetry || abortedForRetry) {
+      if (!allowErrorRetry || state !== RENDER_ATTEMPT_STATE.RENDERING) {
```
jahredhope: Is `allowErrorRetry` still required?

It seems like the goal of this change is to have a more complex state rather than just a boolean of "are we already retrying", so unless I'm mistaken we'd probably want to lean into that.

packages/sku/src/services/vite/ssr/streamDocument.tsx
SpiralLogic: not really, I went back and forth between the 2, I'll change it to a union

packages/sku/src/services/vite/ssr/streamDocument.tsx
SpiralLogic: atm the state of each streamDocument call is contained within it. the allowErrorRetry prevents a throwing boundary condition from retrying forever. a way for a child call to know that it should not retry again.

the reason for the 'retrying' state tells the rest of that call that a retry has been initiated and to ignore all further callbacks. as the child call now owns the resolve/reject and allows the parents to know that handover has happened.

they are separate concerns, they could be wrapped into the render attempt state, but it gets a bit ugly.

PR #1694, opened 2026-08-19
Allow `--convert-loadable` on mjs and catch vanilla-extract in uncompiled packages

williamlark:
This PR fixes a few issues raised by a team migrating from the webpack bundler. Overall this should ease the migration processes when moving over.

* Loosen the Vanilla Extract dep-scan filter so `./styles.css` imports (typically from uncompiled compilePackage entries) that resolve to `.css.ts` are externalized and handed to the VE plugin, instead of being prebundled.
* Run loadable conversion during dep optimization as well as in the main transform, including aliased loadable imports and mixed loadable/loadableReady imports. This also now catches mjs and cjs files that need webpack to vite loadable conversions.
* Include polyfills in optimizeDeps.entries so they are scanned on cold start. This helps reduce apps reloading a bunch of times on start when they use pollyfills, which is slower than any performance hit incurred from doing scanning up front.


knip.jsonc
```diff
@@ -178,5 +178,12 @@
     "fixtures/vite-render-error": {
       "entry": ["src/client.tsx", "src/render.tsx", "sku.config.*"],
     },
+    "fixtures/vite-compile-package-styles": {
+      "entry": ["src/client.tsx", "src/render.tsx", "sku.config.*"],
+      "ignoreDependencies": [
+        "@sku-fixtures/package-with-vanilla-extract-source",
+        "@vanilla-extract/css",
```
askoufis: Strange that it doesn't think `@sku-fixtures/package-with-vanilla-extract-source` is used 🤷. You can get away with removing `@vanilla-extract/css` from the fixture's dev deps if you want: the app itself doesn't use VE so I'd say it's fine to remove.

.changeset/vite-vanilla-extract-source-packages.md
```diff
@@ -0,0 +1,7 @@
+---
+'sku': patch
+---
+
+`build|start (vite)`: Vanilla extract styles from source-shipping packages are now picked up automatically.
```
askoufis: Just a slight rewording.

```suggestion
`build|start (vite)`: Handle Vanilla Extract styles from packages that ship source code
```

williamlark: Much better!

askoufis (approved):

PR #1695, opened 2026-08-25
Harden SSR stream abort, ErrorBoundary recovery, and disconnect teardown

SpiralLogic:
## Summary
- Abort/disconnect owns the render attempt: no ErrorBoundary recovery, no Express `next(error)`, and no `onError` for the abort itself.
- Fail closed when recovery cannot produce a document, including `waitForAll` Suspense rejections.
- Skip loaders/actions when the signal is already aborted. Abort React if the client drops during HTML headers or while buffering a short-circuit Response.
- Insert/pipeline failures abort React with the error and error the destination.
- Spec and tasks in `openspec/changes/vite-ssr` cover those windows.

## Why
- A cancelled `waitForAll` render could still start ErrorBoundary recovery after the client was gone.
- Disconnect during header writes or a Response body read could still write to the client.
- Insert transform errors could hang the React stream without reaching `onError`.


SpiralLogic:
after rebasing vite-ssr-abort the latest branch, all that was left were some tests, which we sitll may want to keep

PR #1700, opened 2026-08-26
`csp`: Update `sku serve` command to include CSP headers in response

kjots:
This change follows on from #1630 and #1631 and updates the `sku serve` command to include any CSP headers as written to the JSON metadata file alongside the rendered HTML content in the response for that content. With this change the `sku serve` command matches the expected behaviour of a compliant static serving solution in a deployed production-like environment.

### DIFF
--- a/packages/sku/src/program/commands/serve/serve.action.ts
+++ b/packages/sku/src/program/commands/serve/serve.action.ts
@@ -1,3 +1,4 @@
+import { glob, readFile } from 'node:fs/promises';
 import { join } from 'node:path';
 import exists from '../../../utils/exists.js';
 import express from 'express';
@@ -25,6 +26,13 @@ import {
 import type { SkuContext } from '../../../context/createSkuContext.js';
 import { serverUrls } from '@sku-private/utils';
 import { accent, critical, strong } from '@sku-private/utils/console';
+import type { Metadata } from '../../../types/types.js';
+
+const metadataHeaderMap: Record<keyof Metadata, string> = {
+  csp: 'content-security-policy',
+  cspReportOnly: 'content-security-policy-report-only',
+  reportingEndpoints: 'reporting-endpoints',
+};
 
 export const serveAction = async ({
   site: preferredSite,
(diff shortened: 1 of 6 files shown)


packages/sku/src/program/commands/serve/serve.action.ts
```diff
@@ -25,6 +26,13 @@ import {
 import type { SkuContext } from '../../../context/createSkuContext.js';
 import { serverUrls } from '@sku-private/utils';
 import { accent, critical, strong } from '@sku-private/utils/console';
+import type { Metadata } from '../../../types/types.ts';
```
SpiralLogic: the other imports are using .js 


```suggestion
import type { Metadata } from '../../../types/types.js';
```

packages/sku/src/program/commands/serve/serve.action.ts
```diff
@@ -102,6 +110,34 @@ export const serveAction = async ({
...
+  const metadataFilesPattern = join(environment, '**/index.html.json');
+  const metadataFiles = glob(metadataFilesPattern, { cwd: paths.target });
+
+  for await (const metadataFile of metadataFiles) {
+    const { metadata } = JSON.parse(
+      await readFile(join(paths.target, metadataFile), 'utf-8'),
+    ) as { metadata: Metadata };
+
+    headers.push({
+      source: metadataFile.slice(0, -5),
```
SpiralLogic: minor, 
```suggestion
      source: metadataFile.replace(/\.json$/, '')
```

to avoid a magic number and make it more explicit that the extension is being removed. pulling in something from 'node:path' could better, but it's likely a bit excessive in extra code..

just some suggestions, I'll leave it to you

SpiralLogic (approved):

packages/sku/src/program/commands/serve/serve.action.ts
askoufis: Trimming the `.json` extension could be done with `basename` instead.

```suggestion
      source: path.basename(metadataFile, ".json"),
```

Though I guess if we were to change the format of the metadata file this would break, I don't know if relying on a fixed extension length is any better/worse.

kjots: (Merging with @SpiralLogic's review comment above).

`basename` will strip the leading path as well, so that would not be suitable here.

I'm reluctant to use a regex because it seems overkill to use pattern matching to remove a known extension which we've explicitly matched via `glob` above. 

We could use something like `'.json'.length`, but that just resolves statically to `5` and again feels like overkill. 

In the end we know that the filename will end with `.json` and we know the length of `.json` is `5`, so I think we should keep the hard-coded number, but I'll add a comment to explain what's happening for clarity.

PR #1702, opened 2026-08-27
Allow `getClientContext` and `getReactContext` to return a Promise

jahredhope:
## Purpose

Stacked on #1687 (`vite-ssr-impl`). Completes OpenSpec `vite-ssr` section 21.

`getClientContext` and dual-entry `getReactContext` may return a Promise when that bag needs I/O. sku awaits each so later getters, `SkuProvider` hooks, and the hydrate bootstrap see the resolved value. Page data still belongs in loaders / Suspense.

## Changes

- Type `getClientContext` and dual-entry `getReactContext` as `T | Promise<T>`.
- Unwrap with `Awaited` in extractors, `define*Entry` sibling args, and `createSkuContexts` hooks.
- Keep `getSite`, `getLanguage`, and client `getRouterContext` synchronous.
- Server: await `getClientContext` then `getReactContext` before `getRouterContext` / `query()`.
- Client: await `getReactContext` before `createBrowserRouter` / `hydrateRoot`.
- Extract hydrate into `hydrateClient.tsx` so the await-before-hydrate path is unit-tested.
- Rejection of either Promise fails the document render or hydrate.
- Docs: getters MAY be async; page data stays in loaders / Suspense.


packages/sku/src/services/vite/ssr/defineEntry.test.ts
```diff
@@ -130,3 +130,102 @@ describe('defineClientEntry from ServerEntry', () => {
...
+describe('async getClientContext / getReactContext inference', () => {
+  it('unwraps Promise returns for hooks and sibling args', () => {
+    const server = defineServerEntry({
+      getSite() {
+        return 'au' as const;
+      },
+      async getClientContext() {
+        return { userId: 'fixture-user' as const };
+      },
+      async getReactContext({ site, clientContext }) {
```
williamlark, line 143: Does this need to be called async?

packages/sku/src/services/vite/ssr/render.test.tsx
```diff
@@ -138,6 +138,175 @@ describe('render', () => {
...
+        order.push('reactContext');
+        return new Promise<{ api: string }>((resolve) => {
+          release = resolve;
+        });
+      },
+      getRouterContext: () => {
+        order.push('routerContext');
+        return new RouterContextProvider();
+      },
+    });
```
williamlark, line 218: Nit: this shares a bunch of code with the above test. Could be nice to join combine the shared stuff

site/docs/ssr/data-loading.md
```diff
@@ -19,6 +19,10 @@ That keeps shared UI portable without per-app loader wiring.
 2. Mount isomorphic providers in your [root layout](./providers.md#root-layout-for-providers) and read values with `useReactContext()` / `useClientContext()`.
 3. Fetch in the React tree with Suspense (for example `useQuery`) so the same components work on SSR and client navigations.
 
+[`getClientContext`](./entries.md#getclientcontext) and [`getReactContext`](./entries.md#getreactcontext) may be async when the serialisable seed or env-differing values need I/O.
```
williamlark, line 22: nit: this part sounds a little artificial 😅 `when the serialisable seed or env-differing values need I/O.`

williamlark (approved):

site/docs/ssr/data-loading.md
jahredhope: Discussed, I'm going to address "seed" throughout the docs, it's a common problem but will do as a separate PR

PR #1703, opened 2026-08-28
Enhance middleware handling and client context normalization

jahredhope:
## Purpose

Responding to test app feedback.

Several smaller improvements around middleware context were found when trying out the use of context, including lots of spam calls to middleware when attempting to use it to load request context into the app.

## Changes

- Load vite assets BEFORE devMiddleware and production middleware
- Clarify the use of client context, react context and router context in docs
- Allow undefined in client context, and normalise this between Server and Client


askoufis (approved):

packages/sku/src/services/vite/ssr/render.test.tsx
```diff
@@ -106,6 +106,89 @@ describe('render', () => {
...
+        userId: undefined,
+        tags: [undefined, 'a'],
+      }),
+      getReactContext: ({ clientContext }) => {
+        siblingContext = clientContext;
+        return { api: 'server-api' };
+      },
+    });
+
+    if ('response' in result) {
```
williamlark: A lot of this code is already in the renderToHtml helper up top. Could that be used here instead (or parts of)?

PR #1705, opened 2026-08-31
Fix security-controls SSR fixture so client assets load from http-server

askoufis:
The adjacent Vite loadable work surfaced that the `security-controls` fixture was producing some page errors we weren't seeing. This explains why the app snapshot shows no post-hydration diff: the client was never loading.

This can be fixed by serving static assets the same way other SSR fixtures do: point publicPath at a hardcoded asset-server port.

### DIFF
--- a/fixtures/security-controls/sku-server.config.ts
+++ b/fixtures/security-controls/sku-server.config.ts
@@ -5,6 +5,8 @@ import {
 
 export default {
   port: 9847,
+  // Required for the test to serve client assets from http-server.
+  publicPath: 'http://localhost:9848',
   clientEntry: './src/serverClient.jsx',
   renderEntry: './src/render.jsx',
   serverEntry: './src/server.jsx',
--- a/fixtures/security-controls/src/serverClient.jsx
+++ b/fixtures/security-controls/src/serverClient.jsx
@@ -1,7 +1,10 @@
-import { hydrateRoot } from 'react-dom';
+import { hydrateRoot } from 'react-dom/client';
 import { loadableReady } from 'sku/@loadable/component';
 
 import App from './App';
+
+// Set to true to demonstrate attempting load unsafe script execution
+const ATTEMPT_LOAD_UNSAFE_SCRIPT = false;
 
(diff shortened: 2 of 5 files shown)


jahredhope (approved):

fixtures/security-controls/src/serverClient.jsx
```diff
@@ -1,8 +1,11 @@
-import { hydrateRoot } from 'react-dom';
+import { hydrateRoot } from 'react-dom/client';
 import { loadableReady } from 'sku/@loadable/component';
 
 import App from './App';
 
+// Set to true to demonstrate attempting load unsafe script execution
+const ATTEMPT_LOAD_UNSAFE_SCRIPT = false;
+
```
askoufis, line 8: Loading the unsafe script should've triggered CSP, but because the fixture itself never worked properly, this was never tested.

I added this as a manual way to test the CSP functionality. We could unit test this, but I don't know if that's valuable.

tests/browser/__snapshots__/security-controls.test.ts.snap
```diff
@@ -4,91 +4,105 @@ exports[`security-controls > build-ssr > should start a server with content-secu
...
+       href="http://localhost:9848/main.js"
+       crossorigin="anonymous"
+     >
+   </head>
+   <body>
+     <div id="app">
+       <div class="_14y9ssh0">
+         <div class="_14y9ssh1">
+-          Initial
++          Client
```
askoufis, line 55: This missing hydration diff was happening because the client never actually ran.

fixtures/security-controls/sku-server.config.ts
```diff
@@ -5,6 +5,8 @@ import {
 
 export default {
   port: 9847,
+  // Required for the test to serve client assets from http-server.
+  publicPath: 'http://localhost:9848',
```
askoufis, line 9: This is what other SSR fixtures do. It's not great because there's potential for a port clash, but I think that's unlike.y

PR #1707, opened 2026-08-31
vite: Load required chunks before executing client

askoufis:
Hydration errors have been occurring in multi-language Vite apps for a while because the translation chunk that Vocab emits imports the sku client and cannot be guaranteed to run before hydration kicks in. https://github.com/seek-oss/vocab/pull/428 fixes the Vocab side of this by ensuring translation chunks are isolated from the client.

The other side of the fix requires us to actually implement an API similar to webpack's `loadableReady` in the Vite client. There's more info in the spec, but the implementation effectively boils down to:
1. Mark registered chunks with `data-required-chunk`
2. Run those chunks before calling the client

This bug went unnoticed for a long time because we don't capture [`pageerrors`](https://playwright.dev/docs/api/class-page#page-event-page-error) in our playwright tests. I believe the hydration error ends up as a page error only when it bubbles to the top of a React app (i.e. an error boundary does not catch it). The other subtle signal that things were broken were snapshot mismatches that are now fixed in this PR.


tests/browser/__snapshots__/translations.test.ts.snap
```diff
@@ -4,63 +4,62 @@ exports[`translations > bundler vite > should render en 1`] = `
...
+      crossorigin
+    >
+  </head>
+  <body>
+    <div id="app">
+      <div>
+        Hello sku
+        Company name
+      </div>
+    </div>
```
askoufis, line 35: This is the fix in action. Previously the client would hydrate with an empty string, and then the translation text would be inserted in a way that left out the newline for some reason. Now both the pre- and post- hydration DOMs are identical.

fixtures/translations/src/client.tsx
```diff
@@ -1,14 +1,22 @@
...
 import { hydrateRoot } from 'react-dom/client';
 
 import App from './App';
 import type { RenderContext } from './types';
 
 export default ({ language: skuLanguage }: RenderContext) => {
   const Client = () => {
-    const urlParameters = new URLSearchParams(window.location.search);
-    const isPseudo = Boolean(urlParameters.get('pseudo'));
-    const language = isPseudo ? 'en-PSEUDO' : skuLanguage;
```
askoufis, line 11: The old implementation still caused hydration errors even with the fix, and actually wasn't an accurate reflection of the test: the test was intended to test switching the language _post-hydration_.

tests/browser/translations.test.ts
```diff
@@ -59,7 +64,7 @@ describe('translations', () => {
 });
 
 describe('webpack ssr translations', () => {
-  const backendUrl = `http://localhost:8310`;
+  const backendUrl = `http://localhost:8314`;
```
askoufis, line 67: This port was wrong.

pnpm-workspace.yaml
```diff
@@ -65,6 +65,7 @@ overrides:
   # Multiple versions of esbuild tend to crop up in the lockfile. Pin to a specific minor version
   # purely to reduce installation size.
   esbuild: ~0.28.0
+  '@vocab/vite': '1.0.6-vite-isolate-translations-chunk-20260828063901'
```
askoufis: Will be removed once https://github.com/seek-oss/vocab/pull/428 is merged and released.

.changeset/vite-ssg-loadable-ready.md
askoufis, line 1: Does this changeset also need to mention that `@sku-lib/vite` needs to be updated, and `@vocab/vite` if you happen to depend on it (dependency should no longer necessary after #1687)?

jahredhope (approved):

.changeset/vite-ssg-loadable-ready.md
askoufis: Answer: we decided not to add more info to changeset given that the issue is relatively isolated and will resolve itself over time via renovate PRs.

PR #1711, opened 2026-09-01
Identify production SSR bootstrap as the ssr-client manifest chunk

SpiralLogic:
## Summary
- Identify the production SSR Document bootstrap as the Vite client manifest chunk named `ssr-client`, instead of the first `isEntry` chunk.
- Fail production server start if that chunk is missing.
- Add unit tests and a patch changeset.

## Why
- The client build can emit more than one `isEntry` chunk, so the first match is not necessarily sku’s SSR client.
- Wrong bootstrap is user-visible whenever extra `isEntry` chunks exist.

Made with [Cursor](https://cursor.com)


packages/sku/src/services/vite/ssr/resolveAssets.ts
```diff
@@ -38,10 +38,17 @@ export const warnUnknownModuleIdsWithoutManifest = (
...
 };
 
+/** Rolldown `[name]` for `#entries/ssr-client` (`entryFileNames: '[name]-[hash].js'`). */
+const SSR_CLIENT_CHUNK_NAME = 'ssr-client';
+
 export const findEntryChunk = (manifest: ClientManifest) => {
-  const entry = Object.values(manifest).find((chunk) => chunk.isEntry);
+  const entry = Object.values(manifest).find(
+    (chunk) => chunk.isEntry && chunk.name === SSR_CLIENT_CHUNK_NAME,
+  );
```
jahredhope: This can likely use `findManifestChunk`

e.g. 


```suggestion
export const findEntryChunk = (manifest: ClientManifest) => {
  const entry = findManifestChunk(manifest, SSR_CLIENT_CHUNK_NAME);
  if (!entry) {
    throw new Error(
      `No "${SSR_CLIENT_CHUNK_NAME}" chunk found in the Vite client manifest.`,
    );
  }
  return entry;
};
```

packages/sku/src/services/vite/ssr/resolveAssets.ts
```diff
@@ -38,10 +38,17 @@ export const warnUnknownModuleIdsWithoutManifest = (
   }
 };
 
+/** Rolldown `[name]` for `#entries/ssr-client` (`entryFileNames: '[name]-[hash].js'`). */
+const SSR_CLIENT_CHUNK_NAME = 'ssr-client';
```
jahredhope: SSR_CLIENT_CHUNK_NAME can likely get used in `vite/plugins/ssr.ts` to ensure it's named correctly.

jahredhope (changes requested):

jahredhope (approved):

askoufis (approved):

PR #1716, opened 2026-09-01
Swapping findByText with toMatchExitCode

williamlark:
Replacing the `findByText('Sku build complete'))` with `toMatchExitCode(0)`. While `toMatchExitCode` is slightly quicker, the main benefit is that it fails faster if builds fail. `findByText` will wait the entire timeout before erroring if it can't find the text, while `toMatchExitCode` will error as soon as the process exits with an incorrect code, which is usually much earlier than the max timeout limit. It also means we aren't reliant on the console build text in every test. 

Also, I moved the vite specific `security-controls` tests into their own `describe()` so they don't show up as `skipped`. They did technically get skipped since it used a `runIf('vite')` check, which skipped the tests depending on the bundler. It's a little confusing though since they did still run, and can just be their own`describe()`.


tests/browser/__snapshots__/security-controls.test.ts.snap
askoufis, line 1: These snapshots look like they just moved around a bunch, resulting in this large diff even though nothing has changed. Flagging this just in case...

askoufis (approved):

PR #1717, opened 2026-09-01
Add entrySideEffects so Vite apps can import isomorphic modules before any consumer module

jahredhope:
## Purpose

Sku owns Vite client and SSR entries, so an app cannot put a module first in the graph by importing it first in `RootLayout` or `App`.
Braid’s CSS reset must evaluate before any Braid component, and `sku start` throws if it does not.
Import order in consumer files is not a reliable contract because eslint sorts local imports.

## Changes

- Add `entrySideEffects` (default `[]`) so Vite static and SSR graphs evaluate listed modules first, in array order, resolved from the app.
- Inject them via `virtual:sku/entry-side-effects` on client, Node render, and SSR server entries, and prepend that virtual id to start CSS collection.
- Set `entrySideEffects: ['braid-design-system/reset']` on the SSR create template, and document the option versus `polyfills`.
- Cover the virtual module, entry import order, and Braid SSR start via the braid-design-system fixture (reset only in config).
- Include a minor sku / patch `@sku-lib/create` changeset and the active OpenSpec change (not archived in this PR).


fixtures/braid-design-system/src/webpack-static/render.tsx
```diff
@@ -1,15 +1,17 @@
+import 'braid-design-system/reset';
```
askoufis, line 1: Shouldn't this (and the same line in the SSG client entry) not be necessary if you add `entrySideEffects` to the SSG sku config?

packages/sku/src/types/types.ts
```diff
@@ -338,6 +338,19 @@ export interface SkuConfigBase {
    */
   libraryFile?: string;
 
+  /**
+   * An array of isomorphic modules to import before any consumer module on Vite
+   * client, SSR server, and static start render graphs.
```
askoufis: I'm not a fan of `static start render`. Feel like the AI seems to think the render entrypoint is `start` only or something?

packages/sku/src/types/types.ts
```diff
@@ -338,6 +338,19 @@ export interface SkuConfigBase {
...
+   *
+   * Use this for CSS resets (for example `braid-design-system/reset`) that must
+   * evaluate first. Specifiers resolve from the app, in array order.
+   * Webpack ignores this option.
+   *
+   * @default []
+   * @link https://seek-oss.github.io/sku/configuration#entrysideeffects
+   */
+  entrySideEffects?: string[];
```
askoufis: This should be under the Vite-specific config type.

site/docs/ssr/providers.md
```diff
@@ -152,9 +150,24 @@ See [Multi-language](./multi-language.md) and [Apollo streaming hydration](./dat
...
+import type { SkuConfig } from 'sku';
+
+export default {
+  bundler: 'vite',
+  buildType: 'ssr',
+  entrySideEffects: ['braid-design-system/reset'],
+} satisfies SkuConfig;
+```
+
+A reset import in the root layout is harmless (ESM singleton) but is not the graph guarantee.
```
askoufis: I either am struggling to understand this sentence, or it just reads a bit weird. Particularly `is not the graph guarantee`.

fixtures/braid-design-system/src/webpack-static/render.tsx
jahredhope: Only for vite. Fixed for vite.

site/docs/ssr/providers.md
jahredhope: Cleaned up (mostly removed)

packages/sku/src/types/types.ts
jahredhope: Agree. Done.

packages/sku/src/types/types.ts
jahredhope: Cleaned up.

packages/sku/src/services/vite/plugins/entrySideEffects.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    expect(() =>
+      loadEntrySideEffectsModule(['this-package-definitely-does-not-exist']),
+    ).toThrow();
+  });
+
+  it('exports the virtual module id used by sku entries', () => {
+    expect(ENTRY_SIDE_EFFECTS_VIRTUAL_ID).toBe(
+      'virtual:sku/entry-side-effects',
+    );
+  });
```
askoufis: Probably don't need this test.

packages/create/templates/ssr/sku.config.ts
```diff
@@ -6,6 +6,7 @@ const skuConfig = {
   publicPath: '/',
   expressTrustProxy: true,
   testRunner: 'vitest',
+  entrySideEffects: ['braid-design-system/reset'],
```
askoufis, line 9: Should this be applied to the Vite SSG template too?

.changeset/entry-side-effects.md
```diff
@@ -0,0 +1,6 @@
+---
+'sku': minor
+'@sku-lib/create': patch
+---
+
+Add `entrySideEffects` so apps can import isomorphic modules (such as Braid reset) before any consumer module.
```
askoufis, line 5: This _could_ be 2 separate changesets so we can specify in the `create` changeset that only the `vite` templates are affected, but I'm not too fussed if you want to keep it as one.

askoufis (approved):

.changeset/entry-side-effects.md
jahredhope: Separated

packages/create/templates/ssr/sku.config.ts
jahredhope: Yes, extended to vite

packages/sku/src/services/vite/plugins/entrySideEffects.test.ts
jahredhope: removed

PR #1718, opened 2026-09-02
Replace pnpm-plugin-sku with statically managed values

williamlark:
> Blocked until the next major release

This PR contains the openspec files outlining the proposed changes. Check out the proposal.md first to get an idea of the changes and challenge anything you see fit.

Short summary:
Sku currently injects its recommended pnpm settings at runtime via pnpm-plugin-sku. This hides the config from static tooling like Renovate, requires version-gated install logic and runtime detection workarounds, and splits ownership across two packages — consumers must update pnpm-plugin-sku separately from sku. Writing the config statically into pnpm-workspace.yaml makes it visible to all tooling and lets sku own the sync, so updating sku is the only upgrade step.

Renovate in particular is impacted since it doesn't handle the checksums generated when using pnpm config plugins at the moment, so manual `pnpm i` needs to be run on renovate branches (see [Slack thread](https://seekchat.slack.com/archives/CDL5VP5NU/p1786935852766509))


openspec/changes/static-pnpm-workspace-config/design.md
```diff
@@ -0,0 +1,158 @@
...
+
+- Skip conditions mirror existing behaviour: `configureProject` already skips all of configure when `skuSkipConfigure` is set, postinstall is already gated by `skuSkipPostInstall`, and `sku configure` is a manual, intentional invocation that always syncs.
+  No new wiring is needed: the sync lives in `configureApp` and inherits these gates.
+
+### Decision: Managed means enforced, no never-downgrade
+
+Managed scalar/enum values (`blockExoticSubdeps`, `minimumReleaseAge`, `strictDepBuilds`, `trustPolicy`) are always overwritten with sku's current defaults, in both directions.
+
+- Over never-downgrade for `trustPolicy`/`minimumReleaseAge`: a strength-ordering table is extra machinery with a maintenance burden (pnpm adding new policy values) and a fuzzy fail-safe.
+  Simplicity and predictability win: sku enforces org-wide policy, and the escape hatch (`skuSkipConfigure`) is documented.
```
SpiralLogic: `skuSkipConfigure` also skips configuring eslint, prettier etc.. and doesn't provide a practical escape-hatch. write-if-absent for trustPolicy/strictDepBuilds, and enforce only the keys sku actually needs to pin, would preserve the goal without clobbering tighter policy.

openspec/changes/static-pnpm-workspace-config/specs/pnpm-workspace-config/spec.md
```diff
@@ -0,0 +1,154 @@
...
+
+Values written by the sync SHALL carry a trailing `# managed by sku` comment marker.
+This applies to managed scalars and to each sku-owned entry within merged collections, so user-added entries stay visually distinct.
+Markers MUST NOT replace existing user comments.
+An entry carrying a user comment is never marked, so it is never adopted.
+A file whose values and markers already match sku's defaults MUST NOT be rewritten.
+
+#### Scenario: Managed scalar is marked
+
+- **WHEN** the sync adds or overwrites a managed scalar such as `minimumReleaseAge`
```
SpiralLogic: will this cause an the explanatory comment to become stale, for example if `#minimumReleaseAge: 1440 # we want 1 day` becomes 3 days, the comment will then be stale

williamlark:
Moving this draft while I finish the implementation

packages/create/package.json
```diff
@@ -26,7 +26,6 @@
     "dedent": "^1.5.1",
     "eta": "^4.5.0",
     "fdir": "^6.1.1",
-    "pnpm-plugin-sku": "workspace:*",
     "prompts": "^2.4.2",
     "yaml": "^2.8.3"
```
SpiralLogic, line 30: can remove this one now its gone the pnpm workspace generator

knip.jsonc
```diff
@@ -28,6 +28,10 @@
         "__sku_alias__webpackStats",
       ],
     },
+    "packages/create": {
+      // Bundled through @sku-private/utils by the shared workspace config writer
+      "ignoreDependencies": ["yaml"],
```
SpiralLogic, line 37: does it still need this?

samchungy:
I reckon skuba and sku might duke it out in repos where they have both soon 😆

williamlark:
@samchungy yea I did catch it overriding things in monorepos. I'm making some changes so that doesn't happen in monorepos unless a user runs something like `sku configure workspace` manually. Gives the user the option to opt in if they want or just leave what they have so it doesn't collide with skuba. Downside is they won't get notices during lint in a monorepo if sku changes pnpm default values, but I think we need to think about how sku and skuba can live happily together in a monorepo for that to happen 👪

PR #1721, opened 2026-09-09
Pin new sku apps to the running pnpm version

SpiralLogic:
## Summary
- Pin new `@sku-lib/create` apps to the pnpm version that ran the CLI (`packageManagerVersion`), instead of looking up npm's `latest-11` dist-tag.
- Point interactive sku-create tests at `interactive-*` directories so they no longer share `new-project-*` with the full-install suite.

## Why
- `latest-11` can be ahead of the pnpm installed in CI (for example 11.26.0 vs 11.24.0). Create then ran `pnpm add --config pnpm-plugin-sku` in a directory whose `packageManager` field forced a version switch, which failed under `pnpm/action-setup` with ENOENT.
- The prompt tests and the full-install suite both used `new-project-*`, so one run could leave files the other then refused to overwrite.

### DIFF
--- a/.changeset/create-skip-pnpm-version-switch.md
+++ b/.changeset/create-skip-pnpm-version-switch.md
@@ -1,0 +1,7 @@
+---
+'@sku-lib/create': patch
+---
+
+Pin new pnpm apps to sku's `packageManager` version, instead of the running CLI or npm's `latest-11` tag.
+
+Looking up `latest-11` or using the invoking pnpm can disagree with sku's pin (for example 11.26.0 vs 11.24.0 in CI). `pnpm add` then tried to switch versions and failed under `pnpm/action-setup`.
--- a/.changeset/create-skip-pnpm-version-switch.md
+++ b/.changeset/create-skip-pnpm-version-switch.md
@@ -2,6 +2,4 @@
 '@sku-lib/create': patch
 ---
 
-Pin new pnpm apps to sku's `packageManager` version, instead of the running CLI or npm's `latest-11` tag.
-
-Looking up `latest-11` or using the invoking pnpm can disagree with sku's pin (for example 11.26.0 vs 11.24.0 in CI). `pnpm add` then tried to switch versions and failed under `pnpm/action-setup`.
+`@sku-lib/create` version now tracks the `pnpm` version it was released with, so scaffolded apps always get a consistent, repeatable result.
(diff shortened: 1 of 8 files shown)


tests/node/sku-create.test.ts
```diff
@@ -39,6 +39,8 @@ const templates = ['vite', 'webpack', 'ssr'] as const;
 type Template = (typeof templates)[number];
 
 const projectName = (template: Template) => `new-project-${template}`;
+const interactiveProjectName = (template: Template) =>
+  `interactive-${template}`;
```
SpiralLogic, line 42: initially had this in a separate PR, but both fixes need to be in for the pipeline to pass:
https://github.com/seek-oss/sku/actions/runs/34297007191/job/102295733896#step:12:386

because the tests are running concurrently, and are all creating the project file in the same directory they collide. which manifested as a directory not being open when one of the cleanups attempts to run

packages/create/src/generators/packageJson.ts
```diff
@@ -42,23 +43,15 @@ export const generatePackageJson = async (
   await writeFile(packageJsonPath, packageJsonContent, 'utf8');
 };
 
-const resolvePackageManagerField = async (
+const resolvePackageManagerField = (
   packageManager: SupportedPackageManager,
-): Promise<string | null> => {
-  if (packageManager !== 'pnpm') {
+): string | null => {
+  if (packageManager !== 'pnpm' || !packageManagerVersion) {
```
askoufis: I'm hesitant to rely on the consumer's PNPM version here. The entire point of controlling the version is to increase the reliability of project init. If anything we should tie it to the monorepo's pnpm version because then we can guarantee all 3 places where `create` is run align: local, CI, and the consumer's machine.

.changeset/create-skip-pnpm-version-switch.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sku-lib/create': patch
+---
+
+Pin new pnpm apps to sku's `packageManager` version, instead of the running CLI or npm's `latest-11` tag.
+
+Looking up `latest-11` or using the invoking pnpm can disagree with sku's pin (for example 11.26.0 vs 11.24.0 in CI). `pnpm add` then tried to switch versions and failed under `pnpm/action-setup`.
```
askoufis: This is a bit too much implementation detail. I think it can be much terser:

```suggestion
`@sku-lib/create` versions now track the `pnpm` version it was released with, so scaffolded apps always get a consistent, repeatable result.
```

askoufis (approved):

PR #1726, opened 2026-09-15
Document ErrorBoundary placement for the html route

SpiralLogic:
## Summary
- Docs describe two `ErrorBoundary` placements: a child route for page failures, and an optional html-route fallback that renders a full `<html>` document for shell failure.
- `SkuRouteObject` allows `ErrorBoundary` and `errorElement` at every level, including via `lazy`.
- OpenSpec and a sku patch changeset record the rule.

## Why
- An `ErrorBoundary` on the route that renders `<html>` replaces that `Component`.
- A type error on `routes[0]` does not encode the html `Component` rule.
- A document-shaped html-route fallback is still needed for `onShellError`.


.changeset/root-errorboundary-type.md
```diff
@@ -0,0 +1,6 @@
+---
+'sku': patch
+---
+
+Do not set `ErrorBoundary` on a `SkuRouteObject` html root route.
+Set the boundary on a child route.
```
askoufis: The wording here should be about the change, not necessarily an instruction saying "don't do this".

site/docs/ssr/error-pages.md
```diff
@@ -17,6 +17,7 @@ That can briefly flash a stack trace before the default “Hey developer” page
 
 Because [`RootLayout`](./providers.md#root-layout-for-providers) renders `<html>`, attaching `ErrorBoundary` to the root route itself replaces that layout on failure and drops `<html>` from the response.
 Instead, attach `ErrorBoundary` to a child route under `RootLayout` so the document shell stays mounted.
+Do not set `ErrorBoundary` on a top-level [`SkuRouteObject`](./routing.md).
```
askoufis: The lines above sort of already say not to do this. Maybe they can be reworded to incorporate the route object link.

site/docs/ssr/entries.md
```diff
@@ -331,6 +331,8 @@ export const mapRoutePath?: MapRoutePath;
 ```
 
 `SkuRouteObject` is a React Router `RouteObject` plus optional `sites` for multi-site membership.
+Do not set `ErrorBoundary` on a top-level route.
```
askoufis: Again, these instructions feel a bit blunt. Should we link somewhere to justify them? I.e. have a single source of truth stating why you should not set `ErrorBoundary` at the top-level. In fact, rather than saying "don't do this", it might be better to say "you can't do this. here's why:" instead.

SpiralLogic: will link to the docs on why, the reason it seems blunt especially in current context, I had it written in STE, so that when Jahred's PR lands, it wasn't out of place.

PR #1729, opened 2026-09-22
Removing postinstall

williamlark:
Most of the postinstall scripts are run before every command already, so we can remove it without losing any functionality.

This would slightly impact fresh git clones of sku projects since they will not get auto-generated files (.prettierrc, tsconfig, etc.) files until they run a sku command (or manually run `sku configure`) but most projects will run `sku start` or similar once the projects is installed anyway, so impact is minimal.

Now that we don't have a postinstall sku doesn't need to be included in `pnpm-workspace.yaml#allowBuilds`, but I've left it there since it doesn't hurt to keep. Open to removing it though.

I've moved the devDep check to the configure function to keep its functionality since that was the only check that ran exclusively in postinstall, but I'm open to just removing it if we don't think it's that useful anymore.


openspec/changes/remove-postinstall/design.md
```diff
@@ -0,0 +1,92 @@
...
+changeset. Typical projects need no action. Automation that reads generated
+configs before running a sku command adds a `sku configure` step.
+
+Rollback: revert the PR and re-release.
+
+## Open Questions
+
+- Drop `sku: true` from sku's recommended pnpm `allowBuilds` in the same
+  release? Leaning toward no — harmless while present, and removal belongs to
+  the pnpm workspace config capability.
```
williamlark, line 92: I've kept sku in `allowBuilds` but open to removing it since it wouldn't be needed anymore now that we don't run a postinstall

packages/sku/src/utils/configureApp.ts
```diff
@@ -33,11 +35,32 @@ const convertToForwardSlashPaths = (pathStr: string) =>
 
 const addSep = (p: string) => `${p}${path.sep}`;
 
+const warnOnSkuDependency = () => {
```
williamlark, line 38: This used to live in `postinstall`. Moved to configure, but we could remove it if this isn't as much of an issue anymore.

.changeset/remove-postinstall.md
```diff
@@ -0,0 +1,9 @@
+---
+'sku': minor
+---
+
+Remove the `postinstall` hook
+
+Sku no longer runs a script at install time. Generated project files (`tsconfig.json`, `eslint.config.mjs`, `.prettierrc`, etc.) are no longer created during `install` and instead are written by the first run of any sku command or by running `sku configure` explicitly.
```
askoufis: Just a clarification here: the project files are emitted every time you run a command, not just the first time.

```suggestion
Sku no longer runs a script at install time. Generated project files (`tsconfig.json`, `eslint.config.mjs`, `.prettierrc`, etc.) are no longer created during `install` and instead are written by running any sku command or by running `sku configure` explicitly.
```

openspec/changes/remove-postinstall/design.md
askoufis: If we do remove it, I'd say it would be a breaking change (0-version `minor`) in the pnpm config package. But given we're planning to ditch pnpm config soon I don't think it's worth the churn.

askoufis (approved):

